PR #3, opened 2014-02-05
Fix dealing with Unicode paths in v2 API

mistydemeo:
This adds a new API revision, v2, which attempts to deal with the many problems we've had dealing with Unicode paths.

Paths are basically bytestrings to the OS, and we've caused ourselves a lot of grief by trying to treat them as sets of characters instead of bytes. v2 of the API sidesteps this by treating all paths as sets of arbitrary bytes, and communicates them to/from outside services in base64-encoded strings.

The tastypie resources are now separated out into base classes and versioned classes, which inherit from the base resources.

This will get squashed down a bit on merge.

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -1,3 +1,6 @@
+# This file contains the base models that individual versioned models
+# are based on. They shouldn't be directly used with Api objects.
+
 # stdlib, alphabetical
 import json
 import logging
@@ -53,6 +56,7 @@ class PipelineResource(ModelResource):
         authorization = Authorization()
         # authorization = DjangoAuthorization()
         # validation = CleanedDataFormValidation(form_class=PipelineForm)
+        resource_name = 'pipeline'
 
         fields = ['uuid', 'description']
         list_allowed_methods = ['get', 'post']
@@ -82,6 +86,7 @@ class SpaceResource(ModelResource):
         authorization = Authorization()
         # authorization = DjangoAuthorization()
         validation = CleanedDataFormValidation(form_class=SpaceForm)
+        resource_name = 'space'
 
         fields = ['access_protocol', 'last_verified', 'location_set', 'path',
             'size', 'used', 'uuid', 'verified']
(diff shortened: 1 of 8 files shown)


storage_service/locations/api/urls.py
```diff
@@ -1,15 +1,22 @@
 from django.conf.urls import patterns, include, url
 from tastypie.api import Api
-from .resources import (LocationResource, SpaceResource, PackageResource,
-    PipelineResource)
+import locations.api.v1 as v1
```
Hwesta: `from locations.api import v1` ?

storage_service/locations/api/v2.py
```diff
@@ -0,0 +1,45 @@
+import base64
+
+import locations.api.resources as resources
+
+from tastypie import fields
+
+
+class PipelineResource(resources.PipelineResource):
+    create_default_locations = fields.BooleanField(use_in=lambda x: False)
+    shared_path = fields.CharField(use_in=lambda x: False)
```
Hwesta, line 10: These are included in resources.PipelineResource but not in v1.PipelineResource - do the child classes need to have these defined or no?  Same for the rest of the resources.

storage_service/locations/api/resources.py
```diff
@@ -140,6 +145,9 @@ def obj_create(self, bundle, **kwargs):
         obj.save()
         return bundle
 
+    def get_objects(self, space, path):
+        pass
```
Hwesta: Should this raise a NotImplementedError, or do we actually want it to return nothing if a later API version doesn't implement it?  Same for LocationResource.get_objects

storage_service/static/js/base64.js
```diff
@@ -0,0 +1,216 @@
+/*
+Copyright (c) 2008 Fred Palmer fred.palmer_at_gmail.com
```
Hwesta: Another merge, but we should probably move all the stuff written externally to a vendor/ directory.

mistydemeo:
OK, rebased and addressed comments - think this is ready to merge.

PR #4, opened 2014-03-14
Move files more generic

Hwesta:
Rather than every Space knowing how to move to every other Space, they only need to know how to move to and from a staging area in the storage service.  The staging area must be locally accessible to the storage service (ie with tools like mv and os.rename), and is configured for each Space.

Each space implements functions to move files/folders to and from the storage service. Moving to the storage service from the original location deposits them in a staging area associated with the destination Space.  Moving from the storage service to the final destination moves the file from the destination Space's staging area to the destination path.

Also added support for moving folders, not just files.  Rsync uses -r, and extract files looks for the file on disk before trying to extract it from a Package.

No Archivematica changes needed.

### DIFF
--- a/storage_service/locations/models.py
+++ b/storage_service/locations/models.py
@@ -172,15 +172,150 @@ class Space(models.Model):
             entries = []
         return {'directories': directories, 'entries': entries}
 
+    def move_to_storage_service(self, source_path, destination_path,
+                                destination_space, *args, **kwargs):
+        """ Move source_path to destination_path in the staging area of destination_space.
+
+        If source_path is not an absolute path, it is assumed to be relative to
+        Space.path.
+
+        destination_path must be relative and destination_space.staging_path
+        MUST be locally accessible to the storage service.
+
+        This is implemented by the child protocol spaces.
+        """
+        logging.debug('TO: src: {}'.format(source_path))
+        logging.debug('TO: dst: {}'.format(destination_path))
+        logging.debug('TO: staging: {}'.format(destination_space.staging_path))
+        # TODO move the path mangling to here?
+        # TODO enforce source_path is inside self.path
+        try:
+            self.get_child_space().move_to_storage_service(
+                source_path, destination_path, destination_space, *args, **kwargs)
+        except AttributeError:
+            raise NotImplementedError('{} space has not implemented move_to_storage_service'.format(self.get_access_protocol_display()))
+
+    def post_move_to_storage_service(self, *args, **kwargs):
+        """ Hook for any actions that need to be taken after moving to the storage service. """
+        try:
+            self.get_child_space().post_move_to_storage_service(*args, **kwargs)
+        except AttributeError:
+            # This is optional for the child class to implement
+            pass
+
+    def move_from_storage_service(self, source_path, destination_path,
+                                  *args, **kwargs):
+        """ Move source_path in this Space's staging area to destination_path in this Space.
+
+        That is, moves self.staging_path/source_path to self.path/destination_path.
+
+        If destination_path is not an absolute path, it is assumed to be
+        relative to Space.path.
+
+        source_path must be relative to self.staging_path.
+
+        This is implemented by the child protocol spaces.
+        """
+        logging.debug('FROM: src: {}'.format(source_path))
+        logging.debug('FROM: dst: {}'.format(destination_path))
+        # TODO move the path mangling to here?
+        # TODO enforce destination_path is inside self.path
+        try:
+            self.get_child_space().move_from_storage_service(
+                source_path, destination_path, *args, **kwargs)
+        except AttributeError:
+            raise NotImplementedError('{} space has not implemented move_from_storage_service'.format(self.get_access_protocol_display()))
+
+    def post_move_from_storage_service(self, *args, **kwargs):
+        """ Hook for any actions that need to be taken after moving from the storage service to the final destination. """
+        try:
+            self.get_child_space().post_move_from_storage_service(*args, **kwargs)
+        except AttributeError:
+            # This is optional for the child class to implement
+            pass
+
+    # HELPER FUNCTIONS
+
+    def _move_locally(self, source_path, destination_path, mode=None):
+        """ Moves a file from source_path to destination_path on the local filesystem. """
+        logging.info("Moving from {} to {}".format(source_path, destination_path))
+
+        # Create directories
+        self._create_local_directory(destination_path, mode)
+
+        # Move the file
+        os.rename(source_path, destination_path)
+
+    def _move_rsync(self, source, destination):
+        """ Moves a file from source to destination using rsync.
+
+        All directories leading to destination must exist.
+        Space._create_local_directory may be useful.
+        """
+        # Create directories
+        logging.info("Rsyncing from {} to {}".format(source, destination))
+
+        # Rsync file over
+        # TODO Do this asyncronously, with restarting failed attempts
+        command = ['rsync', '--chmod=ugo+rw,ugo-x', source, destination]
+        logging.info("rsync command: {}".format(command))
+        try:
+            subprocess.check_call(command)
+        except subprocess.CalledProcessError as e:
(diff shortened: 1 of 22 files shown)


storage_service/locations/models.py
```diff
@@ -168,15 +172,160 @@ def browse(self, path):
...
+        # Create directories
+        logging.info("Rsyncing from {} to {}".format(source, destination))
+
+        # Rsync file over
+        # TODO Do this asyncronously, with restarting failed attempts
+        command = ['rsync', '--chmod=ugo+rw', '-r', source, destination]
+        logging.info("rsync command: {}".format(command))
+        try:
+            subprocess.check_call(command)
+        except Exception as e:
```
mistydemeo: Catch `CalledProcessError` specifically?

storage_service/locations/models.py
```diff
@@ -168,15 +172,160 @@ def browse(self, path):
...
+            logging.warning("Rsync failed: {}".format(e))
+            raise
+
+    def _create_local_directory(self, path, mode=None):
+        """ Creates a local directory at 'path' with 'mode' (default 755). """
+        if mode is None:
+            mode = (stat.S_IRUSR + stat.S_IWUSR + stat.S_IXUSR +
```
mistydemeo, line 280: Pass octal 0755 for brevity?

storage_service/locations/models.py
```diff
@@ -168,15 +172,160 @@ def browse(self, path):
...
+                    stat.S_IROTH +                stat.S_IXOTH)
+        try:
+            os.makedirs(os.path.dirname(path), mode)
+        except os.error as e:
+            # If the leaf node already exists, that's fine
+            if e.errno != errno.EEXIST:
+                logging.warning("Could not create storage directory: {}".format(e))
+                raise
+
+        # Mode not getting set correctly in os.makedirs
```
mistydemeo: Is this always true, or only in some circumstances?

storage_service/locations/models.py
```diff
@@ -168,15 +172,160 @@ def browse(self, path):
...
+        # dest_path must be relative
+        if os.path.isabs(dest_path):
+            dest_path = dest_path.lstrip(os.sep)
+            # os.path.join(*dest_path.split(os.sep)[1:]) # Strips up to first os.sep
+        destination_path = os.path.join(dest_space.staging_path, dest_path)
+        # Archivematica expects the file to still be on disk even after stored
+        self.space._create_local_directory(destination_path)
+        return self.space._move_rsync(source_path, destination_path)
+
+    def move_from_storage_service(self, src_path, dest_path):
```
mistydemeo, line 319: These two methods are doing very similar things, maybe the logic can be extracted?

storage_service/locations/models.py
```diff
@@ -236,24 +418,105 @@ class PipelineLocalFS(models.Model):
...
+        verbose_name = "Pipeline Local FS"
+
+    def move_to_storage_service(self, src_path, dest_path, dest_space):
+        """ Moves src_path to dest_space.staging_path/dest_path. """
+        # if dest_space == self:
+        #     # If moving within same space, don't bring to storage service
+        #     # FIXME dest_path is relative, and intended for staging_path, need
+        #     # real destination path - memoize something and retrieve it in
+        #     # move_from_storage_service?  Move to self.space.path/dest_path?
+        #     command = 'mkdir -p {dest_dir} && mv {src_path} {dest_path}'.format(
```
mistydemeo: Quote `dest_dir` and `src_path`?

mistydemeo: Uh, of course this isn't actual code just a comment. Ignore the noise.

storage_service/static/js/file-explorer.js
```diff
@@ -199,6 +199,10 @@
           var child = entry.children[index]
             , allowDisplay = true;
 
+          // names are base64-encoded from the server; needs to be
+          // decoded for human reading
+          child.attributes.name = Base64.decode(child.attributes.name);
```
mistydemeo, line 204: Is this reused elsewhere? Do you want to mutate `child`, or can you work from a copy of the name?

storage_service/static/js/vendor/base64.js
```diff
@@ -0,0 +1,216 @@
+/*
```
mistydemeo, line 1: What is this branched from? This file already exists as `static/js/vendor/base64.js`.

storage_service/storage_service/urls.py
```diff
@@ -43,7 +43,9 @@ def startup():
     logging.basicConfig(level=logging.INFO)
     logging.info("Running startup")
     space, space_created = locations_models.Space.objects.get_or_create(
-        access_protocol=locations_models.Space.LOCAL_FILESYSTEM, path='/')
+        access_protocol=locations_models.Space.LOCAL_FILESYSTEM,
+        path='/',
```
mistydemeo: Still keeping `path='/'` when using `os.sep` below?

mistydemeo:
I only have nitpicks; this looks good! Seems like a good approach.

storage_service/static/js/vendor/base64.js
Hwesta, line 1: Originally branched from 0.x, then rebased onto qa/0.x when that became available.  Pull request is still against 0.x, so it has 4 extra commits relating to AICs and non-unicode/base64 stuff.

mistydemeo, line 1: Ah, okay.

storage_service/locations/models.py
Hwesta, line 280: I like the explicitness of the constants, though I'm convince-able on this one.

mistydemeo, line 280: I've used too many Unix commands in my lifetime, so 0755 is actually more explicit/readable to me! Not a super strong opinion though.

storage_service/locations/models.py
Hwesta: Seems to be true most of the time - `os.makedirs('foo/bar/baz', 0o777)` gives me directories with 755 permissions.  http://stackoverflow.com/questions/5231901/permission-problems-when-creating-a-dir-with-os-makedirs-python points out that the makedirs docs say the mode may be ignored.

storage_service/locations/models.py
Hwesta: Given that may be copy-pasted into actual code, probably worthwhile to do anyway.

storage_service/static/js/file-explorer.js
Hwesta, line 204: This code is actually from #3, so I won't deal with this.

storage_service/locations/models.py
mistydemeo: Okay, maybe add more detail to the comment? It wasn't clear to me what this was guarding against.

storage_service/static/js/file-explorer.js
mistydemeo, line 204: Oh no! The monster at the end of the PR... is _me_!

storage_service/locations/models.py
Hwesta: Would this be better? `os.makedirs may ignore the mode when creating directories, so force the permissions here. Some spaces (eg CIFS) doesn't allow chmod, so wrap it in a try-catch and ignore the failure.`

mistydemeo: :thumbsup: Thank you!

storage_service/locations/models.py
Hwesta, line 319: They are similar, but the path munging is different.  The similar logic is already extracted into `_create_local_directory` and `_move_rsync`.  Also, they may not always be identical (move_from used move_local instead of move_rsync for a while, and probably will again in future)

mistydemeo, line 319: Okay, works for me.

Hwesta:
Update with feedback from code review.  Also rebased off qa/0.x again, and got the migration changes, so added a migration.

mistydemeo:
Migrations! :D

PR #5, opened 2014-03-24
Add link to download pointer files

Hwesta:
Added a link to download pointer files associated with a Package, if they exist.  Refactored extract and download request code to pull out the section that returns a file stream, for use with the pointer file
downloads.  Also moved extract files to the Package, instead of in the resource.
#4 should be merged first, as this branch is based off it.  Supports artefactual/archivematica#25

### DIFF
--- a/storage_service/locations/models.py
+++ b/storage_service/locations/models.py
@@ -257,7 +257,7 @@ class Space(models.Model):
 
         # Rsync file over
         # TODO Do this asyncronously, with restarting failed attempts
-        command = ['rsync', '--chmod=ugo+rw,ugo-x', source, destination]
+        command = ['rsync', '--chmod=ugo+rw', '-r', source, destination]
         logging.info("rsync command: {}".format(command))
         try:
             subprocess.check_call(command)
--- a/storage_service/locations/models.py
+++ b/storage_service/locations/models.py
@@ -238,6 +238,11 @@ class Space(models.Model):
 
     def _move_locally(self, source_path, destination_path, mode=None):
         """ Moves a file from source_path to destination_path on the local filesystem. """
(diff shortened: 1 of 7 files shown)


storage_service/locations/models.py
```diff
@@ -503,78 +753,20 @@ def store_aip(self, origin_location, origin_path):
...
-        try:
-            # Mode not getting set correctly in os.makedirs
-            os.chmod(os.path.dirname(destination_path), mode)
-        except os.error as e:
-            logging.warning(e)
+        Returns path to the extracted file and a temp dir that needs to be
+        deleted. """
+        temp_dir = tempfile.mkdtemp()
+        output_path = os.path.join(temp_dir, os.path.basename(relative_path))
+        command = ['atool', '--cat', self.full_path(), relative_path]
```
mistydemeo, line 763: If I can bikeshed for a moment: any particular reason for picking atool? I was considering proposing `unar` as the new package extraction tool for main Archivematica. Might as well standardize on one we can use both there and in the storage service.

Hwesta, line 763: Because I used it in AM? https://github.com/artefactual/archivematica/search?q=atool Unar looks like it does all the same things, so I'd be happy to switch to it.

Whichever we choose, they need to be added to the debian/control

mistydemeo, line 763: Oh dang, I missed that!

unar doesn't create archives, just extract, so even if we started using that we'd still need atool.

mistydemeo:
:thumbsup:

storage_service/locations/models.py
Hwesta, line 763: The compression is still done by tar and 7z, just the extraction I moved to atool so it didn't care what the packaging was.

PR #6, opened 2014-04-01
Add migrations

Hwesta:
refs #6466

Enable South, and update version.  Add migrations for 0.2.2 and 0.3.0 for locations and administration.

### DIFF
--- a/debian/postrm
+++ b/debian/postrm
@@ -1,0 +1,18 @@
+#!/bin/sh
+
+#postrm 
+#run after the package is uninstalled
+
+#clean up storage service directories
+rm -rf /usr/share/python/archivematica-storage-service
+rm -rf /var/archivematica/storage*
+rm -r /var/archivematica/.storage-service
+rm -f /tmp/storage-service.log
+rm -f /etc/nginx/sites-enabled/storage
+rm -f /etc/nginx/sites-available/storage
+rm -f /etc/uwsgi/apps-enabled/storage.ini
+rm -f /etc/uwsgi/apps-available/storage.ini
+rm -rf /usr/lib/archivematica/storage*
+
+#DEBHELPER#
+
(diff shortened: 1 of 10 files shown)


mistydemeo:
Whoa, nice! Looks good. This is working well in testing?

Will this work

a) For fresh installs;
b) For installs from previous stable storage-service releases?

jhsimpson:
I added a commit that makes this work with the ubuntu packages.  I tested with a fresh install and with an update from 0.2.2 to 0.3.0.  This is ready to be merged into qa/0.x branch.

debian/postrm
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+#postrm 
+#run after the package is uninstalled
+
+#clean up storage service directories
+rm -rf /usr/share/python/archivematica-storage-service
+rm -rf /var/archivematica/storage*
```
Hwesta, line 8: This will remove pointer files stored in the SS internal storage - do we want that?

jhsimpson, line 8: I think so yes.  This is run when the user is un-installing the storage service.  We are not supporting an interactive un-install - I think it is fair to say don't un-install unless you have backed up whatever data you want to keep.

Hwesta:
Looks good.

debian/postrm
mistydemeo, line 8: I think Debian packages don't typically remove content from `/var`, but we stomp all over the FHS already so it's not really a big deal.

jhsimpson, line 8: Agreed, Archivematica and the Storage Service are not FHS compliant.  Archivematica puts the shared Directory structure used in processing in /var/archivematica, and the storage service followed this convention by putting its internal data in /var/archivematica/storage-service.

Both should probably be in /var/lib/, not /var.  It seems to be okay according to both FHS and debian/ubuntu practice for a package to delete items from its own /var/lib/<package-name> directory.

mistydemeo, line 8: Issue for discussing FHS compliance here: https://projects.artefactual.com/issues/6578

PR #7, opened 2014-04-15
Add FEDORA API

Hwesta:
Add support for depositing files from FEDORA, using SWORD2 as the API.  See https://www.archivematica.org/wiki/Sword_API for the full spec.

Adds a new FEDORA Space, new 'FEDORA/SWORD Deposit' purpose for locations, and Deposit type of package.  Pipelines now contain information for using the dashboard API.  Also adds several models to track asynchronous downloads to a deposit.

Interactions described on wiki (and in code docstrings), but the API includes:
- service document IRI (information about collections)
- collection IRI (information about deposits, create new deposits)
- deposit IRI (for getting information about the deposit, and finalizing it)
- deposit media IRI (for posting new media to a specific deposit)

History cleaned up as best I can - more can be done if requested.

### DIFF
--- a/debian/archivematica-storage-service.install
+++ b/debian/archivematica-storage-service.install
@@ -5,7 +5,7 @@ install/storage etc/nginx/sites-available/
 install/.storage-service var/archivematica/
 install/make_key.py var/archivematica/storage-service/
 
-#lib/* var/archivematica/storage-service/lib/
+lib/* var/archivematica/storage-service/lib/
 
 storage_service/static/* /var/archivematica/storage-service/static/
 storage_service/templates/* /var/archivematica/storage-service/templates/
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -415,7 +415,7 @@ class DepositResource(ModelResource):
 
     def sword_deposit(self, request, **kwargs):
         self.log_throttled_access(request)
-        return sword_views.deposit(request, kwargs['uuid'])
+        return sword_views.deposit_edit(request, kwargs['uuid'])
 
     def sword_deposit_media(self, request, **kwargs):
         self.log_throttled_access(request)
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -14,6 +14,7 @@ from django.core.exceptions import ObjectDoesNotExist, MultipleObjectsReturned
 from django.forms.models import model_to_dict
 
 # Third party dependencies, alphabetical
+from annoying.functions import get_object_or_None
 import bagit
 from tastypie.authentication import (BasicAuthentication, ApiKeyAuthentication,
     MultiAuthentication, Authentication)
@@ -233,13 +234,6 @@ class SpaceResource(ModelResource):
 
         return self.create_response(request, objects)
 
-    def sword_collection(self, request, **kwargs):
-        space = Space.objects.get(uuid=kwargs['uuid'])
-        if space.access_protocol != Space.SWORD_SERVER:
-            return http.HttpBadRequest('This is not a SWORD server space.')
-        self.log_throttled_access(request)
-        return sword_views.collection(request, kwargs['uuid'])
-
 
 class LocationResource(ModelResource):
     space = fields.ForeignKey(SpaceResource, 'space')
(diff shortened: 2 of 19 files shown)


storage_service/locations/api/resources.py
```diff
@@ -105,6 +107,7 @@ class Meta:
     def prepend_urls(self):
         return [
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/browse%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('browse'), name="browse"),
+
```
mistydemeo: New newlines?

storage_service/locations/api/sword/views.py
```diff
@@ -0,0 +1,616 @@
...
+    Example POST creation of deposit from another location:
+      curl -v -H "In-Progress: true" --request POST http://localhost:8000/api/v1/location/96606387-cc70-4b09-b422-a7220606488d/sword/collection/?source_location=aab142a9-018f-4452-8f93-67c1bf7fd486&relative_path_to_files=archivematica-sampledata/SampleTransfers/Images
+    """
+    if isinstance(location, basestring):
+        try:
+            location = models.Location.active.get(uuid=location)
+        except models.Location.DoesNotExist:
+            return helpers.sword_error_response(request, 404, 'Collection {uuid} does not exist.'.format(uuid=location))
+
+    # assert(isinstance(location, models.Location))
```
mistydemeo: Leftover?

storage_service/locations/api/sword/views.py
```diff
@@ -0,0 +1,616 @@
...
+
+                        if request.META['HTTP_IN_PROGRESS'] == 'true':
+                            return _deposit_receipt_response(request, deposit, 201)
+                        else:
+                            return _deposit_receipt_response(request, deposit, 200)
+                    else:
+                        return helpers.sword_error_response(request, 412, 'Error parsing XML')
+                except Exception as e:
+                    return helpers.sword_error_response(request, 400, traceback.format_exc())
+            elif source_location != '' or relative_path_to_files != '':
```
mistydemeo: If we're not distinguishing between empty strings and `None`, could do `elif source_location or relative_path_to_files` here and `if not source_location or not relative_path_to_files` on the next line, since `''` is falsy.

storage_service/locations/api/sword/views.py
```diff
@@ -0,0 +1,616 @@
...
+        deposit.save()
+
+    # create subprocess so content URLs can be downloaded asynchronously
+    helpers.spawn_download_task(deposit.uuid, mets_data['objects'])
+
+def _parse_name_and_content_urls_from_request_body(request):
+    """
+    From a request's body, parse deposit name and control URLs from METS XML
+
+    Returns None if parsing fails
```
mistydemeo, line 222: This isn't significant enough an error state to report?

storage_service/locations/api/sword/views.py
```diff
@@ -0,0 +1,616 @@
...
+
+    for element in elements:
+       url = element.get('{http://www.w3.org/1999/xlink}href')
+       filename = element.get('{http://www.w3.org/1999/xlink}title')
+
+       # only MD5 checksums currently supported
+       checksumtype = element.get('CHECKSUMTYPE')
+       checksum = element.get('CHECKSUM')
+
+       if checksum != None and checksumtype != 'MD5':
```
mistydemeo: `is not None`?

mistydemeo:
:thumbsup:

storage_service/locations/api/sword/views.py
Hwesta, line 222: Probably.  Both places that call it check for None and return an error if that's the case, however.

mistydemeo, line 222: Might make more sense to raise an exception here, then?

Hwesta, line 222: Probably, but a more thorough cleanup of this code is out of scope currently. :(

debian/archivematica-storage-service.install
```diff
@@ -5,7 +5,7 @@ install/storage etc/nginx/sites-available/
 install/.storage-service var/archivematica/
 install/make_key.py var/archivematica/storage-service/
 
-#lib/* var/archivematica/storage-service/lib/
+lib/* var/archivematica/storage-service/lib/
```
mistydemeo, line 8: Was this uncommented on purpose? There isn't a `lib` directory.

gneeek, line 8: There is a lib directory that is not under version control, that is
required in order for dh-virtualenv to be able to build packages on
launchpad.

The process of building a storage-service deb is this:

Update Debian/changelog
Run pip ( pip install -r requirements.txt -d lib )
Run debuild
Run dput

The lib directory is populated by pip.  If the lib directory is not
included in the Debian install file, that directory will  not be uploaded
to launchpad, and the build will fail, because it is not possible for pip
to fetch source from pypi via a launchpad server.

Possibly this needs to be documented a bit better, perhaps the build
process should itself be under version control.

PR #8, opened 2014-04-16
Add LOCKSS-o-matic support

Hwesta:
Adds support for LOCKSS (via LOCKSS-o-matic using the SWORD2 API) as an AIP Storage destination.  Details found https://www.archivematica.org/wiki/LOCKSS_Integration

In short,
- SS requests a service document from LOM, and parses it for max AU size, and collection to create AIPs in
- If the package is larger than the max AU size, use tar to break the file into chunks
- SS posts an atom entry creation document to LOM, including the URL(s) of the package to be uploaded
- SS parses the returned deposit receipt for the state and edit IRIs
- When package status is updated, query the state IRI for status.  If all servers are in agreement, optionally delete local copy of file

Other features added:
- common NSMAP for use with xml
- storage service version function (common/version.py)
- update_package_status function,which queries the space to check the current status of a package.  For LOM, will resend atom creation and/or query for the server states
- move_from SS but not move_to
- Package.misc_attributes - for storing information about packages that is only needed in some cases, eg state and edit IRI for LOM

Includes code from #5


storage_service/locations/models.py
```diff
@@ -490,6 +508,497 @@ def post_move_from_storage_service(self, staging_path, destination_path, package
...
+        help_text="If checked, keep a local copy even after the AIP is stored in the LOCKSS network.")
+
+    # Uses the SWORD protocol to talk to LOM
+    sword_connection = None
+    # Parsed pointer file
+    pointer_root = None
+
+    def move_to_storage_service(self, src_path, dest_path, dest_space):
+        """ Moves src_path to dest_space.staging_path/dest_path. """
+        # Check if in SS internal, if not then fetch from LOM
```
mistydemeo: Is this going to be done here?

storage_service/locations/models.py
```diff
@@ -490,6 +508,497 @@ def post_move_from_storage_service(self, staging_path, destination_path, package
...
+        # Create the atom entry XML
+        entry, slug = self._create_resource(package, output_files)
+
+        # Post to SWORD2 server
+        receipt = self.sword_connection.create(col_iri=self.collection_iri, metadata_entry=entry, suggested_identifier=slug)
+        try:
+            state_iri = receipt.atom_statement_iri
+            edit_iri = receipt.edit
+        except AttributeError:
+            logging.warning('Unable to contact LOCKSS for package %s', package.uuid)
```
mistydemeo: When would this AttributeError occur? If `receipt is None`?

mistydemeo:
:thumbsup:

storage_service/locations/models.py
Hwesta: When receipt is None, or if there's any error with receipt that results in atom_state_iri or edit_iri not being populate (eg not in the returned XML, wrong version of the SWORD2 library that doesn't populate them, etc)

Hwesta:
Rebased on current qa/0.x, fixed some rebase errors, and added a small fix for older versions of the pointer file (so this can be released before artefactual/archivematica#28 in 1.2)

storage_service/locations/models.py
mistydemeo: Maybe add a comment to this effect so I remember this, and don't try to figure out why this happening once I forget? :D

Hwesta: Done, and added a check that state_iri and edit_iri are populated before adding them to misc_attributes.

storage_service/common/utils.py
```diff
@@ -100,8 +118,101 @@ def download_file_stream(filepath, temp_dir=None):
...
+
 ############ OTHER ############
 
+def generate_checksum(file_path, checksum_type='md5'):
+    """
+    Returns checksum object for `file_path` using `checksum_type`.
+    """
+    try:
+        checksum = hashlib.new(checksum_type)
+    except ValueError:  # Invalid checksum_type
```
mistydemeo: Maybe it's worth leaving the raise in this case - silently returning a different checksum type could lead to weird hard-to-debug issues later.

storage_service/locations/api/resources.py
```diff
@@ -368,6 +416,7 @@ def prepend_urls(self):
         return [
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/delete_aip%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('delete_aip_request'), name="delete_aip_request"),
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/extract_file%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('extract_file_request'), name="extract_file_request"),
+            url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/download/(?P<chunk_number>\d+)%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('download_request'), name="download_lockss"),
```
mistydemeo, line 423: Not sure how I feel about LOCKS numbered chunk downloads sharing an API endpoint w/ regular downloads; it seems like it's a pretty dissimilar case, unless I'm misunderstanding.

storage_service/locations/models.py
```diff
@@ -486,6 +501,497 @@ def post_move_from_storage_service(self, staging_path, destination_path, package
...
+
+    def move_from_storage_service(self, source_path, destination_path):
+        """ Moves self.staging_path/source_path to destination_path. """
+        self.space._create_local_directory(destination_path)
+        return self.space._move_rsync(source_path, destination_path)
+
+    def post_move_from_storage_service(self, staging_path, destination_path, package):
+        # LOCKSS can only save packages in the storage service, since it needs
+        # to track information on it over time
+        if package is None:
```
mistydemeo, line 543: In what circumstances is `package` `None`? (I know this isn't specific to LOCKSS.)

storage_service/locations/models.py
```diff
@@ -486,6 +501,497 @@ def post_move_from_storage_service(self, staging_path, destination_path, package
...
+            return
+        # Post to Lockss-o-matic with the create resource atom entry
```
mistydemeo, line 545: This implies something else is meant to be happening here, rather than just the logging entry?

storage_service/locations/models.py
Hwesta, line 543: Hopefully, never.  `post_move_from_storage_service` is the hook for the LOM Space to talk to LOM.  Since the only thing that should be stored in LOM is AIPs, I don't want to tell LOCKSS about any non-AIP (or at least, non-package) files that are accidentally put in LOM Space.  Also, it expects `package` to actually exist (and calls functions on it), so this is a check to ensure it does.

storage_service/locations/models.py
Hwesta, line 545: `Post to Lockss-o-matic with the create resource atom entry` is what the whole function does; this is intended as a section header rather than a description of the next line.  Is there a better way to designate this?

storage_service/common/utils.py
Hwesta: Move `ValueError` checks to the calling function to make this more explicit.

storage_service/locations/api/resources.py
Hwesta, line 423: Both are trying to download the entire AIP (or, in the case of chunks, a pre-split part of the AIP).  `Package.get_download_path` will return the package's full path unless both a chunk number is provided and the package is stored in LOM.  Can you clarify what seems dissimilar about the cases?

PR #9, opened 2014-04-16
Enabled backlogged transfers

Hwesta:
Added backlog purpose to location and config for default location.  POSTing to a Location with a pipeline, origin location, and list of files will move the files from the provided location to the location posted to.  Fixed some problems with moving folders.

Supports AM PR for SIP Arrangement.

### DIFF
--- a/storage_service/common/utils.py
+++ b/storage_service/common/utils.py
@@ -119,7 +119,10 @@ def removedirs(relative_path, base=None):
     Cribbed from the implementation of os.removedirs. """
     if not base:
         return os.removedirs(relative_path)
-    os.rmdir(os.path.join(base, relative_path))
+    try:
+        os.rmdir(os.path.join(base, relative_path))
+    except os.error:
+        pass
     head, tail = os.path.split(relative_path)
     if not tail:
         head, tail = os.path.split(head)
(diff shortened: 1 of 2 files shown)


storage_service/storage_service/urls.py
```diff
@@ -58,6 +58,11 @@ def startup():
         space=space,
         relative_path=os.path.join('var', 'archivematica', 'sharedDirectory', 'www', 'AIPsStore'),
         description='Store AIP in standard Archivematica Directory')
+    backlog, _ = locations_models.Location.objects.get_or_create(
+        purpose=locations_models.Location.BACKLOG,
+        space=space,
+        relative_path=os.path.join('var', 'archivematica', 'sharedDirectory', 'www', 'AIPsStore', 'transferBacklog'),
```
mistydemeo, line 64: The `relative_path` is meant to be relative to `/`?

mistydemeo:
:thumbsup:

storage_service/storage_service/urls.py
Hwesta, line 64: Because the default Space (created line 45) is '/'.

mistydemeo, line 64: Ah, right.

PR #10, opened 2014-04-17
UI-generated Location path is relative to Space

Hwesta:
refs #5608

Javascript that generates the path from browse dialog stores the base path, and removes that from selected path.  Also continues to strip leading '/'s.  Updated URL to not be hard-coded.


mistydemeo:
:thumbsup:

PR #11, opened 2014-04-29
Support storing DIPs via the storage service

mistydemeo:
This creates a new storage purpose, "DIP Storage", which is primarily used to store uncompressed directory DIPs.

Refs #6564

### DIFF
--- a/storage_service/locations/models.py
+++ b/storage_service/locations/models.py
@@ -1040,6 +1040,7 @@ class Location(models.Model):
 
     TRANSFER_SOURCE = 'TS'
     AIP_STORAGE = 'AS'
+    DIP_STORAGE = 'DS'
     # QUARANTINE = 'QU'
     BACKLOG = 'BL'
     CURRENTLY_PROCESSING = 'CP'
@@ -1048,6 +1049,7 @@ class Location(models.Model):
     PURPOSE_CHOICES = (
         (TRANSFER_SOURCE, 'Transfer Source'),
         (AIP_STORAGE, 'AIP Storage'),
+        (DIP_STORAGE, 'DIP Storage'),
         # (QUARANTINE, 'Quarantine'),
         (BACKLOG, 'Transfer Backlog'),
         (CURRENTLY_PROCESSING, 'Currently Processing'),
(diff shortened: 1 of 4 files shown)


storage_service/locations/models.py
```diff
@@ -721,18 +723,20 @@ def store_aip(self, origin_location, origin_path):
 
         # Store AIP Pointer File at
         # internal_usage_location/uuid/split/into/chunks/pointer.xml
-        self.pointer_file_location = Location.active.get(purpose=Location.STORAGE_SERVICE_INTERNAL)
-        self.pointer_file_path = os.path.join(uuid_path, 'pointer.xml')
-        pointer_file_src = os.path.join(self.origin_location.relative_path, os.path.dirname(self.origin_path), 'pointer.xml')
-        pointer_file_dst = self.full_pointer_file_path()
+        if self.package_type in (Package.AIP, Package.AIC):
```
Hwesta, line 1276: Maybe we should add a `PACKAGE_TYPE_HAS_POINTER_FILE` to check against for this sort of thing? 

This may have conflicts with #16

Hwesta:
:+1:

storage_service/locations/models.py
```diff
@@ -754,15 +758,16 @@ def store_aip(self, origin_location, origin_path):
...
-        root = etree.parse(pointer_file_dst)
-        element = root.find('mets:fileSec/mets:fileGrp/mets:file', namespaces=nsmap)
-        flocat = element.find('mets:FLocat', namespaces=nsmap)
-        xlink = 'http://www.w3.org/1999/xlink'
-        if self.uuid in element.get('ID', '') and flocat is not None:
-            flocat.set('{{{ns}}}href'.format(ns=xlink), self.full_path())
-        with open(pointer_file_dst, 'w') as f:
-            f.write(etree.tostring(root, pretty_print=True))
+        if self.package_type in (Package.AIP, Package.AIC):
+            nsmap = {'mets': 'http://www.loc.gov/METS/'}
```
Hwesta: Warning that #8 imports a nsmap instead of defining it here, so you'll probably get conflicts on rebase.

mistydemeo: Thanks! Ran into this with #16 too; seems like they'll slightly conflict here, but it's easy to resolve.

PR #12, opened 2014-05-07
Edit user URL use ID not username

Hwesta:
refs #6674

Django's 'reverse' bails on characters that are not alphanumeric or - (including . and @).  Update user-related URLs to use the ID instead of the username, as some special characters are allowed in the username.


jhsimpson:
I looked at this and I don't see any issues with it.

Hwesta:
Also committed to qa/0.x as 1adc1e647e4c5f963ed050801a769fdf84703a93

PR #13, opened 2014-05-08
Package: add support for checking fixity

mistydemeo:
This teaches Packages how to check their own fixity, and report the results. This is currently only supported for bags (which have a record of their own fixity), e.g. AICs or AIPs.

This also adds a new API call which allows external services to request fixity scans and receive a report of the results. The URL is in the format /file/<uuid>/check_fixity/

Note that this fixity scan isn't intended to be 100% authoritative against tampering (since it depends on a manifest file in the same physical location as the files it documents that is not signed), just against data integrity.

refs #6567

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -15,6 +15,7 @@ from django.core.exceptions import ObjectDoesNotExist, MultipleObjectsReturned
 from django.forms.models import model_to_dict
 
 # Third party dependencies, alphabetical
+import bagit
 from tastypie.authentication import (BasicAuthentication, ApiKeyAuthentication,
     MultiAuthentication, Authentication)
 from tastypie.authorization import DjangoAuthorization, Authorization
@@ -377,6 +378,9 @@ class PackageResource(ModelResource):
 
     api/v1/file/<uuid>/delete_aip/ supports:
     POST: Create a delete request for that AIP.
+
+    Validate fixity (api/v1/file/<uuid>/validate/) supports:
+    GET: Scan package for fixity
     """
     origin_pipeline = fields.ForeignKey(PipelineResource, 'origin_pipeline')
     origin_location = fields.ForeignKey(LocationResource, None, use_in=lambda x: False)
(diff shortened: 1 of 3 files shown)


storage_service/locations/api/resources.py
```diff
@@ -505,3 +510,52 @@ def pointer_file_request(self, request, **kwargs):
         self.log_throttled_access(request)
 
         return response
+
+    def check_fixity_request(self, request, **kwargs):
+        # Tastypie checks
```
Hwesta: Whichever of this PR or #14 that gets committed first should update this to use the custom endpoint decorator.

storage_service/locations/models.py
```diff
@@ -814,6 +830,42 @@ def backlog_transfer(self, origin_location, origin_path):
         self.status = Package.UPLOADED
         self.save()
 
+    def check_fixity(self):
+        """ Scans the package to verify its checksums. This is implemented
```
Hwesta: Descriptive docstring = great!  [PEP257](http://legacy.python.org/dev/peps/pep-0257/#multi-line-docstrings) suggests a one-line summary at the beginning, and then a full explanation.

Hwesta:
Looks good!

Later, we should hook this up to the 'Update Package Status' button coming in #8, and possibly also the Space.verify() stubs.

storage_service/locations/models.py
mistydemeo: This reminds me that I was thinking of documenting the responses that resources return in their docstrings. Didn't get around to adding one here yet.

mistydemeo: Thanks! Updated the docstring.

Not using docstring testing since this requires a real package to test against, though. At some point I'll write a test that tests a tiny mock bag in /tmp or some such.

storage_service/locations/models.py
```diff
@@ -669,6 +670,20 @@ def full_pointer_file_path(self):
             return os.path.join(self.pointer_file_location.full_path(),
                 self.pointer_file_path)
 
+    def is_compressed(self):
+        """ Determines whether or not the package is a compresesd file. """
```
Hwesta: s/compresesd/compressed/

PR #14, opened 2014-05-08
Fixes and infrastructure support

Hwesta:
Several fixes/updates that are useful to multiple dev branches.
- custom endpoint decorator, to reduce boilerplate
- Packages.misc_attributes
- Package.extract_file can extract the entire package

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -11,6 +11,7 @@ import urllib
 
 # Core Django, alphabetical
 from django.conf.urls import url
+from django.core.exceptions import ObjectDoesNotExist, MultipleObjectsReturned
 from django.forms.models import model_to_dict
 
 # Third party dependencies, alphabetical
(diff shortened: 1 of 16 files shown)


storage_service/locations/api/resources.py
```diff
@@ -41,6 +43,56 @@
...
+    def decorator(func):
+        """ The decorator applied to the endpoint """
+        def wrapper(resource, request, **kwargs):
+            """ Wrapper for custom endpoints with boilerplate code. """
+            # Tastypie API checks
+            resource.method_check(request, allowed=expected_methods)
+            resource.is_authenticated(request)
+            resource.throttle_check(request)
+
+            # Get object
```
mistydemeo, line 64: I definitely like standardizing this.

storage_service/locations/migrations/0004_v0_4_0.py
```diff
@@ -109,6 +117,7 @@ def backwards(self, orm):
             'current_location': ('django.db.models.fields.related.ForeignKey', [], {'to': u"orm['locations.Location']", 'to_field': "'uuid'"}),
             'current_path': ('django.db.models.fields.TextField', [], {}),
             u'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'misc_attributes': ('jsonfield.fields.JSONField', [], {'default': '{}', 'null': 'True', 'blank': 'True'}),
```
mistydemeo, line 120: Is it worth adding a comment to the effect of "Don't add anything too important here because you can't query on it"?

Hwesta, line 120: Probably.  Fixed in the model.

mistydemeo:
:sparkling_heart:

PR #15, opened 2014-05-29
Package.extract_file: determine basename using lsar

mistydemeo:
The current code assumes there will only be one file extension. This switches to using lsar to determine the actual directory name. Ideally we'll switch to using lsar's JSON output, but it's currently broken for nested archives.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -14,7 +14,7 @@ X-Python-Version: 2.7
 
 Package: archivematica-storage-service
 Architecture: i386 amd64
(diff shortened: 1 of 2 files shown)


Hwesta:
Looks good.

Maybe add a snippet of sample output from lsar, so I can understand what the string manipulation is doing?

mistydemeo:
This has broken on recent versions of lsar, so I'm going to put effort into getting JSON support working instead.

Hwesta:
:+1:

PR #16, opened 2014-06-03
Store uncompressed AIPs

mistydemeo:
This is the sister PR to artefactual/archivematica#40.

This allows storing uncompressed AIPs via the storage service.
- Store AIP now works if no pointer file is present.
- Packages can now compress themselves, and the download API from the package resource will compress uncompressed packages before serving them.

Ignore the last commit here; it's duplicated from the other branch.

### DIFF
--- a/storage_service/locations/models.py
+++ b/storage_service/locations/models.py
@@ -1205,6 +1205,8 @@ class Package(models.Model):
     def get_download_path(self, lockss_au_number=None):
         full_path = self.full_path()
         if lockss_au_number is None:
+            if not self.is_compressed():
+                raise StorageException("Cannot return a download path for an uncompressed package")
             path = full_path
         elif self.current_location.space.access_protocol == Space.LOM:
             # Only LOCKSS breaks files into AUs
--- a/storage_service/locations/models.py
+++ b/storage_service/locations/models.py
@@ -1285,8 +1285,14 @@ class Package(models.Model):
         # Move pointer file
         if self.package_type in (Package.AIP, Package.AIC):
             pointer_file_name = 'pointer-'+self.uuid+'.xml'
-            src_space.move_to_storage_service(pointer_file_src, pointer_file_name, self.pointer_file_location.space)
-            self.pointer_file_location.space.move_from_storage_service(pointer_file_name, pointer_file_dst)
+            try:
+                src_space.move_to_storage_service(pointer_file_src, pointer_file_name, self.pointer_file_location.space)
+                self.pointer_file_location.space.move_from_storage_service(pointer_file_name, pointer_file_dst)
+            except:
+                logging.warning("No pointer file found")
+                self.pointer_file_location = None
+                self.pointer_file_path = None
+                self.save()
 
         # Move AIP
         src_space.move_to_storage_service(
@@ -1312,7 +1318,7 @@ class Package(models.Model):
         self.save()
 
         # Update pointer file's location information
-        if self.package_type in (Package.AIP, Package.AIC):
+        if self.pointer_file_path and self.package_type in (Package.AIP, Package.AIC):
             root = etree.parse(pointer_file_dst)
             element = root.find('.//mets:file', namespaces=utils.NSMAP)
             flocat = element.find('mets:FLocat', namespaces=utils.NSMAP)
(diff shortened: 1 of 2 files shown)


storage_service/locations/models.py
```diff
@@ -1281,8 +1285,14 @@ def store_aip(self, origin_location, origin_path):
 
         # Move pointer file
         pointer_file_name = 'pointer-'+self.uuid+'.xml'
-        src_space.move_to_storage_service(pointer_file_src, pointer_file_name, self.pointer_file_location.space)
-        self.pointer_file_location.space.move_from_storage_service(pointer_file_name, pointer_file_dst)
+        try:
+            src_space.move_to_storage_service(pointer_file_src, pointer_file_name, self.pointer_file_location.space)
+            self.pointer_file_location.space.move_from_storage_service(pointer_file_name, pointer_file_dst)
+        except:
+            logging.warn("No pointer file found")
```
Hwesta: `logging.warn` has been [deprecated by removing docs about it](http://bugs.python.org/issue13235), and replaced by `logging.warning`

mistydemeo: Thanks, fixed.

storage_service/locations/models.py
```diff
@@ -1203,6 +1203,10 @@ def is_compressed(self):
     def get_download_path(self, lockss_au_number=None):
         full_path = self.full_path()
         if lockss_au_number is None:
+            if not self.is_compressed():
+                path, _ = self.compress_package()
```
Hwesta: The temp dir this is compressed to should be deleted.  If the temp dir path can make it to `utils.download_file_stream`'s temp_dir parameter, that will delete it.

mistydemeo: Decided I didn't like pushing this into `get_download_path` anyway (I definitely wouldn't expect `get_download_path` to churn away on 500 gigs of data!), so made it raise an exception in that case and had the resource handle compressing the file and passing the temp dir as appropriate.

storage_service/locations/models.py
```diff
@@ -1180,7 +1180,7 @@ def full_pointer_file_path(self):
         """ Return the full path of the AIP's pointer file, None if not an AIP.
 
         Includes the space, location and package paths joined."""
-        if self.package_type not in (self.AIP, self.AIC):
+        if not self.pointer_file_location:
```
Hwesta, line 1185: `resources.PackageResource.pointer_file_request` should be updated to handle the path being None, since it's also called from AM (archival storage tab).

mistydemeo, line 1185: Pushed a commit to cause it to return 404 if no pointer file exists.

storage_service/locations/models.py
```diff
@@ -1376,6 +1385,39 @@ def extract_file(self, relative_path='', extract_path=None):
...
+        full_path = self.full_path()
+        basename = os.path.splitext(os.path.basename(full_path))[0]
+        compressed_filename = os.path.join(extract_path, basename + '.tar')
+
+        command = [
+            "tar", "-cf", compressed_filename,
+            # The filenames are transformed to ensure only the
+            # directory basename is present, not the entire path.
+            # FIXME This syntax isn't portable and is only compatible with GNU tar
+            "--xform", "s,{}/,,".format(os.path.dirname(full_path[1:-1])),
```
Hwesta: As mentioned in person, this can be simplified to `command = ['tar', 'c', '-C', relative_path, '-f', compressed_filename, basename]` where `relative_path = ''.join(full_path.rsplit(basename, 1))` (replace once, but starting from the right).  You can update here, or I can update in the AIP reingest branch.

storage_service/locations/models.py
```diff
@@ -1203,6 +1203,8 @@ def is_compressed(self):
     def get_download_path(self, lockss_au_number=None):
         full_path = self.full_path()
         if lockss_au_number is None:
+            if self.is_compressed():
```
Hwesta: Should be `if not self.is_compressed():`

Hwesta:
`PackageResource.extract_file_request` needs to be updated as well (though I think the problem is actually that extract_file doesn't append the relative_path instead of the basename if the AIP is uncompressed).

mistydemeo:
Updated branch with a fix to extract_file_request to adjust the path, and a couple of other commits to return correct status codes and useful error messages in failure cases.

Hwesta:
:+1:

PR #17, opened 2014-06-12
Add email notification for deletion requests, fixity failures

mistydemeo:
This adds email notifications to a couple of events: AIP deletion requests, and fixity check failures.

The storage service will send emails to every user account with superuser privileges, are the accounts with permission to do things like approve AIP deletion. It's implemented using Django signals, and a new `signals` module had been added to `locations`.

It uses Django's email support; in order for it to work, the email server and authentication information to use must be configured in Django's settings files (or in environment variables). If email isn't configured, then these signals won't send an email (they'll print the message they would have sent to stdout instead).

Django is not configured with a way to get the site's URL by default; in order for the AIP deletion request to provide a URL, this must be configured with the SITE_BASE_URL config setting, with a value similar to "http://example.com". If that parameter is omitted, then the AIP deletion request email won't include a link to the page where requests can be approved or denied.

### DIFF
--- a/storage_service/locations/signals.py
+++ b/storage_service/locations/signals.py
@@ -1,0 +1,50 @@
+import logging
+from django.dispatch import receiver, Signal
+from django.contrib.auth.models import User
+from django.core.urlresolvers import reverse
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(filename="/tmp/storage_service.log",
+                    level=logging.INFO)
+
+
+deletion_request = Signal(providing_args=["uuid", "location", "url"])
+failed_fixity_check = Signal(providing_args=["uuid", "location", "report"])
+
+
+def _notify_administrators(subject, message):
+    admin_users = User.objects.filter(is_superuser=True)
+    for user in admin_users:
+        try:
+            user.email_user(subject, message)
+        except Exception:
+            logging.exception("Unable to send email to %s", user.email)
+
+
+@receiver(deletion_request, dispatch_uid="deletion_request")
+def report_deletion_request(sender, **kwargs):
+    subject = "Deletion request for package {}".format(kwargs["uuid"])
+    message = """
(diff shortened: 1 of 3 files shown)


storage_service/locations/signals.py
```diff
@@ -0,0 +1,49 @@
...
+failed_fixity_check = Signal(providing_args=["uuid", "location", "report"])
+
+
+def _notify_administrators(subject, message):
+    admin_users = User.objects.filter(is_superuser=True)
+    for user in admin_users:
+        try:
+            user.email_user(subject, message)
+        except Exception as e:
+            logging.error("Unable to send email to {}; failed with {} ({})".format(user.email, type(e).__name__, e))
```
Hwesta: `logging.exception` logs an error-level message and automatically attaches the exception information. :)

storage_service/locations/signals.py
```diff
@@ -0,0 +1,49 @@
...
+    subject = "Deletion request for package {}".format(kwargs["uuid"])
+    message = """
+A package deletion request was received for the package with UUID {}. This package is currently stored at: {}.
+""".format(kwargs["uuid"], kwargs["location"])
+
+    # The URL may not be configured in the site; if it isn't,
+    # don't try to tell the user the URL to approve/deny the request.
+    if kwargs["url"]:
+        message = message + """
+To approve this deletion request, visit: {}/packages/aip_delete_request
```
Hwesta: Use `django.core.urlresolvers.reverse('aip_delete_request')` to get the delete request URL, in case it changes later?

Hwesta:
Looks good

storage_service/locations/signals.py
mistydemeo: Good call - switched to using that.

mistydemeo:
Updated.

PR #18, opened 2014-06-12
Make internal processing location configurable

mistydemeo:
The only location in the storage service that is not user-configurable is the internal processing location. This ended up causing a problem at one client where the machine the SS is installed on has limited hard drive space, and rapidly ran out of room due almost entirely to the contents of this location. There were two issues trying to change this:
- The location type isn't available in the menu, so even though there is an option to edit this location, it couldn't be saved because its own type wasn't available in the dropdown
- On every startup, a new location with the default path is created if none exists, so trying to move the original location would result in a second location at the original path appearing

### DIFF
--- a/storage_service/storage_service/urls.py
+++ b/storage_service/storage_service/urls.py
@@ -44,8 +44,9 @@ def startup():
     logging.info("Running startup")
     space, space_created = locations_models.Space.objects.get_or_create(
         access_protocol=locations_models.Space.LOCAL_FILESYSTEM,
-        path=os.sep,
-        staging_path=os.path.join(os.sep, 'var', 'archivematica', 'storage_service'))
+        path=os.sep, defaults={
+            "staging_path": os.path.join(os.sep, 'var', 'archivematica', 'storage_service')
+        })
     if space_created:
         local_fs = locations_models.LocalFilesystem(space=space)
         local_fs.save()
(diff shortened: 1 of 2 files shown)


storage_service/storage_service/urls.py
```diff
@@ -68,16 +68,19 @@ def startup():
...
-        purpose=locations_models.Location.STORAGE_SERVICE_INTERNAL,
-        space=space,
-        relative_path=os.path.join('var', 'archivematica', 'storage_service'),
-        description='For storage service internal usage.')
-    try:
-        os.mkdir(internal_use.full_path())
-    except OSError as e:
-        if e.errno != errno.EEXIST:
-            logging.error("Internal storage location {} not accessible.".format(internal_use.full_path()))
+    internal_use, created = locations_models.Location.objects.get_or_create(
```
Hwesta, line 72: Setting values only if created should use `defaults` as a parameter to `get_or_create` https://docs.djangoproject.com/en/dev/ref/models/querysets/#get-or-create

mistydemeo, line 72: Updated.

storage_service/storage_service/urls.py
```diff
@@ -68,16 +68,20 @@ def startup():
...
-    except OSError as e:
-        if e.errno != errno.EEXIST:
-            logging.error("Internal storage location {} not accessible.".format(internal_use.full_path()))
+        defaults={
+            'space': space,
+            'relative_path': os.path.join('var', 'archivematica', 'storage_service'),
+            'description': 'For storage service internal usage.'
+        }
+    )
+    if created:
```
Hwesta, line 80: Do we still need `if created` around mkdir?

storage_service/storage_service/urls.py
```diff
@@ -44,9 +44,9 @@ def startup():
     logging.info("Running startup")
     space, space_created = locations_models.Space.objects.get_or_create(
         access_protocol=locations_models.Space.LOCAL_FILESYSTEM,
-        path=os.sep,
-        staging_path=os.path.join(os.sep, 'var', 'archivematica', 'storage_service'))
+        path=os.sep)
     if space_created:
+        space.staging_path = os.path.join(os.sep, 'var', 'archivematica', 'storage_service')
```
Hwesta: `defaults` here too?

storage_service/storage_service/urls.py
mistydemeo, line 80: I felt in this case it made sense for it to not opportunistically create a non-default space; I feel like the administrator is responsible for making sure the path exists (and if they fat-finger it, it's probably more confusing if the path is created for them and functions).

Hwesta, line 80: That's reasonable

Hwesta:
:+1:

PR #21, opened 2014-06-16
Dev/issue 6596 aip restore v3

mcantelon:


storage_service/administration/forms.py
```diff
@@ -104,6 +107,14 @@ class DefaultLocationsForm(SettingsForm):
     new_backlog = DefaultLocationField(
         required=False,
         label="New Transfer Backlog:")
+    default_recovery = forms.MultipleChoiceField(
+        choices=[],
+        required=False,
+        label="Default AIP recovery locations for new pipelines:"
+        )
+    new_backlog = DefaultLocationField(
```
mistydemeo: Wrong variable name? This is shadowing the previous definition of `new_backlog`.

storage_service/administration/forms.py
```diff
@@ -78,6 +78,9 @@ class CommonSettingsForm(SettingsForm):
     """ Configures common or generic settings that don't belong elsewhere. """
     pipelines_disabled = forms.BooleanField(required=False,
         label="Pipelines are disabled upon creation?")
+    recover_location_uuid = forms.CharField()
```
mistydemeo: Instead of requiring the user to set a UUID, how about making the AIP recovery location per-pipeline, and setting it as a default the way that transfer source, AIP storage, etc. gets set below?

storage_service/locations/api/resources.py
```diff
@@ -449,33 +450,25 @@ def delete_aip_request(self, request, bundle, **kwargs):
...
-                'message': 'Delete request created successfully.'
-            }
+    @_custom_endpoint(expected_methods=['post'],
+        required_fields=('event_reason', 'pipeline', 'user_id', 'user_email'))
+    def recover_aip_request(self, request, bundle, **kwargs):
+        request_info = bundle.data
+        package = bundle.obj
+        if package.package_type not in Package.PACKAGE_TYPE_CAN_RECOVER:
+            # Can only request recovery of AIPs
+            return http.HttpMethodNotAllowed()
```
mistydemeo: Include a message here?

storage_service/locations/models.py
```diff
@@ -1244,6 +1248,83 @@ def _update_quotas(self, space, location):
...
+            temp_aip.delete()
+            return (success, failures, message)
+
+        origin_space = temp_aip.current_location.space
+        destination_space = self.current_location.space
+
+        # Copy corrupt files to storage service staging
+        backup_path = utils.get_setting('recover_backup_path_within_location')
+
+        if backup_path == None:
```
mistydemeo: `is None`

storage_service/locations/models.py
```diff
@@ -1244,6 +1248,83 @@ def _update_quotas(self, space, location):
...
+            logging.warn("No AIP restore backup directory found, skipping backup")
+        else:
+            source_path = os.path.join(
+                self.current_location.relative_path,
+                self.current_path)
+            destination_path = os.path.join(
+                origin_location.relative_path,
+                utils.get_setting('recover_backup_path_within_location'))
```
mistydemeo: Use the variable assigned above?

storage_service/locations/views.py
```diff
@@ -44,9 +45,49 @@ def package_list(request):
     packages = Package.objects.all()
     return render(request, 'locations/package_list.html', locals())
 
+def aip_recover_request(request):
+    if request.method == 'GET' and utils.get_setting('recover_backup_path_within_location') == None:
```
mistydemeo: `is None` here too

storage_service/locations/views.py
```diff
@@ -44,9 +45,49 @@ def package_list(request):
...
+        messages.error(request, 'Recover backup path not set.')
+
+    def execution_logic(aip): 
+        recover_location = LocationPipeline.objects.filter(
```
mistydemeo: Do you expect more than one recovery location to be defined for a pipeline?

storage_service/locations/views.py
```diff
@@ -44,9 +45,49 @@ def package_list(request):
     packages = Package.objects.all()
     return render(request, 'locations/package_list.html', locals())
 
+def aip_recover_request(request):
+    if request.method == 'GET' and utils.get_setting('recover_backup_path_within_location') == None:
+        messages.error(request, 'Recover backup path not set.')
+
+    def execution_logic(aip):
```
mistydemeo, line 52: I like this refactoring of the base request logic!

storage_service/storage_service/urls.py
```diff
@@ -63,6 +63,11 @@ def startup():
         space=space,
         relative_path=os.path.join('var', 'archivematica', 'sharedDirectory', 'www', 'AIPsStore', 'transferBacklog'),
         description='Default transfer backlog')
+    recovery, _ = locations_models.Location.objects.get_or_create(
+        purpose=locations_models.Location.AIP_RECOVERY,
+        space=space,
+        relative_path=os.path.join('var', 'archivematica', 'storage_service', 'recover'),
```
mistydemeo: I'm moving away from hardcoding these paths; see #18. (This will create a new default space with the hardcoded path if the user changes the path of the existing one.) Mind changing it to be like the ones in that other PR?

mistydemeo:
How do the good files make it into the AIP recover path? Is that manual?

mcantelon:
Yeah, good files have to be manually copied to the recover path.

storage_service/locations/api/resources.py
```diff
@@ -447,35 +448,35 @@ def delete_aip_request(self, request, bundle, **kwargs):
         package = bundle.obj
         if package.package_type not in Package.PACKAGE_TYPE_CAN_DELETE:
             # Can only request deletion on AIPs
-            return http.HttpMethodNotAllowed()
+            response = {
+                "message": "Deletes not allowed on this package type."
+            }
+            response_json = json.dumps(response)
+            return http.HttpMethodNotAllowed(response_json)
```
mistydemeo: Can you specify the `content_type` too?

storage_service/locations/views.py
```diff
@@ -44,9 +45,43 @@ def package_list(request):
     packages = Package.objects.all()
     return render(request, 'locations/package_list.html', locals())
 
+def aip_recover_request(request):
+    def execution_logic(aip): 
+        recover_location = LocationPipeline.objects.get(
+            pipeline=aip.origin_pipeline, location__purpose='AR').location
```
mistydemeo: Just to double-check, do we have a constraint that prevents more than one recovery location being set for a pipeline?

mistydemeo:
:thumbsup:

mistydemeo:
Oops, just realized that I accidentally closed this because I deleted the branch it's branched from.

mistydemeo:
Recreated the branch.

storage_service/locations/views.py
mcantelon: Hrm, I don't think we do... I'll have to take a look at that.

mistydemeo: Okay! This'll raise when the `.get` occurs, which would be confusing/inconvenient, so it seems like it'd make sense to avoid the multiple locations happening in the first place.

mistydemeo: (Raise if there's more or less than one recovery location in the space, that is.)

mcantelon: Okay, I've added some logic in the admin interface to prevent creation of more than one recover location per pipeline.

storage_service/locations/api/resources.py
```diff
@@ -449,40 +450,35 @@ def delete_aip_request(self, request, bundle, **kwargs):
...
             response = {
-                'message': 'Delete request created successfully.'
+                "message": "Recovery not allowed on this package type."
             }
-
             response_json = json.dumps(response)
-            status_code = 202
+            return http.HttpMethodNotAllowed(response_json, content_type='application/json')
 
-            # This isn't configured by default
```
mistydemeo, line 507: This whole block from the AIP deletion request path should be moved to the new AIP deletion request instead of deleted.

mistydemeo:
Added a new commit to fix an issue that can occurred when the AIP recovery location was altered from the defaults.

PR #22, opened 2014-06-28
Add new post-store callback API

mistydemeo:
This adds support for post-store callbacks, via a new API that would be called by Archivematica. The current usecase is to support reporting back to Islandora for each file that's stored.

The storage service currently doesn't maintain a persistent memory of files; this adds a new minimal File model that can be used to track files on transfer start. Since we don't currently assign or track UUIDs to individual files in the storage service, this instead calculates checksums on arrival and uses that to identify the file at a later point in time. This is kind of a hack.

(In the long term, I think it'd make sense to do some of the file cleanup and UUID assignment in the storage service, and send the files over to Archivematica with UUIDs assigned and the transfer METS already built; then we can just inspect the AIP METS to locate files. That would be best done when we have a) a background task scheduler, b) a standalone METS writer, though.)

The settings for the callback are global; may make more sense to make them controlled per-pipeline.

Another few supporting improvements:
- The SWORD helpers checksum function has been replaced with a checksum function in common.utils that uses hashlib, and supports any checksum algorithm that hashlib supports. File's checksum column was sized large enough to fit a whirlpool or sha512 hash<s>,though that's bigger than we'd actually use for the time being</s>. UPDATE: Storing sha512 now.
- Package gained a new instance method, `Package.each_file`, that iterates over each file in a package as a convenience wrapper around `os.walk`. It also supports extracting compressed packages if necessary.

This isn't tested against an actual Islandora instance yet.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -7,6 +7,7 @@ South==0.8.4
 django-tastypie==0.9.15
 django-extensions==1.1.1
 lxml==3.2.3
+requests==2.2.1
 six==1.4.1
 django-jsonfield==0.9.12
 bagit==1.3.7
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -5,6 +5,7 @@
 import json
 import logging
 import os
+import shutil
 import urllib
 
 # Core Django, alphabetical
@@ -30,7 +31,7 @@ from tastypie.utils import trailing_slash
 from common import utils
 from locations.api.sword import views as sword_views
 
-from ..models import (Event, Package, Location, Space, Pipeline, StorageException)
+from ..models import (Callback, CallbackError, Event, File, Package, Location, Space, Pipeline, StorageException)
 from ..forms import LocationForm, SpaceForm
 from ..constants import PROTOCOL
 from locations import signals
@@ -447,6 +448,7 @@ class PackageResource(ModelResource):
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/download%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('download_request'), name="download_request"),
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/pointer_file%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('pointer_file_request'), name="pointer_file_request"),
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/check_fixity%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('check_fixity_request'), name="check_fixity_request"),
+            url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/send_callback/post_store%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('aip_store_callback_request'), name="aip_store_callback_request"),
 
             # FEDORA/SWORD2 endpoints
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/sword%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('sword_deposit'), name="sword_deposit"),
@@ -632,6 +634,79 @@ class PackageResource(ModelResource):
             mimetype="application/json"
         )
         return response
+
+    @_custom_endpoint(expected_methods=['get'])
+    def aip_store_callback_request(self, request, bundle, **kwargs):
+        package = bundle.obj
+
+        callbacks = Callback.objects.filter(event="post_store")
+        if len(callbacks) == 0:
+            return http.HttpNoContent()
+
+        fail = 0
+
+        if package.is_compressed:
+            # Don't extract the entire AIP, which could take forever;
+            # instead, just extract bagit.txt and manifest-sha512.txt,
+            # which is enough to get bag.entries with the
+            # precalculated sha512 checksums
+            try:
+                basedir = package.get_base_directory()
+            # Currently we only support this for local packages.
+            except NotImplementedError:
+                return http.HttpNoContent()
+
+            _, tmpdir = package.extract_file(os.path.join(basedir, 'bagit.txt'))
+            package.extract_file(os.path.join(basedir, 'manifest-sha512.txt'),
+                extract_path=tmpdir)
+
+            package_dir = os.path.join(tmpdir, basedir)
+        else:
+            package_dir = package.full_path()
+            tmpdir = None
+
+        safe_files = ('bag-info.txt', 'manifest-sha512.txt', 'bagit.txt')
+
+        bag = bagit.Bag(package_dir)
+        for f, checksums in bag.entries.iteritems():
+            try:
+                cksum = checksums['sha512']
+            except KeyError:
+                # These files do not typically have an sha512 hash, so it's
+                # fine for these to be missing that key; every other file should.
+                if f not in safe_files:
+                    logging.warning("Post-store callback: sha512 missing for file %s", f)
+                continue
+
+            files = File.objects.filter(checksum=cksum, stored=False)
+            if len(files) > 1:
+                logging.warning("Multiple File entries found for sha512 {}".format(cksum))
+
+            for file_ in files:
+                for callback in callbacks:
+                    uri = callback.uri.replace('<source_id>', file_.source_id)
+                    try:
+                        callback.execute(uri)
+                        file_.stored = True
+                        file_.save()
+                    except CallbackError:
+                        fail += 1
+
+        if tmpdir is not None:
+            shutil.rmtree(tmpdir)
+
+        if fail > 0:
+            response = {
+                "message": "Failed to POST {} responses to callback URI".format(fail),
+                "failure_count": fail,
+                "callback_uris": [c.uri for c in callbacks]
+            }
+            return http.ApplicationError(
+                json.dumps(response),
+                mimetype="application/json"
+            )
+        else:
+            return http.HttpNoContent()
 
     def sword_deposit(self, request, **kwargs):
         package = get_object_or_None(Package, uuid=kwargs['uuid'])
--- a/storage_service/locations/views.py
+++ b/storage_service/locations/views.py
@@ -11,7 +11,7 @@ from django.views.decorators.csrf import csrf_exempt
 
 from common import decorators
 from common import utils
-from .models import Space, Location, Package, Event, Pipeline, LocationPipeline
+from .models import Callback, Space, Location, Package, Event, Pipeline, LocationPipeline
 from . import forms
 from .constants import PROTOCOL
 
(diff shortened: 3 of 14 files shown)


storage_service/locations/api/resources.py
```diff
@@ -524,6 +526,63 @@ def pointer_file_request(self, request, **kwargs):
...
+            }
+            return http.HttpBadRequest(
+                json.dumps(response),
+                mimetype="application/json"
+            )
+
+        fail = 0
+
+        for f in package.each_file(extract=True):
+            cksum = utils.get_file_checksum(f, 'sha256')
```
mistydemeo: Can probably fetch this from the bag's manifest if it exists, and only fall back to calculating it if not.

mistydemeo: We're storing sha512; I'll switch to calculating those on creation and request file info from the bag.

mistydemeo:
Since I'm no longer using it, may delete `Package.each_file`.

storage_service/common/utils.py
```diff
@@ -102,6 +103,21 @@ def download_file_stream(filepath, temp_dir=None):
 
 ############ OTHER ############
 
+def get_file_checksum(filepath, algorithm):
```
Hwesta: This was also added in d03e4e59dad82fd810b058ce7cd91efde2bc2c19, which returns the checksum object instead of the hexdigest.  One will get replaced when this gets rebased.

storage_service/locations/api/resources.py
```diff
@@ -524,6 +528,79 @@ def pointer_file_request(self, request, **kwargs):
 
         return response
 
+    def aip_store_callback_request(self, request, **kwargs):
+        package = get_object_or_None(Package, uuid=kwargs['uuid'])
+        if package and package.package_type != Package.AIP:
```
Hwesta: What if package is None?

storage_service/locations/api/resources.py
```diff
@@ -524,6 +528,79 @@ def pointer_file_request(self, request, **kwargs):
 
         return response
 
+    def aip_store_callback_request(self, request, **kwargs):
```
Hwesta: Should be updated to use @_custom_endpoint decorator, once rebased (which will also handle the case of the Package being None)

storage_service/locations/api/resources.py
```diff
@@ -524,6 +528,79 @@ def pointer_file_request(self, request, **kwargs):
...
+                    continue
+            try:
+                source_id = File.objects.get(checksum=cksum).source_id
+            except File.DoesNotExist:
+                continue
+
+            uri = base_callback_uri.replace('<source_id>', source_id)
+            try:
+                requests.post(uri)
+            except:
```
Hwesta: Catch a more specific exception (even just Exception).  Maybe log,  `logging.info(..., exc_info=True)`?

storage_service/locations/api/resources.py
mistydemeo: Updated to use the request endpoint instead. (Didn't exist where this was originally branched from.)

mistydemeo:
Rebased and reworked on the Fedora class.

Now introduces a new `Callback` class, which allows for generic callbacks to be associated with given types of events. The only such event right now is "post-store", but this could be used other places. For instance, AIP reingest could use this instead of using its custom callback.

This also accommodates multiple callbacks for a given event, so it's possible to POST reports to multiple services.

storage_service/locations/api/resources.py
```diff
@@ -633,6 +635,61 @@ def check_fixity_request(self, request, bundle, **kwargs):
...
+        package = bundle.obj
+
+        callbacks = Callback.objects.filter(event="post_store")
+        if len(callbacks) == 0:
+            return http.HttpNoContent()
+
+        fail = 0
+
+        if package.is_compressed():
+            package_dir, tmpdir = package.extract_file()
```
mistydemeo: Possible optimization: see if I can extract just the bag service files and not the bag contents for a quicker way to access the bag properties from the manifest.

storage_service/locations/api/resources.py
```diff
@@ -633,6 +635,61 @@ def check_fixity_request(self, request, bundle, **kwargs):
...
+                except CallbackError:
+                    fail += 1
+
+        if tmpdir is not None:
+            shutil.rmtree(tmpdir)
+
+        if fail > 0:
+            response = {
+                "message": "Failed to POST {} responses to callback URI".format(fail),
+                "failure_count": fail,
```
mistydemeo, line 701: Now that there's multiple URLs, failures should probably be more specific, e.g. a list of:

``` python
{
    "file": "filename",
    "callback_uri": "uri"
}
```

mistydemeo:
Updated with forms to create new callbacks, and a migration for callbacks.

storage_service/locations/api/resources.py
```diff
@@ -633,6 +635,61 @@ def check_fixity_request(self, request, bundle, **kwargs):
...
+            tmpdir = None
+
+        bag = bagit.Bag(package_dir)
+        for f, checksums in bag.entries.iteritems():
+            try:
+                cksum = checksums['sha512']
+            except KeyError:
+                # These files do not typically have an sha512 hash, so it's
+                # fine for these to be missing that key; every other file should.
+                safe_files = ('bag-info.txt', 'manifest-sha512.txt', 'bagit.txt')
```
Hwesta: Define this outside the loop?

storage_service/locations/api/resources.py
```diff
@@ -633,6 +635,61 @@ def check_fixity_request(self, request, bundle, **kwargs):
...
+                if f not in safe_files:
+                    logging.warning("Post-store callback: sha512 missing for file %s", f)
+                    continue
```
Hwesta: What if `f in safe_files`?  Won't `cksum` be undefined below?

storage_service/locations/api/resources.py
```diff
@@ -633,6 +635,61 @@ def check_fixity_request(self, request, bundle, **kwargs):
...
+            try:
+                source_id = File.objects.get(checksum=cksum).source_id
+            except File.DoesNotExist:
```
Hwesta: Catch `File.MultipleObjectsReturned` as well?

storage_service/locations/api/resources.py
```diff
@@ -633,6 +635,61 @@ def check_fixity_request(self, request, bundle, **kwargs):
...
+        if package.is_compressed():
+            package_dir, tmpdir = package.extract_file()
+        else:
+            package_dir = package.full_path()
+            tmpdir = None
+
+        bag = bagit.Bag(package_dir)
+        for f, checksums in bag.entries.iteritems():
+            try:
+                cksum = checksums['sha512']
```
Hwesta, line 673: Nice trick, using the existing bagit checksums instead of recalculating.

requirements/base.txt
```diff
@@ -7,6 +7,7 @@ South==0.8.4
 django-tastypie==0.9.15
 django-extensions==1.1.1
 lxml==3.2.3
+requests==2.2.1
```
mistydemeo, line 10: Should we look at upgrading to requests 2.3.1 across the board? I used 2.2.1 because another branch is using the same version.

storage_service/locations/api/resources.py
mistydemeo: Good point.

storage_service/locations/api/resources.py
mistydemeo: This isn't a subloop - it'll continue to processing the next file, and won't try to do anything with this one.

storage_service/locations/api/resources.py
mistydemeo: Good point!

Hwesta:
Minor: The commit order of adding models (File, Callback) should come before they're used.

Do we want to merge the Callback and File migrations?  They'll be coming in together, and are conceptually one feature.  To keep individual commits working, the later migration could use --update.

storage_service/locations/views.py
```diff
@@ -332,3 +332,47 @@ def space_delete(request, uuid):
...
 
+########################## CALLBACKS ##########################
+
+def callback_detail(request, uuid):
+    try:
+        callback = Callback.objects.get(uuid=uuid)
+    except Callback.DoesNotExist:
+        messages.warning(request, "Callback {} does not exist.".format(location_uuid))
+        return redirect('callback_list')
+    return render(request, 'locations/callback_detail.html', locals())
```
Hwesta, line 343: I don't see the templates committed...

mistydemeo:
Committed callback templates - looks like those got missed!

mistydemeo:
> Do we want to merge the Callback and File migrations?

Makes sense, I think. Planning to merge those together when this branch gets merged; right now they're separate just because they represent the order in which I wrote features in the branch.

mistydemeo:
> Minor: The commit order of adding models (File, Callback) should come before they're used.

Github is showing the incorrect order. The order is:
- sword/helpers md5
- callback model
- file model
- commits those use those models

requirements/base.txt
Hwesta, line 10: Sure - do we want to just specify `requests>=2.2.1?`?

Hwesta:
> > Minor: The commit order of adding models (File, Callback) should come before they're used.
> 
> Github is showing the incorrect order. The order is:
>    sword/helpers md5
>    callback model
>    file model
>    commits those use those models
> Okay

storage_service/locations/api/resources.py
Hwesta: I'm thinking this case:

``` python
for f, checksums in bag.entries.iteritems():
    try:
        cksum = checksums['sha512']  # Raises exception
    except KeyError:
        safe_files = ('bag-info.txt', 'manifest-sha512.txt', 'bagit.txt')

        if f not in safe_files:  # f is in safe_files, does not trigger this
            logging.warning("Post-store callback: sha512 missing for file %s", f)
            continue
    try:
        # Continues to here - cksum not defined because exception raised,
        # continue not hit because f in safe_files
        source_id = File.objects.get(checksum=cksum).source_id
```

storage_service/locations/views.py
```diff
@@ -332,3 +332,47 @@ def space_delete(request, uuid):
...
+
+def callback_switch_enabled(request, uuid):
+    callback = get_object_or_404(Callback, uuid=uuid)
+    callback.enabled = not callback.enabled
```
Hwesta, line 347: Are [F()](https://docs.djangoproject.com/en/dev/ref/models/queries/#f-expressions) expressions overkill here? Would prevent race conditions.  (Discovered them recently, which is why they don't show up elsewhere...)

storage_service/locations/views.py
```diff
@@ -332,3 +332,47 @@ def space_delete(request, uuid):
...
+
+    if request.method == "POST":
+        form = forms.CallbackForm(request.POST or None, instance=callback)
+        if form.is_valid():
+            callback = form.save()
+            messages.success(request, "Callback saved.")
+            return redirect('callback_detail', callback.uuid)
+    else:
+        form = forms.CallbackForm(instance=callback)
+        return render(request, 'locations/callback_form.html', locals())
```
Hwesta: The previous 9 lines can be simplified to

``` python
form = forms.CallbackForm(request.POST or None, instance=callback)
if form.is_valid():
    callback = form.save()
    messages.success(request, "Callback saved.")
    return redirect('callback_detail', callback.uuid)
return render(request, 'locations/callback_form.html', locals())
```

which I think is a really neat trick.

Hwesta:
Currently we only checksum and add files to the Files table for FEDORA stuff.  That means if someone creates an AIP not using the SWORD API, and sets up a callback, it will silently fail trying to match files, correct? Is this a problem? Should we try to support more general usecases, since the callback structure is more generic?  Post a warning that this is the case?

requirements/base.txt
mistydemeo, line 10: :thumbsup:

storage_service/locations/api/resources.py
Hwesta: extract_file takes a parameter for the relative path within the package - you could use that to just get the service file(s)

storage_service/locations/api/resources.py
```diff
@@ -633,6 +635,65 @@ def check_fixity_request(self, request, bundle, **kwargs):
         )
         return response
 
+    @_custom_endpoint(expected_methods=['get'])
+    def aip_store_callback_request(self, request, bundle, **kwargs):
+        package = bundle.obj
+
+        callbacks = Callback.objects.filter(event="post_store")
```
Hwesta, line 642: Might make sense to have the type of event included in the callback, so we can reuse this endpoint. Ie, instead of calling `send_store_callback` and having the endpoint know,  do `send_callback/post_store/` or `send_callback` with the body having {"type": "post_store"}, so it's more generic.

storage_service/locations/migrations/0006_auto__add_callback.py
```diff
@@ -0,0 +1,220 @@
+# -*- coding: utf-8 -*-
```
Hwesta: Merge the two migrations into one file and give it a more meaningful name?

mistydemeo: Merged into `0005_v0_5_0.py`.

storage_service/locations/api/resources.py
mistydemeo, line 642: Changed the URL to `send_callback/post_store/`, but the logic is too tightly coupled to the post-store AIP usecase to genericize the function logic.

storage_service/locations/views.py
mistydemeo, line 347: The logic here is copied from other places, so if we change them to use `F()` expressions let's do it out of this PR.

mistydemeo:
Updated the post-store callback to not extract the entire AIP, and added a new function in 118d9bafd56fda66b592040b96559a43f53173d5 to get the directory basename.

Going to test this to make sure it actually _works_, given the number of changes and rebasing.

Issue #1, opened 2014-01-10
Debian postinst script still not robust


BHSPitMonkey:
After having tried to install 0.2.2ppa8.deb yesterday (before the fixes were pushed out), I attempted to apt-get purge the package and try it again today (ppa11). With some parts of the filesystem not left exactly as the script expects, it fails unnecessarily:

```
Setting up archivematica-storage-service (0.2.2ppa11) ...
creating archivematica user
User archivematica exists
creating django secret key
[redacted]
creating symlink in /usr/lib/archivematica
ln: failed to create symbolic link `/usr/lib/archivematica/storage-service/storage_service': File exists
mv: cannot move `/var/archivematica/storage-service/static' to `/usr/lib/archivematica/storage-service/static/static': Directory not empty
mv: cannot move `/var/archivematica/storage-service/templates' to `/usr/lib/archivematica/storage-service/templates/templates': Directory not empty
configuring django database and static files
/var/lib/dpkg/info/archivematica-storage-service.postinst: 33: /var/lib/dpkg/info/archivematica-storage-service.postinst: /usr/share/python/archivematica-storage-service/bin/python: not found
/var/lib/dpkg/info/archivematica-storage-service.postinst: 34: /var/lib/dpkg/info/archivematica-storage-service.postinst: /usr/share/python/archivematica-storage-service/bin/python: not found
updating directory permissions
rm: cannot remove `/tmp/storage-service.log': No such file or directory
dpkg: error processing archivematica-storage-service (--configure):
 subprocess installed post-installation script returned error exit status 1
E: Sub-process /usr/bin/dpkg returned an error code (1)
```

mistydemeo:
Sorry for the radio silence here, but: this is fixed in the `dev/issue-xxxx-packaging` branch, which will be merged in preparation for the upcoming 0.4.0 release. This bug shouldn't affect future Storage Service versions.

Sorry for all the inconvenience this caused - I'm aware the `postinst` issue was a major pain.

BHSPitMonkey:
Thanks for the update.

PR #24, opened 2014-07-16
Infrastructure updates - get_local_path, browse & delete in Spaces, etc

Hwesta:
Several infrastructure and miscellaneous updates in one, coming from AIP reingest work but independent of it.
- `browse` and `delete_path` are now implemented in protocol spaces, instead of in Space or Package, as they are very dependent on the protocol and should be handled by the appropriate space.
- `full_path` and `full_pointer_file_path` are now `@property`'s, since that's effectively  what they are
- Added `get_local_path` to help work with packages that are not store locally, but need to be worked with locally.  As the name implies, gets a local path for that package - either the actual local path, or by copying it to the SS internal location
- Pipelines store some pipeline API information so they can call pipeline APIs
- `compress_package` accepts an algorithm parameter

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -282,7 +282,7 @@ class LocationResource(ModelResource):
         location = bundle.obj
         path = request.GET.get('path', '')
         path = self.decode_path(path)
-        path = os.path.join(str(location.full_path()), path)
+        path = os.path.join(str(location.full_path), path)
 
         objects = self.get_objects(location.space, path)
 
@@ -477,7 +477,7 @@ class PackageResource(ModelResource):
             # This isn't configured by default
             site_url = getattr(settings, "SITE_BASE_URL", None)
             signals.deletion_request.send(sender=self, url=site_url,
-                uuid=package.uuid, location=package.full_path())
+                uuid=package.uuid, location=package.full_path)
         else:
             response = {
                 'error_message': 'A deletion request already exists for this AIP.'
@@ -500,10 +500,10 @@ class PackageResource(ModelResource):
 
         # Get Package details
         package = bundle.obj
-        full_path = package.full_path()
+        full_path = package.full_path
 
         # If local file exists - return that
-        if not package.is_compressed():
+        if not package.is_compressed:
             # The basename of the AIP is included with the request, because
             # all packages contain a base directory. That directory is already
             # inside the full path though, so remove it here.
@@ -547,7 +547,7 @@ class PackageResource(ModelResource):
     @_custom_endpoint(expected_methods=['get'])
     def pointer_file_request(self, request, bundle, **kwargs):
         # Get AIP details
-        pointer_path = bundle.obj.full_pointer_file_path()
+        pointer_path = bundle.obj.full_pointer_file_path
         if not pointer_path:
             response = http.HttpNotFound("Resource with UUID {} does not have a pointer file".format(bundle.obj.uuid))
         else:
@@ -596,7 +596,7 @@ class PackageResource(ModelResource):
         report = json.dumps(response)
         if not success:
             signals.failed_fixity_check.send(sender=self,
-                uuid=bundle.obj.uuid, location=bundle.obj.full_path(),
+                uuid=bundle.obj.uuid, location=bundle.obj.full_path,
                 report=report)
 
         return http.HttpResponse(
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -508,7 +508,7 @@ class PackageResource(ModelResource):
 
         # Get Package details
         package = bundle.obj
-        full_path = package.get_local_path()
+        full_path = package.fetch_local_path()
 
         # If local file exists - return that
         if not package.is_compressed:
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -112,7 +112,7 @@ class PipelineResource(ModelResource):
         # validation = CleanedDataFormValidation(form_class=PipelineForm)
         resource_name = 'pipeline'
 
-        fields = ['uuid', 'description']
+        fields = ['uuid', 'description', 'remote_name', 'api_key', 'api_username']
         list_allowed_methods = ['get', 'post']
         detail_allowed_methods = ['get']
         detail_uri_name = 'uuid'
@@ -122,10 +122,20 @@ class PipelineResource(ModelResource):
             'uuid': ALL,
         }
 
+    def dehydrate(self, bundle):
+        # Don't return API username or key
+        del bundle.data['api_username']
+        del bundle.data['api_key']
+        return bundle
+
     def obj_create(self, bundle, **kwargs):
         bundle = super(PipelineResource, self).obj_create(bundle, **kwargs)
         bundle.obj.enabled = not utils.get_setting('pipelines_disabled', False)
         create_default_locations = bundle.data.get('create_default_locations', False)
+        # Try to guess Pipeline's IP, if doing default setup
+        if create_default_locations:
+            ip = bundle.request.META.get('REMOTE_ADDR') or None
+            bundle.obj.remote_name = ip
         shared_path = bundle.data.get('shared_path', None)
         bundle.obj.save(create_default_locations, shared_path)
         return bundle
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -381,7 +381,7 @@ class PackageResource(ModelResource):
     api/v1/file/<uuid>/delete_aip/ supports:
     POST: Create a delete request for that AIP.
 
-    Validate fixity (api/v1/file/<uuid>/validate/) supports:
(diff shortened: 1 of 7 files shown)


storage_service/locations/models.py
```diff
@@ -1283,19 +1404,19 @@ def store_aip(self, origin_location, origin_path):
         # internal_usage_location/uuid/split/into/chunks/pointer.xml
```
mistydemeo: Update comment too?

storage_service/locations/models.py
```diff
@@ -1164,26 +1231,45 @@ class Package(models.Model):
...
+    @property
     def full_path(self):
         """ Return the full path of the package's current location.
 
         Includes the space, location, and package paths joined. """
         return os.path.normpath(
-            os.path.join(self.current_location.full_path(), self.current_path))
+            os.path.join(self.current_location.full_path, self.current_path))
 
+    @property
```
mistydemeo, line 1264: Move 1b03c073cdf6676c9f9df1f87cbf822d06455c3e after this commit so it can take advantage of this?

storage_service/locations/api/resources.py
```diff
@@ -122,10 +120,20 @@ class Meta:
...
+        return bundle
+
     def obj_create(self, bundle, **kwargs):
         bundle = super(PipelineResource, self).obj_create(bundle, **kwargs)
         bundle.obj.enabled = not utils.get_setting('pipelines_disabled', False)
         create_default_locations = bundle.data.get('create_default_locations', False)
+        # Try to guess Pipeline's IP, if doing default setup
+        if create_default_locations:
+            ip = bundle.request.META['REMOTE_ADDR'] or None
+            bundle.obj.remote_name = ip
```
mistydemeo, line 136: It's okay for this to be `None`?

storage_service/locations/api/resources.py
```diff
@@ -122,10 +120,20 @@ class Meta:
...
+        del bundle.data['api_key']
+        return bundle
+
     def obj_create(self, bundle, **kwargs):
         bundle = super(PipelineResource, self).obj_create(bundle, **kwargs)
         bundle.obj.enabled = not utils.get_setting('pipelines_disabled', False)
         create_default_locations = bundle.data.get('create_default_locations', False)
+        # Try to guess Pipeline's IP, if doing default setup
+        if create_default_locations:
+            ip = bundle.request.META['REMOTE_ADDR'] or None
```
mistydemeo: `bundle.request` always has this key?

storage_service/locations/models.py
```diff
@@ -1224,6 +1311,40 @@ def get_download_path(self, lockss_au_number=None):
             path = full_path
         return path
 
+    def get_local_path(self):
```
mistydemeo, line 1306: Maybe split this into a couple functions? This can cause a potentially very slow operation, which isn't necessarily what I'd expect from something called `get_local_path` (which implies it just returns a path, rather than copies).

e.g. maybe a function to check if there is a local path, and a version of this that returns a local path w/o copying.

storage_service/locations/models.py
```diff
@@ -1507,7 +1669,7 @@ def check_fixity(self):
             failures = failure.details
             message = failure.message
 
-        if self.is_compressed():
+        if delete_after and (self.local_path_location != self.current_location or self.local_path != self.full_path):
```
mistydemeo, line 1684: What's the reason for the secondary condition here?

storage_service/locations/models.py
```diff
@@ -1393,14 +1518,18 @@ def extract_file(self, relative_path='', extract_path=None):
...
             shutil.copytree(aip_path, output_path)
 
+        if not relative_path:
+            self.local_path_location = ss_internal
+            self.local_path = output_path
+
         return (output_path, extract_path)
 
-    def compress_package(self, extract_path=None):
+    def compress_package(self, algorithm, extract_path=None):
```
mistydemeo, line 1535: Make `algorithm` a kwarg with default of `Package.COMPRESSION_TAR`?

storage_service/locations/models.py
```diff
@@ -1393,14 +1518,18 @@ def extract_file(self, relative_path='', extract_path=None):
...
         """
         Produces a compressed copy of the package.
 
         If `extract_path` is provided, will compress there, otherwise to a temp
-        directory.
+        directory in the SS internal location.
```
mistydemeo: Update docstring with `algorithm` too

mistydemeo:
:thumbsup:

storage_service/locations/api/resources.py
Hwesta, line 136: No, but this is a best guess attempt.  `Pipeline.remote_name` defaults to None, and remote calls will not work if it's unset.  However, it can be set by the user - this is just a convenience to set it automatically if possible.

storage_service/locations/api/resources.py
Hwesta: Probably, but using `.get` is safer.  Fixing.

storage_service/locations/api/resources.py
mistydemeo, line 136: Ah, okay.

storage_service/locations/models.py
Hwesta, line 1684: It should only delete the package if we know it's a copy (created from get_local_path or uncompression or something). This is to prevent deleting an uncompresed package stored locally.

storage_service/locations/models.py
Hwesta: Done

storage_service/locations/api/resources.py
Hwesta: Done

storage_service/locations/models.py
Hwesta: Updated this to use Sphinx-style docstring.

storage_service/locations/models.py
Hwesta, line 1306: Hmm. I realize calling a big, slow copy operation is bad, but part of my goal with get_local_path was to abstract as much as possible from the calling functions, so they din't have to worry about checking for a local path and then getting it.

When would you expect there to be a local path when the package isn't stored locally (or locally mounted)?  How would we get a local path without copying?

An earlier version of this checked for a copy of this package in SS_INT and used that if found, but that was problematic since, for example, check_fixity would extract the package, reingest would find the local copy and start to use it, check_fixity would delete local copy when done, and reingest would break. Hence why this creates a new temp dir every time.

(Another problem I ran into was fetching a single file from an uncompressed AIP stored remotely triggered a full package copy, because is_compressed needs to run os.path.isdir locally, but that's a different problem.)

Thoughts? Suggestions?  Maybe rename this to `get_or_fetch_local_path`?  I'm unsure the best way to solve this - this was just my first attempt. :)

mistydemeo, line 1306: > part of my goal with get_local_path was to abstract as much as possible from the calling functions, so they din't have to worry about checking for a local path and then getting it.

My interpretation of the name `get_local_path` was initially that it was simply a getter - like `get_full_path` and friends. I didn't realize it was copying files until I read the code. So, my #1 suggestion would be that this be renamed to something like `copy_to_local_path`so it's clear.

Since there are a couple of code paths that set an attribute and then return a path without expensive disk operations, I think it'd also make sense to expose that distinctly from the rest of this. I could see us making the decision at some point to decide to perform an operation if there's a local path, but skip it entirely otherwise in order to avoid the file copy. Could either do that by making that `get_local_path`, which `copy_to_local_path` checks first as a convenience, or by having 100% separate functions.

Hwesta, line 1306: I like that idea - fetch most of the time, but expose the ability to decide not to if it's expensive.

I'm hesitant to call it `copy_to_local_path` since it may not return a copy (if it's found locally) and I don't want originals getting deleted accidentally. If you want a guaranteed copy, `extract_file` does that (somewhat oddly, now that I think about it. If local and uncompressed will copy it for sure, otherwise returns uncompressed copy). Though, extract_file calls get_local_path...

Maybe `try_local_path` which returns local path or None (or raises an exception?), and `fetch_local_path` which always returns a local path (calls check_local_path, may be the original copy).

storage_service/locations/models.py
Hwesta, line 1535: Why tar as the default? It seems to me if you're compressing something you'd want to specify how you're compressing it.  (And tar isn't technically compression. ;) )

mistydemeo, line 1535: Because then you don't have to change the places that are calling it. ;)

Hwesta, line 1535: Haha, fair, but when reading the code I'd rather know what compression algorithm is being used explicitly, rather than going look for the default param.

storage_service/locations/models.py
mistydemeo, line 1306: How about `get_local_path` (which gets a local path attribute if there is one, and hence makes sense to me as a `get_` method) and `fetch_local_path`?

mistydemeo:
Updates look good! Don't have any more comments besides the ones I already had.

PR #26, opened 2014-08-13
Package.compress_package: fix for files, period in name

mistydemeo:
- If a period is in the name of a directory package that is being
  compressed, os.path.splitext() splits the name of the directory
  as though it were a file extension.
- If full_path was a file and not a directory, the rsplit and subsequent
  join produces incorrect results, e.g. for /path/to/aip/name.7z:
  /path/to/aip/.7z instead of /path/to/aip/

Refs #7102.

### DIFF
--- a/storage_service/locations/models.py
+++ b/storage_service/locations/models.py
@@ -1411,13 +1411,16 @@ class Package(models.Model):
             extract_path = tempfile.mkdtemp()
 
         full_path = self.full_path()
-        basename = os.path.splitext(os.path.basename(full_path))[0]
-        relative_path = ''.join(full_path.rsplit(basename, 1))
+        if os.path.isfile(full_path):
+            basename = os.path.splitext(os.path.basename(full_path))[0]
+        else:
+            basename = os.path.basename(full_path)
+        relative_path = os.path.dirname(full_path)
         compressed_filename = os.path.join(extract_path, basename + '.tar')
 
         command = [
             "tar", "-cf", compressed_filename,
-            "-C", relative_path, basename
+            "-C", relative_path, os.path.basename(full_path)
         ]
         logging.info('Compressing package with: {} to {}'.format(command, compressed_filename))
         rc = subprocess.call(command)


storage_service/locations/models.py
```diff
@@ -1411,13 +1411,16 @@ def compress_package(self, extract_path=None):
             extract_path = tempfile.mkdtemp()
 
         full_path = self.full_path()
-        basename = os.path.splitext(os.path.basename(full_path))[0]
-        relative_path = ''.join(full_path.rsplit(basename, 1))
+        if os.path.isdir(full_path):
+            basename = os.path.splitext(os.path.basename(full_path))[0]
+        else:
+            basename = os.path.basename(full_path)
```
Hwesta, line 1417: I think these are backwards - if it is a dir, just use basename, if not (is file) do splitext(basename()) ?

mistydemeo, line 1417: Oops, you're right.

storage_service/locations/models.py
```diff
@@ -1411,13 +1411,16 @@ def compress_package(self, extract_path=None):
             extract_path = tempfile.mkdtemp()
 
         full_path = self.full_path()
-        basename = os.path.splitext(os.path.basename(full_path))[0]
-        relative_path = ''.join(full_path.rsplit(basename, 1))
+        if os.path.isfile(full_path):
+            basename = os.path.splitext(os.path.basename(full_path))[0]
+        else:
+            basename = os.path.basename(full_path)
+        relative_path = os.path.dirname(full_path)
```
Hwesta, line 1418: This won't work as expected if `full_path = '/path/to/aip/aip-uuid/'` (trailing /).  However, the old code won't work properly in that case either (`os.path.basename` returns ''), so we'll just have to ensure directories _don't_ end with /.

Hwesta:
:+1: I'll try to make sure #24 gets this in rebase since this will probably go in first.

storage_service/locations/models.py
mistydemeo, line 1418: I'm not sure this is possible anyway - don't we vet the path before it gets assigned as an attribute of the `Package` object? If not, I should do that...

PR #27, opened 2014-08-15
updating packaging to fix upgrades

mistydemeo:
added logic to debian/postrm script to prevent files from being deleted during upgrades

### DIFF
--- a/debian/postrm
+++ b/debian/postrm
@@ -1,18 +1,24 @@
-#!/bin/sh
+#!/bin/bash
 
 #postrm 
 #run after the package is uninstalled
 
-#clean up storage service directories
-rm -rf /usr/share/python/archivematica-storage-service
-rm -rf /var/archivematica/storage*
-rm -r /var/archivematica/.storage-service
-rm -f /tmp/storage-service.log
-rm -f /etc/nginx/sites-enabled/storage
-rm -f /etc/nginx/sites-available/storage
-rm -f /etc/uwsgi/apps-enabled/storage.ini
-rm -f /etc/uwsgi/apps-available/storage.ini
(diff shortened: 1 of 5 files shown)


debian/postrm
```diff
@@ -3,16 +3,24 @@
...
-rm -rf /usr/share/python/archivematica-storage-service
-rm -rf /var/archivematica/storage*
-rm -r /var/archivematica/.storage-service
-rm -f /tmp/storage-service.log
-rm -f /etc/nginx/sites-enabled/storage
-rm -f /etc/nginx/sites-available/storage
-rm -f /etc/uwsgi/apps-enabled/storage.ini
-rm -f /etc/uwsgi/apps-available/storage.ini
-rm -rf /usr/lib/archivematica/storage*
+#!/bin/bash
```
mistydemeo: There's already a shebang earlier, and the shebang has to be the first line of the script - replace the `/bin/sh` before?

debian/postrm
```diff
@@ -3,16 +3,24 @@
...
+
+echo "postrm" $1
+if [ $1 = "remove" ]; then
+    #clean up storage service directories
+    rm -rf /usr/share/python/archivematica-storage-service
+    rm -rf /var/archivematica/storage*
+    rm -r /var/archivematica/.storage-service
+    rm -f /tmp/storage-service.log
+    rm -f /etc/nginx/sites-enabled/storage
```
mistydemeo, line 13: Not sure if removing is the best option? Users may have customized these and wouldn't expect their settings to go away.

For example, one of our clients is deploying with significant alterations to this file.

mistydemeo:
(This branch is by @jhsimpson, for reference)

PR #28, opened 2014-08-20
Split models.py into several files

Hwesta:
locations/models.py is approaching 2000 lines long. Move it to a models directory and split into several files.

Most models get a file, with some similar models grouped together in the same file. Use init.py used to maintain the same import structure.

The first commit is just models.py copied into several files with trimmed imports, to make it easier to rebase. The second commit is the changes required to make this work (notably adding app_label = 'locations') as well as pep8 updates. They will probably be rebased together when committed.


mistydemeo:
:thumbsup: 

I am not looking forward to rebasing FEDORA/Islandora on top of this if this goes in first though. :crying_cat_face:

Hwesta:
Rebased on top of #7 and #22

mistydemeo:
Still :thumbsup: 

Though maybe this should be made one commit so we don't leave the SS broken in between commits?

Hwesta:
Okay, I'll squash the first two commits.

PR #29, opened 2014-09-09
Extract file: Optionally strip SIP name & UUID

Hwesta:
refs #7214

Uncompressed AIPs may not have the SIP name & UUID in the file path. Currently, view AIP files has the extra folder, but view backlog files does not. Update code to work whether or not the path is prepended with the SIP name & UUID by optionally stripping it.

This should also be merged into qa/0.x

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -500,15 +500,17 @@ class PackageResource(ModelResource):
 
         # Get Package details
         package = bundle.obj
-        full_path = package.full_path()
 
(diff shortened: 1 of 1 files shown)


jhsimpson:
I haven't tested but this looks good to me.
:thumbsup:

mistydemeo:
:thumbsup:

Hwesta:
Also cherry-picked to qa/0.x as 1548fc9d9c529fb166cd469b7b7e2d183018417e

PR #31, opened 2014-09-15
Restrict Location purposes based on Space's protocol

Hwesta:
Not all Spaces support all purposes. Add the ability to restrict what purpose a Location can be based on the Space's protocol.

This requires #28


mistydemeo:
:thumbsup:

PR #32, opened 2014-09-15
Add Duracloud

Hwesta:
Add Duracloud as a Space.

Note this requires #28 and #31

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -13,3 +13,4 @@ django-jsonfield==0.9.12
 bagit==1.3.7
 django-annoying==0.7.7
 git+https://github.com/swordapp/python-client-sword2.git@0c241438f2ced13ec3364caa03edf28d5ed5c02c
+requests>=2.3.0
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -7,8 +7,6 @@ South==0.8.4
 django-tastypie==0.9.15
 django-extensions==1.1.1
 lxml==3.2.3
-requests==2.2.1
-six==1.4.1
 django-jsonfield==0.9.12
 bagit==1.3.7
 django-annoying==0.7.7
--- a/requirements/local.txt
+++ b/requirements/local.txt
@@ -7,3 +7,4 @@ Sphinx==1.2b1
 ipython==1.1.0
 pytest
 pytest-django
+vcrpy>=1.0.0
--- a/storage_service/locations/models/duracloud.py
+++ b/storage_service/locations/models/duracloud.py
@@ -49,6 +49,23 @@ class Duracloud(models.Model):
     def duraspace_url(self):
         return 'https://' + self.host + '/durastore/' + self.duraspace + '/'
 
+    def _get_files_list(self, prefix):
+        """
+        Fetch full path of all files starting with prefix. Helper function.
+        """
+        params = {'prefix': prefix}
+        response = self.session.get(self.duraspace_url, params=params)
+        if response.status_code != 200:
+            raise StorageException('Unable to get list of files in %s' % prefix)
+        # Response is XML in the form:
+        # <space id="self.durastore">
+        #   <item>path</item>
+        #   <item>path</item>
+        # </space>
+        root = etree.fromstring(response.content)
+        paths = [p.text for p in root]
+        return paths
+
     def move_to_storage_service(self, src_path, dest_path, dest_space):
         """ Moves src_path to dest_space.staging_path/dest_path. """
         # Try to fetch if it's a file
@@ -56,18 +73,13 @@ class Duracloud(models.Model):
         response = self.session.get(url)
         if response.status_code == 404:
             # File cannot be found - this may be a folder
-            # Get all elements with dest_path as a prefix and fetch them
-            params = {'prefix': src_path}
-            response = self.session.get(self.duraspace_url, params=params)
-            if response.status_code != 200:
-                raise StorageException('Unable to fetch %s' % src_path)
-            # Response is XML in the form:
-            # <space id="self.durastore">
-            #   <item>path</item>
-            #   <item>path</item>
-            # </space>
-            root = etree.fromstring(response.content)
-            to_get = [e.text for e in root]
+            to_get = self._get_files_list(src_path)
+            if not to_get:
+                # If nothing found, trying normalizing src_path to remove possible extra characters like /. /* /  These glob-match on a filesystem, but do not character-match in Duracloud.
+                # Normalize dest_path as well so replace continues to work
+                src_path = os.path.normpath(src_path)
+                dest_path = os.path.normpath(dest_path)
+                to_get = self._get_files_list(src_path)
             for entry in to_get:
                 dest = entry.replace(src_path, dest_path, 1)
                 url = self.duraspace_url + entry
--- a/storage_service/locations/models/duracloud.py
+++ b/storage_service/locations/models/duracloud.py
@@ -66,6 +66,15 @@ class Duracloud(models.Model):
         paths = [p.text for p in root]
         return paths
 
+    def browse(self, path):
+        if not path.endswith('/'):
+            path += '/'
+        paths = self._get_files_list(path)
+        paths = [p.replace(path, '', 1) for p in paths]
+        entries = sorted(list(set(p.split('/')[0] for p in paths)))
+        directories = sorted(set(p.split('/')[0] for p in paths if len(p.split('/')) > 1))
+        return {'directories': directories, 'entries': entries}
+
     def move_to_storage_service(self, src_path, dest_path, dest_space):
         """ Moves src_path to dest_space.staging_path/dest_path. """
         # Try to fetch if it's a file
--- a/storage_service/locations/models/duracloud.py
+++ b/storage_service/locations/models/duracloud.py
@@ -1,5 +1,6 @@
 # stdlib, alphabetical
 import os
+from lxml import etree
 
 # Core Django, alphabetical
 from django.db import models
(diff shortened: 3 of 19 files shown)


Hwesta:
We now have two 0005_foo migrations. (South handles this correctly by running them in alphabetical order).  

Previously, we named migrations after the release in question, but we had some problems with that in dev/testing environments. Do we want to switch to a migration per feature or continue updating the release-named migration?

I can rename 0005_v0_5_0.py to 0005_fedora.py and 0005_duracloud.py to 0006_duracloud.py if we go the per-feature route.

mistydemeo:
> I can rename 0005_v0_5_0.py to 0005_fedora.py and 0005_duracloud.py to 0006_duracloud.py if we go the per-feature route.

I would really, really like to do this! It makes it much easier for me to keep the storage service around for extended periods, running smaller migrations forward and backwards as necessary.

storage_service/locations/models/duracloud.py
```diff
@@ -0,0 +1,138 @@
...
+        paths = [p.text for p in root]
+        return paths
+
+    def browse(self, path):
+        if not path.endswith('/'):
+            path += '/'
+        paths = self._get_files_list(path)
+        paths = [p.replace(path, '', 1) for p in paths]
+        for p in paths:
+            print p
```
mistydemeo: Debugging?

storage_service/locations/models/duracloud.py
```diff
@@ -0,0 +1,138 @@
...
+        entries = sorted(list(set(p.split('/')[0] for p in paths)))
+        directories = sorted(list(set(p.split('/')[0] for p in paths if len(p.split('/')) > 1)))
```
mistydemeo: You can run `sorted()` on a `set` directly without needing to create an in-memory list first.

storage_service/locations/models/duracloud.py
Hwesta: Whoops, yes. Removed.

storage_service/locations/models/duracloud.py
Hwesta: Nice catch. Extra list removed.

Hwesta:
`0005_duracloud.py` renamed `0006_duracloud.py`.  Renaming of the fedora/0_5_0 migration done in #34

requirements/base.txt
```diff
@@ -7,9 +7,8 @@ South==0.8.4
 django-tastypie==0.9.15
 django-extensions==1.1.1
 lxml==3.2.3
-requests==2.2.1
-six==1.4.1
 django-jsonfield==0.9.12
 bagit==1.3.7
 django-annoying==0.7.7
 git+https://github.com/swordapp/python-client-sword2.git@0c241438f2ced13ec3364caa03edf28d5ed5c02c
+requests>=2.3.0
```
mistydemeo: Add a newline to the end of the file?

requirements/local.txt
```diff
@@ -7,3 +7,4 @@ Sphinx==1.2b1
 ipython==1.1.0
 pytest
 pytest-django
+vcrpy>=1.0.0
```
mistydemeo: Ditto

storage_service/locations/models/duracloud.py
```diff
@@ -0,0 +1,144 @@
...
+    @property
+    def session(self):
+        if self._session is None:
+            self._session = requests.Session()
+            self._session.auth = (self.user, self.password)
+        return self._session
+
+    @property
+    def duraspace_url(self):
+        return 'https://' + self.host + '/durastore/' + self.duraspace + '/'
```
mistydemeo, line 50: Use `urlparse.urljoin` for this? https://docs.python.org/2/library/urlparse.html#urlparse.urljoin

storage_service/locations/models/duracloud.py
```diff
@@ -0,0 +1,144 @@
...
+        entries = sorted(set(p.split('/')[0] for p in paths if p.split('/')[0]))
+        directories = sorted(set(p.split('/')[0] for p in paths if len(p.split('/')) > 1 and p.split('/')[0]))
+        return {'directories': directories, 'entries': entries}
+
+    def delete_path(self, delete_path):
+        # Files
+        url = self.duraspace_url + delete_path
+        response = self.session.delete(url)
+        if response.status_code == 404:
+            # File cannot be found - this may be a folder
```
mistydemeo, line 83: We can't check ahead of time if it's a folder?

storage_service/locations/models/duracloud.py
```diff
@@ -0,0 +1,144 @@
...
+                for basename in files:
+                    entry = os.path.join(path, basename)
+                    dest = entry.replace(source_path, destination_path, 1)
+                    url = self.duraspace_url + dest
+                    self._upload_file(url, entry)
+        elif os.path.isfile(source_path):
+            url = self.duraspace_url + destination_path
+            self._upload_file(url, source_path)
+        else:
+            raise StorageException('%s is neither a file nor a directory, may not exist' % source_path)
```
mistydemeo: Ass an explicit `os.path.exists` check?

storage_service/locations/tests/test_duracloud.py
```diff
@@ -0,0 +1,113 @@
...
+
+class TestDuracloud(TestCase):
+
+    fixtures = ['initial_data.json', 'duracloud.json']
+
+    def setUp(self):
+        self.ds_object = models.Duracloud.objects.all()[0]
+
+    def test_has_required_attributes(self):
+        assert self.ds_object.host
```
mistydemeo, line 18: This is really just testing the test fixture, not the actual code, isn't it?

storage_service/locations/tests/test_duracloud.py
```diff
@@ -0,0 +1,113 @@
...
+        assert response.status_code == 404
+        response = requests.get('https://trial.duracloud.org/durastore/trial263//ts/test/subfolder/test2.txt', auth=auth)
+        assert response.status_code == 404
+
+    @vcr.use_cassette('locations/fixtures/vcr_cassettes/duracloud_move_from_ss.yaml')
+    def test_move_from_ss(self):
+        # Create test.txt
+        open('test.txt', 'w').write('test file\n')
+        # Upload
+        self.ds_object.move_from_storage_service('test.txt', '/ts/test.txt')
```
mistydemeo, line 55: Not necessarily key at this stage, but might be nice to mock the filesystem code so we don't have to actually hit the filesystem during the test.

storage_service/locations/tests/test_duracloud.py
```diff
@@ -0,0 +1,113 @@
...
+        # Unsure how to verify. Will raise exception if fails?
+        # Cleanup
+        os.remove('test.txt')
+        auth = requests.auth.HTTPBasicAuth(self.ds_object.user, self.ds_object.password)
+        requests.delete('https://' + self.ds_object.host + '/durastore/' + self.ds_object.duraspace + '/ts/test.txt', auth=auth)
```
mistydemeo, line 60: `urljoin` here too?

storage_service/locations/tests/test_duracloud.py
```diff
@@ -0,0 +1,113 @@
...
+        assert os.path.isdir('folder/test')
+        assert os.path.isdir('folder/test/subfolder')
+        assert os.path.isfile('folder/test/test.txt')
+        assert os.path.isfile('folder/test/subfolder/test2.txt')
+        assert open('folder/test/test.txt').read() == 'test file\n'
+        assert open('folder/test/subfolder/test2.txt').read() == 'test file2\n'
+        # Cleanup
+        os.remove('folder/test/test.txt')
+        os.remove('folder/test/subfolder/test2.txt')
+        # BUG Globbing may match and fetch extra things (because of use of normpath)
```
mistydemeo, line 111: Hm, I'm a bit nervous about this. Is this out of scope to fix right now?

storage_service/locations/models/duracloud.py
Hwesta, line 50: I don't think this is worthwhile. Unlike `os.path.join`,  `urlparse.urljoin` only accepts two parameters - the base and relative path.  I would have to make almost all the URL with `+` before passing it into urljoin, which seems silly at that point.  eg `urlparse.urljoin('https://' + self.host, 'durastore/' + self.duraspace + '/')`

storage_service/locations/models/duracloud.py
Hwesta, line 83: Duracloud is object storage - folders don't actually exist. To make it play nice with the existing structure and Archivematica I have it fake folders by looking for the provided path as prefix when it can't find something. If it still doesn't find anything, then the 'folder' must not exist.

storage_service/locations/tests/test_duracloud.py
Hwesta, line 18: Yeah. I set it up when I was testing the fixtures and didn't see a reason to remove it.

mistydemeo, line 18: Makes sense, just checking I wasn't missing something.

storage_service/locations/models/duracloud.py
Hwesta: Done

storage_service/locations/tests/test_duracloud.py
Hwesta, line 60: Probably not, for reasons stated above.

requirements/local.txt
Hwesta: Done for both.

storage_service/locations/tests/test_duracloud.py
Hwesta, line 55: I agree, but I don't know how to do that.

storage_service/locations/models/duracloud.py
mistydemeo, line 50: Oh, good point - makes sense. Never mind, in that case.

storage_service/locations/models/duracloud.py
mistydemeo, line 83: Ah, okay.

PR #33, opened 2014-09-19
Add Swift

Hwesta:
Add Swift as a Space.

This requires #31

### DIFF
--- a/storage_service/locations/models/space.py
+++ b/storage_service/locations/models/space.py
@@ -90,6 +90,7 @@ class Space(models.Model):
     LOM = 'LOM'
     NFS = 'NFS'
     PIPELINE_LOCAL_FS = 'PIPE_FS'
+    SWIFT = 'SWIFT'
     OBJECT_STORAGE = {DURACLOUD}
     ACCESS_PROTOCOL_CHOICES = (
         (DURACLOUD, 'DuraCloud'),
@@ -98,6 +99,7 @@ class Space(models.Model):
         (LOM, "LOCKSS-o-matic"),
         (NFS, "NFS"),
         (PIPELINE_LOCAL_FS, "Pipeline Local Filesystem"),
+        (SWIFT, "Swift"),
     )
     access_protocol = models.CharField(max_length=8,
         choices=ACCESS_PROTOCOL_CHOICES,
--- a/storage_service/locations/models/space.py
+++ b/storage_service/locations/models/space.py
@@ -384,13 +384,23 @@ class Space(models.Model):
             raise StorageException(s)
 
     def _create_local_directory(self, path, mode=None):
-        """ Creates a local directory at 'path' with 'mode' (default 775). """
+        """
+        Creates directory structure for `path` with `mode` (default 775).
+        :param path: path to create the directories for.  Should end with a / or
+            a filename, or final directory may not be created. If path is empty,
+            no directories are created.
+        :param mode: (optional) Permissions to create the directories with
+            represented in octal (like bash or the stat module)
+        """
         if mode is None:
             mode = (stat.S_IRUSR + stat.S_IWUSR + stat.S_IXUSR +
                     stat.S_IRGRP + stat.S_IWGRP + stat.S_IXGRP +
                     stat.S_IROTH +                stat.S_IXOTH)
+        dir_path = os.path.dirname(path)
+        if not dir_path:
+            return
         try:
-            os.makedirs(os.path.dirname(path), mode)
+            os.makedirs(dir_path, mode)
         except os.error as e:
             # If the leaf node already exists, that's fine
             if e.errno != errno.EEXIST:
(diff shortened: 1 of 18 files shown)


storage_service/locations/models/space.py
```diff
@@ -306,13 +317,22 @@ def _move_rsync(self, source, destination):
...
+        :param mode: (optional) Permissions to create the directories with
+            represented in octal (like bash or the stat module)
+        """
         if mode is None:
             mode = (stat.S_IRUSR + stat.S_IWUSR + stat.S_IXUSR +
                     stat.S_IRGRP + stat.S_IWGRP + stat.S_IXGRP +
                     stat.S_IROTH +                stat.S_IXOTH)
+        dir_path = os.path.dirname(path)
+        if not dir_path:
+            return
```
mistydemeo, line 401: Let's make sure to document this, and/or raise - I could see this causing some confusing bugs.

storage_service/locations/models/swift.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def delete_path(self, delete_path):
+        # Try to delete object
+        try:
+            self.connection.delete_object(self.container, delete_path)
+        except swiftclient.exceptions.ClientException:
+            # Swift only stores objects and fakes having folders. If delete_path
+            # doesn't exist, assume it is supposed to be a folder and fetch all
+            # items with that prefix to delete.
+            _, content = self.connection.get_container(self.container, prefix=delete_path)
```
mistydemeo: Is it possible for this to also raise? If so, maybe better to reraise with a more explanatory message? (E.g., "File X does not exist, but no container named X exists either; unable to delete any content")

storage_service/locations/tests/test_swift.py
```diff
@@ -0,0 +1,87 @@
...
+    @vcr.use_cassette('locations/fixtures/vcr_cassettes/swift_delete.yaml')
+    def test_delete_path(self):
+        self.swift_object.delete_path('/test.txt')
+
+    @vcr.use_cassette('locations/fixtures/vcr_cassettes/swift_delete_folder.yaml')
+    def test_delete_folder(self):
+        self.swift_object.delete_path('/aips/c521/')
+
+    def test_update_package_status(self):
+        pass
```
mistydemeo: Is this meant to have contents?

storage_service/locations/models/space.py
Hwesta, line 401: I updated the docstring.  The original (intended?) usage is to pass the path to the file that needs directories, and this handles any directory creation required.  Eg. path/to/foo.txt creates path/to, but just foo.txt creates nothing because nothing needs creating.

storage_service/locations/models/swift.py
Hwesta: Catches the exception and logs it.

Hwesta:
I'd like to beef up the tests and logging, but the testing server I was using to generate the cassettes doesn't exist anymore.  Will have to follow up on that.

Migrations also need to be checked to ensure they're up to date after rebases.

storage_service/locations/tests/test_swift.py
Hwesta: Since there's no update_package_status in Swift yet, I'll remove it until we have something to actually test!

storage_service/locations/models/swift.py
```diff
@@ -0,0 +1,167 @@
...
+
+class Swift(models.Model):
+    space = models.OneToOneField('Space', to_field='uuid')
+    auth_url = models.CharField(max_length=256,
+        help_text='URL to authenticate against')
+    auth_version = models.CharField(max_length=8, default='2',
+        help_text='OpenStack auth version')
+    username = models.CharField(max_length=64,
+        help_text='Username to authenticate as')
+    # HELP how do I store the password?  Has to be plaintext to send to Swift, but that seems like a bad idea
```
mistydemeo, line 29: Not immediately practical since this isn't being executed in the context of a user, but an interesting idea: http://stackoverflow.com/a/8036765 Use a user's password as the encryption key for the password, so the password never has to be stored in plaintext, while making it decodeable.

storage_service/locations/models/swift.py
```diff
@@ -0,0 +1,167 @@
...
+        headers, content = self.connection.get_object(self.container, remote_path)
+        self.space._create_local_directory(download_path)
+        with open(download_path, 'wb') as f:
+            f.write(content)
+        # Check ETag matches checksum of this file
+        if 'etag' in headers:
+            checksum = utils.generate_checksum(download_path)
+            if checksum.hexdigest() != headers['etag']:
+                logging.warning('ETag %s for %s does not match %s', remote_path, headers['etag'], checksum.hexdigest())
+                # TODO what to do here? Raise exception?
```
mistydemeo: I think that makes sense, yeah. It sounds like a fatal condition to me.

mistydemeo:
:+1: aside from comments

Hwesta:
Updated tests, responded to feedback, rebased.

mistydemeo:
:shipit:

PR #34, opened 2014-09-19
Migrations: Move Fedora changes to new migration

Hwesta:
Fedora-related changes should not be in the released 0.4.0 migration, but should be in their own. This is a holdover from when Fedora was supposed to be in 0.4.0.

Note that because Package.origin_pipeline was made NULLable, we cannot migrate backwards from 0005_fedora to 0004_v0_4_0. Suggestions for an actual default value welcome.

### DIFF
--- a/storage_service/locations/migrations/0005_fedora.py
+++ b/storage_service/locations/migrations/0005_fedora.py
@@ -31,6 +31,47 @@ class Migration(SchemaMigration):
         ))
         db.send_create_signal(u'locations', ['Callback'])
 
+        # Adding model 'Fedora'
+        db.create_table(u'locations_fedora', (
+            (u'id', self.gf('django.db.models.fields.AutoField')(primary_key=True)),
+            ('space', self.gf('django.db.models.fields.related.OneToOneField')(to=orm['locations.Space'], to_field='uuid', unique=True)),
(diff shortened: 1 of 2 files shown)


storage_service/locations/migrations/0005_fedora.py
```diff
@@ -77,7 +137,7 @@ def backwards(self, orm):
             'model': ('django.db.models.fields.CharField', [], {'max_length': '100'}),
             'name': ('django.db.models.fields.CharField', [], {'max_length': '100'})
         },
-        u'locations.callback': {
```
mistydemeo, line 80: Skip making these `str` vs `unicode` differences so the diff only has germane changes?

mistydemeo:
:thumbsup:

storage_service/locations/migrations/0005_fedora.py
Hwesta, line 80: The schemamigration thing did that automatically, but sure.

mistydemeo, line 80: I don't generally care about `str` vs `unicode` in this context, it's just nice to minimize the diff so, e.g., `git blame` doesn't get thrown off in the future.

Hwesta, line 80: Done.

mistydemeo:
Thanks!

PR #36, opened 2014-10-06
Logging: Configure

Hwesta:
Configure Django's logging and change references from logging root module (logging.foo) to the fetched logger. Disable South having logging at DEBUG.

The biggest change we'll have to make in future is to use LOGGER.{debug,info,warning,error} instead of logging.{debug,info,warning,error}, as the latter always goes to the root module and makes it hard to track where logs are coming from.

refs #6812


mistydemeo:
Haven't tested, but looks good.

jhsimpson:
These changes have been tested in a couple of instances, very helpful.

Hwesta:
Also cherry-picked to qa/0.6.x

PR #37, opened 2014-10-07
Update to Django 1.7

Hwesta:
Update to use Django 1.7, including fixes for backwards compatibliity and other versions bumps.


mistydemeo:
:+1:

Hwesta:
Updated to handle converting from South to Django migrations.  Ready to be looked at again. Should be tested upgrading packages from 0.3, 0.4 and 0.5.

debian/postinst
```diff
@@ -28,7 +30,16 @@ rsync -a /var/archivematica/storage-service/templates /usr/lib/archivematica/sto
...
 
 echo "configuring django database and static files"
-/usr/share/python/archivematica-storage-service/bin/python manage.py syncdb
+find /usr/lib/archivematica/storage-service/ -name '*.pyc'  -delete
+# 0.3.0 will automatically fake the first migration
+if [[ $2 == '0.6.'* || $2 == '0.5.'* || $2 == '0.4.'* ]]; then
+    echo 'Faking the 0.4 migration'
+    /usr/share/python/archivematica-storage-service/bin/python manage.py migrate locations 0002_v0_4 --fake
+fi
+if [[ $2 == '0.6.'* || $2 == '0.5.'* ]]; then
```
Hwesta: Check with Justin that these actually match the versions postinst will be called with.

requirements/base.txt
```diff
@@ -1,10 +1,9 @@
 # Base requirements - for all installations
-Django==1.5.4
+Django>=1.7,<1.8
```
mistydemeo, line 2: 🚀

storage_service/storage_service/settings/test.py
```diff
@@ -1,10 +1,7 @@
 from base import *
 
 ########## TEST SETTINGS
-TEST_RUNNER = 'discover_runner.DiscoverRunner'
-TEST_DISCOVER_TOP_LEVEL = SITE_ROOT
-TEST_DISCOVER_ROOT = SITE_ROOT
-TEST_DISCOVER_PATTERN = "test_*.py"
+# TEST_RUNNER = 'django.test.runner.DiscoverRunner'
```
mistydemeo: Commented out?

mistydemeo:
Haven't tested but looks good :+1:

Hwesta:
Added south migration for 0.7

mistydemeo:
:+1:

storage_service/storage_service/settings/test.py
Hwesta: Looks like it was moved to base.py and I forgot to delete it here. Fixed.

mistydemeo:
:shipit:

PR #38, opened 2014-10-10
Migrations: Fix to upgrade from 0.4.0 migration

Hwesta:
Fix 0004_v0_4_0 migration to match what's in the stable/0.4.0 branch. Update the fedora and duracloud migrations to upgrade properly.  All generated using `manage.py schemamigration locations --auto <name>`


jhsimpson:
I have tested this twice and the migration from 0.4.0 to 0.5.0 worked both times

:shipit:

requirements/production.txt
```diff
@@ -2,4 +2,6 @@
 #	production that isn't in development.
 -r base.txt
 
+# Install from local in packaging
+Sword2
```
mistydemeo: Let's check the tarball we're using into version control.

jhsimpson:
:shipit:

PR #40, opened 2014-10-20
Default Locations: Add exception handling

Hwesta:
Add try-except blocks around Space and Location creation so that defaults creation does not block interacting with the storage service if there are problems.  Restructure Location creation to reduce boilerplate.

refs #7416

### DIFF
--- a/storage_service/storage_service/urls.py
+++ b/storage_service/storage_service/urls.py
@@ -34,6 +34,7 @@ urlpatterns = patterns('',
 
 
 def startup():
+    import django.core.exceptions
     import errno
     import os.path
     from locations import models as locations_models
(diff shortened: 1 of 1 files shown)


storage_service/storage_service/urls.py
```diff
@@ -41,56 +42,80 @@ def startup():
...
+            access_protocol=locations_models.Space.LOCAL_FILESYSTEM,
+            path=os.sep, defaults={
+                "staging_path": os.path.join(os.sep, 'var', 'archivematica', 'storage_service')
+            })
+        if space_created:
+            locations_models.LocalFilesystem.objects.create(space=space)
+            logger.info('Created default Space %s', space)
+    except django.core.exceptions.MultipleObjectsReturned:
+        logger.info('Multiple default Spaces exist, done default setup.')
+        return
```
mistydemeo, line 57: Maybe don't return here in case there are somehow multiple spaces, but no default locations?

mistydemeo:
:+1:

storage_service/storage_service/urls.py
Hwesta, line 57: If there's an exception, then `space` isn't defined, so we have nowhere to create the default locations in.  I don't want to take the first one we find, since at that point we don't know what's going on, and this is intended for convenient setup not anything more complicated.

mistydemeo, line 57: OK, fair enough.

PR #41, opened 2014-10-22
SWORD: move downloaded MODS into subdirectory

mistydemeo:
This is needed to support artefactual/archivematica#128.


Hwesta:
:+1: Probably should go in after #38 and release though.

Hwesta:
:+1:

PR #42, opened 2014-10-23
Fix delete request package type

Hwesta:
Also fix raise -> return.


jhsimpson:
should change the title of the deletion request page from 'aip deletion' to 'package deletion' to match this change.

Hwesta:
Fixed.

jhsimpson:
:shipit:

PR #44, opened 2014-11-13
Handle unicode better

Hwesta:
Check that browse comparisons are str to str, not mixing str and unicode.

Update all logs to use %s, as it doesn't have problems mixing str and unicode strings.


mistydemeo:
:+1:

PR #45, opened 2014-11-14
Duracloud: Handle fetching many files from a space

Hwesta:
Duracloud returns a maximum of 1000 files when browsing a space. Update code to call that endpoint repeatedly to fetch all files with a given prefix. To handle the larger number of files, convert _get_files_list to an
iterator.  Update browse to deal with paths one at a time, instead of fetching all of them at once.

Update tests for new calls.  Split some tests into multiple for ease of testing and changing later (eg test_move_to_storage_service -> test_move_to_storage_service_file & test_move_to_storage_service_folder

refs #7481


storage_service/locations/models/duracloud.py
```diff
@@ -64,18 +68,38 @@ def _get_files_list(self, prefix):
...
+            entries.add(dirname)
+            if len(path_parts) > 1:
+                directories.add(dirname)
+
+        entries = sorted(entries, key=lambda s: s.lower())  # Also converts to list
+        directories = sorted(directories, key=lambda s: s.lower())  # Also converts to list
         return {'directories': directories, 'entries': entries}
 
     def delete_path(self, delete_path):
+        # BUG If delete_path is a folder but provided without a trailing /, will deleted a file with the same name.
```
mistydemeo: 😱

(Also, `will delete`?)

mistydemeo:
👍

PR #47, opened 2014-11-19
Enable AIP recovery, fix testing fixtures

Hwesta:
Add AIP recovery constants to Spaces that support it.

Update check that ensures only one AIP recovery location is associated with a Pipeline and move it to the clean() function.

Add tests for this.  Fix fixtures so that initial_data.json (renamed to base.json) is actually loaded.


storage_service/locations/views.py
```diff
@@ -234,34 +233,23 @@ def location_edit(request, space_uuid, location_uuid=None):
...
+        # Cannot use form.save_m2m() because of 'through' table
+        for pipeline in form.cleaned_data['pipeline']:
+            LocationPipeline.objects.get_or_create(
+                location=location, pipeline=pipeline)
+
+        # Delete relationships between the location and pipelines not in the form
+        to_delete = LocationPipeline.objects.filter(location=location).exclude(
+            pipeline__in=list(form.cleaned_data['pipeline']))
+        # NOTE Need to convert form.cleaned_data['pipeline'] to a list, or the
+        # SQL generated by pipeline__in is garbage.  Django 1.5
```
mistydemeo: What's the "Django 1.5" note, that this issue is specific to that version?

mistydemeo:
👍

storage_service/locations/views.py
Hwesta: It's the version of Django I ran into the problem in, so I know to revisit and remove the hack later.  Updated to be clearer.

Hwesta:
Added fix to config and rebased on qa/0.x

mistydemeo:
Still :+1:

PR #48, opened 2014-11-19
Minor bugfixes to contribute back from integration tests by discoverygarden

daniel-dgi:
Changed calls to full_path to properties.  Adding MODS folder before download starts.

Don't know if you'll need these, as I see the codebase has shifted a bit, but we've fixed a few minor issues we ran into when trying to integrate our two systems for some client work.  Hope this helps!


mistydemeo:
Incorporated into #41 - thanks!

PR #49, opened 2014-11-22
Duracloud: Handle hidden files in folders

Hwesta:
refs #7549

When trying to fetch the contents of a folder, Archivematica passes `path/to/folder/.` If the folder had files starting with ., they would be returned when querying with the path as prefix. Only if no files were returned was the path normalized. This meant non-hidden files were not fetched if the directory had hidden files in it.

If we think this is a folder, always strip `/.` and `/*` from the end of a path, if they exist.  This removes support for using `..` and other shell-isms from folder paths, but we probably shouldn't be using those anyway.


mistydemeo:
👍

Hwesta:
Also cherry-picked to 0.5.1 branch

PR #50, opened 2014-12-02
Duracloud: Percent encode file paths when fetching

Hwesta:
Duracloud was truncating the file paths at the #, causing gets of those files to fail.  Encode all paths that go to Duracloud.  Add tests.

Should probably be committed to qa/0.5.1 as well

refs #7643


mistydemeo:
:+1:

Hwesta:
Also cherry-picked to qa/0.5.1

PR #51, opened 2014-12-05
Duracloud: Don't encode GET params

Hwesta:
While URLs must be percent encoded, GET parameters do not work if they are encoded.


mistydemeo:
:+1:

PR #52, opened 2014-12-09
improve installer script

jhsimpson:
refs #7312

Do not change permissions on files that are not created by the
storage service.

Previously, the installer chown'ed the entire /var/archivematica directory.  Some users have put their own files or symlinks to other filesystems in /var/archivematica, so this change limits the scope of what is chown'ed.


mistydemeo:
:+1:

PR #53, opened 2014-12-09
increase default timeout values

jhsimpson:
refs #7311

The storage service uses nginx by default as a web server. The default
timeout values are 60 seconds in nginx.  This change increases
those timeout values to 1 hour.  This allows Archivematica to POST
large files (e.g. large AIP's) to the Storage Service.

### DIFF
--- a/install/storage
+++ b/install/storage
@@ -18,6 +18,8 @@ server {
 
     location / {
         uwsgi_pass storage;
+        uwsgi_read_timeout 3600;
+        uwsgi_send_timeout 3600;
         include /etc/nginx/uwsgi_params;
     }
 


mistydemeo:
:+1:

Have we documented the suggestion to users that it can be made even higher than this if necessary?

jhsimpson:
This is not documented in our docs, it has come up I believe on the discussion forum, but it would take a dedicated person to find it.  That change should be done in the new archivematica docs.  As you can see here https://ww.archivematica.org/en/docs/storage-service-0.4/administrators/#administrators there isn't anything about nginx or uwsgi config mentioned yet.

PR #54, opened 2014-12-11
Duracloud: Coerce all strings to byte strings

Hwesta:
Base64 and percent encoding assume that strings are strs, not unicode. Also, str.replace() behaves poorly if the string and the replacement strings are not the same type of string.

refs #7694


mistydemeo:
:+1:

PR #55, opened 2014-12-12
Replace Base64 with atob/btoa

mistydemeo:
This is the companion PR to artefactual/archivematica#144.

Fixes #7698.


Hwesta:
:+1:

PR #56, opened 2015-01-09
Duracloud: Handle chunked files

Hwesta:
Add support for browsing, uploading, downloading and deleting chunked files.  This is a temporary measure until chunking and stitching is done in the Duracloud API.

Updated several tests to make more robust/consistent, and generated new cassettes for tests that changed.


jhsimpson:
This code has been tested on a number of test sites and is working well so far.  Testing continues

mistydemeo:
Are we planning to get this into 0.7.0?

Hwesta:
I'd like to, yes.

storage_service/locations/models/duracloud.py
```diff
@@ -129,16 +145,95 @@ def delete_path(self, delete_path):
...
+                    size = int(e.findtext('byteSize'))
+                    md5 = e.findtext('md5')
+                    # Download
+                    chunk_url = self.duraspace_url + urllib.quote(chunk)
+                    LOGGER.debug('Chunk URL: %s', chunk_url)
+                    chunk_path = chunk_url.replace(url, download_path)
+                    LOGGER.debug('Chunk path: %s', chunk_path)
+                    self._download_file(chunk_url, chunk_path, size, md5)
+                    # Append to output
+                    shutil.copyfileobj(open(chunk_path, 'rb'), output_f)
```
mistydemeo: Use a `with` block?

storage_service/locations/models/duracloud.py
```diff
@@ -158,36 +253,93 @@ def move_to_storage_service(self, src_path, dest_path, dest_space):
...
+            content = etree.SubElement(header, 'sourceContent', contentId=relative_path)
+            etree.SubElement(content, 'mimetype').text = 'application/octet-stream'
+            etree.SubElement(content, 'byteSize').text = str(filesize)
+            etree.SubElement(content, 'md5').text = checksum.hexdigest()
+            chunks = etree.SubElement(root, 'chunks')
+            # Split file into chunks
+            with open(upload_file, 'rb') as f:
+                i = 0
+                chunk_data = f.read(self.CHUNK_SIZE)
+                while chunk_data:
```
mistydemeo, line 298: Nitpick, but do we need to use the filesystem as staging for these chunks? If we're already reading them into memory, can we just checksum and then upload them from memory instead?

storage_service/locations/models/duracloud.py
Hwesta: Good catch. I wonder if that's why downloading chunked files takes up so much memory?

storage_service/locations/models/duracloud.py
Hwesta, line 298: Good idea. Maybe I could put that in the duracloud-performance branch instead?

mistydemeo, line 298: Sure!

storage_service/locations/models/duracloud.py
```diff
@@ -129,16 +145,96 @@ def delete_path(self, delete_path):
...
+            LOGGER.debug('Response: %s', response)
+            if response.ok:
+                # Get list of file chunks
+                root = etree.fromstring(response.content)
+                to_delete = [e.attrib['chunkId'] for e in root.findall('chunks/chunk')]
+                to_delete.append(delete_path + self.MANIFEST_SUFFIX)
+                LOGGER.debug('Chunks to delete: %s', to_delete)
+            else:
+                # File cannot be found - this may be a folder
+                to_delete = self._get_files_list(delete_path, show_split_files=True)
```
mistydemeo, line 165: Is there any danger here if isn't a folder either (e.g. because it doesn't exist)?

mistydemeo:
:shipit:

storage_service/locations/models/duracloud.py
Hwesta, line 165: `_get_files_list` searches based on prefix and returns all those files, which may be an empty list.  As listed above, delete_path is expecting a folder that ends with / to prevent it from deleting similarly-named files (or other folders).

PR #57, opened 2015-01-21
Duracloud: set chunk size to 1GB

hakamine:
Workaround for a bug in the DuraCloud REST API that kicks in when the size
of the file being transferred is larger than 2^31 bytes (2.14 GB).
After the bug is fixed, it should be OK to restore to 5GB.


Hwesta:
:+1:

PR #59, opened 2015-01-23
Track file metadata from transfer METS when sending to transfer backlog

mistydemeo:
In artefactual/archivematica#153, the transfer METS has been augmented with additional metadata about the transfer and files, providing the information that ES also carries about them. This improves the File model to add the appropriate fields to store that metadata, then provides new endpoints to

a) List all files in a package, with their metadata;
b) Query for file metadata using the same names as in the transferfile ES index.

The latter can be used by Archivematica to look up specific records via absolute values from the SS, instead of querying the ES index.

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -454,6 +454,8 @@ class PackageResource(ModelResource):
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/pointer_file%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('pointer_file_request'), name="pointer_file_request"),
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/check_fixity%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('check_fixity_request'), name="check_fixity_request"),
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/send_callback/post_store%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('aip_store_callback_request'), name="aip_store_callback_request"),
+            url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/contents%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view("package_contents"), name="package_contents"),
+            url(r"^(?P<resource_name>%s)/metadata%s$" % (self._meta.resource_name, trailing_slash()), self.wrap_view("file_data"), name="file_data"),
 
             # FEDORA/SWORD2 endpoints
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/sword%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('sword_deposit'), name="sword_deposit"),
@@ -770,3 +772,95 @@ class PackageResource(ModelResource):
             status_code = 200
 
         return (status_code, response)
+
+    @_custom_endpoint(expected_methods=['get'])
+    def package_contents(self, request, bundle, **kwargs):
+        """
+        Returns metadata about every file within a specified Storage Service
+        package, specified via Storage Service UUID.
+
+        The file properties provided are the properties of the ~:class:`~locations.models.event.File` class; see the class definition for more information.
+
+        :returns: a JSON object in the following format:
+        {
+            "success": True,
+            "package": "UUID (as string)",
+            "files": [
+                # array containing zero or more objects containing
+                # all of the file's properties, in the format:
+                {
+                    "source_id": "",
+                    # ...
+                }
+            ]
+        }
+        """
+        response = {
+            "success": True,
+            "package": bundle.obj.uuid,
+            "files": []
+        }
+
+        for f in bundle.obj.file_set.all():
+            response["files"].append({attr: getattr(f, attr) for attr in ('source_id', 'name', 'source_package', 'checksum', 'accessionid', 'origin')})
+
+        return http.HttpResponse(status=200, content=json.dumps(response),
+            content_type='application/json')
+
+    def file_data(self, request, **kwargs):
+        """
+        Returns file metadata as a JSON array of objects.
+
+        This maps properties of the File class to the names of the
+        Elasticsearch indices' Transferfile index, allowing this to directly
+        substitute for Elasticsearch when looking up metadata on specific files.
+
+        Acceptable parameters are:
+            * relative_path (searches the `name` field)
+            * fileuuid (searches the `source_id` field)
+            * accessionid (searches the `accessionid` field)
+            * sipuuid (searches the `source_package` field)
+
+        :returns: an array of one or more objects. See the transferfile
+        index for information on the return format.
+        If no results are found for the specified query, returns 404.
+        If no acceptable query parameters are found, returns 400.
+        """
+        property_map = {
+            "relative_path": "name",
+            "fileuuid": "source_id",
+            "accessionid": "accessionid",
+            "sipuuid": "source_package"
+        }
+        query = {}
+        for source, dest in property_map.iteritems():
+            try:
+                query[dest] = request.GET[source]
+            except KeyError:
+                pass
+
+        if not query:
+            response = {
+                "success": False,
+                "error": "No supported query properties found!"
+            }
+            return http.HttpBadRequest(content=json.dumps(response),
+                content_type="application/json")
+
+        files = File.objects.filter(**query)
+        if not files.exists():
+            return http.HttpNotFound()
+
+        response = []
+        for f in files:
+            response.append({
+                "accessionid": f.accessionid,
+                "file_extension": os.path.splitext(f.name)[1],
+                "filename": os.path.basename(f.name),
+                "relative_path": f.name,
+                "fileuuid": f.source_id,
+                "origin": f.origin,
+                "sipuuid": f.source_package
+            })
+
+        return http.HttpResponse(content=json.dumps(response), content_type="application/json")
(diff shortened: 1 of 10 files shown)


storage_service/locations/migrations/0007_auto__add_field_file_package__add_field_file_source_package__add_field.py
```diff
@@ -0,0 +1,248 @@
+# -*- coding: utf-8 -*-
```
Hwesta: This should have a more useful name.  Migrations can be generated with names with `manage.py schemamigration <app> --auto <name>`; it will prefix the number automatically.

storage_service/locations/api/resources.py
```diff
@@ -769,3 +771,57 @@ def _attempt_package_request_event(self, package, request_info, event_type, even
...
+            "files": []
+        }
+
+        for f in bundle.obj.file_set.all():
+            response["files"].append({attr: getattr(f, attr) for attr in ('source_id', 'name', 'source_package', 'checksum', 'accessionid', 'origin')})
+
+        return http.HttpResponse(status=200, content=json.dumps(response),
+            content_type='application/json')
+
+    def file_data(self, request, **kwargs):
```
Hwesta, line 897: Docstrings for these? I know we don't have them on other functions, but they sure are nice.

storage_service/locations/models/package.py
```diff
@@ -559,6 +560,130 @@ def compress_package(self, algorithm, extract_path=None):
...
+                ]
+            }
+        :raises StorageException: if the requested METS file cannot be found,
+            or if required elements are missing.
+        """
+        if prefix is None:
+            prefix = self.full_path
+
+        mets_path = os.path.join(prefix, *relative_path)
+        if not os.path.exists(mets_path):
```
Hwesta: `if not os.path.isfile(mets_path):`?

storage_service/locations/models/package.py
```diff
@@ -559,6 +560,130 @@ def compress_package(self, algorithm, extract_path=None):
...
+            raise StorageException("No METS found at location: {}".format(mets_path))
+
+        doc = etree.parse(mets_path)
+
+        mets_ns = utils.NSMAP['mets']
```
Hwesta: Unnecessary, since mets_ns is only used once?

storage_service/locations/models/package.py
```diff
@@ -559,6 +560,130 @@ def compress_package(self, algorithm, extract_path=None):
...
+        namespaces = {'m': mets_ns}
+        mets = doc.xpath('/m:mets', namespaces=namespaces)
```
Hwesta, line 604: `doc.find` would return one element or None, allowing you to avoid mets[0].

storage_service/locations/models/package.py
```diff
@@ -559,6 +560,130 @@ def compress_package(self, algorithm, extract_path=None):
...
+            creation_date = header.attrib['CREATEDATE']
+        except KeyError:
+            raise StorageException("<metsHdr> element did not have a CREATEDATA attribute!")
+
+        accession_ids = header.xpath('./m:altRecordID[@TYPE="Accession number"]',
+                                     namespaces=namespaces)
+        if accession_ids:
+            accession_id = accession_ids[0].text
+        else:
+            accession_id = ""
```
Hwesta: `header.findtext` may be useful here.

storage_service/locations/models/package.py
```diff
@@ -589,6 +714,12 @@ def backlog_transfer(self, origin_location, origin_path):
             source_path=os.path.join(self.origin_location.relative_path, self.origin_path),
             destination_path=self.current_path,  # This should include Location.path
             destination_space=dest_space)
+
+        try:
+            self._index_file_data_from_transfer_mets(prefix=os.path.join(dest_space.staging_path, self.current_path))  # create File entries for every file in the transfer
+        except StorageException as e:
+            LOGGER.warning("Transfer METS data could not be read: %s", str(e))
```
Hwesta, line 730: If this happens, has it failed to be stored in backlog correctly?  Should we return a failed status code?  It can't be arranged if the Files table isn't populated, and if it can't be removed from backlog it's kinda useless.

storage_service/locations/api/resources.py
mistydemeo, line 897: Added!

storage_service/locations/migrations/0007_auto__add_field_file_package__add_field_file_source_package__add_field.py
mistydemeo: Updated to something descriptive.

storage_service/locations/models/package.py
mistydemeo, line 604: `doc.find` doesn't implement the full XPATH syntax (while `doc.xpath`) does, so not all of these queries work with `.find`. I'll change the ones that do though.

storage_service/locations/models/package.py
mistydemeo, line 730: Yeah, I'm a little torn on that one. You can still fetch it, just not necessarily in a useful or straightforward way.

mistydemeo:
Pushed some changes in response to feedback.

Hwesta:
SIP arrange fails with files that differ between sanitize and unsanitized, eg evelyn's photo.jpg vs evelyn_s_photo.jpg.  Dunno if fix belongs here or artefactual/archivematica#153

mistydemeo:
Ugh, I didn't consider that. Even though the `structMap` gets updated the `fileSec` doesn't, and the `fileSec` will have pre-sanitization names.

mistydemeo:
Updated: now uses the `amdSec` from the transfer METS to look up changed names. I've tested this with Unicode and SJIS-named files that get sanitized during the transfer.

storage_service/locations/models/package.py
```diff
@@ -559,6 +561,135 @@ def compress_package(self, algorithm, extract_path=None):
...
+            transfer_uuid = mets.attrib['OBJID']
+        except KeyError:
+            raise StorageException("<mets> element did not have an OBJID attribute!")
+
+        header = doc.xpath('/m:metsHdr', namespaces=namespaces)
+        if header is None:
+            raise StorageException("<metsHdr> element not found in METS file!")
+
+        try:
+            creation_date = header.attrib['CREATEDATE']
```
Hwesta, line 620: `header` is a list from xpath, so this fails.  `header = doc.find('m:metsHdr', namespaces=namespaces)` seems to work.

storage_service/locations/models/package.py
```diff
@@ -559,6 +561,135 @@ def compress_package(self, algorithm, extract_path=None):
...
+            prefix = self.full_path
+
+        mets_path = os.path.join(prefix, *relative_path)
+        if not os.path.isfile(mets_path):
+            raise StorageException("No METS found at location: {}".format(mets_path))
+
+        doc = etree.parse(mets_path)
+
+        namespaces = {'m': utils.NSMAP['mets'],
+                      'p': utils.NSMAP['premis']}
```
Hwesta, line 603: utils.NSMAP has all the namespaces I've used in this project.

storage_service/locations/models/package.py
Hwesta, line 730: We discussed this with Justin - what was the decision?  To fail loudly?

Hwesta:
Tests for the new endpoints?

storage_service/locations/models/package.py
mistydemeo, line 603: Yes, I was just creating an nsmap with shorter names - the longer xpath queries were getting quite long, so I found it easier to read with brief prefixes.

storage_service/locations/models/package.py
mistydemeo, line 620: Fixed.

mistydemeo:
Added tests for the new endpoints and for the METS parsing, and rebased those into their respective commits.

storage_service/locations/tests/test_models.py
```diff
@@ -1,9 +1,34 @@
+import os
 
 from django.test import TestCase
 
 from locations import models
 
 
+class TestPackage(TestCase):
```
Hwesta: I had started separating the tests as well (see test_api, test_duracloud, test_locations).  Move this to test_package.py?  Eventually we can rename test_models to test_lockss, sinces that's what it really is...

:tada: Tests! :tada:

storage_service/locations/api/resources.py
```diff
@@ -769,3 +773,176 @@ def _attempt_package_request_event(self, package, request_info, event_type, even
...
+    def remove_files_from_package(self, request, bundle, **kwargs):
+        """
+        Removes all file records associated with this package.
+        """
+
+        bundle.obj.file_set.all().delete()
+        return http.HttpNoContent()
+
+    @_custom_endpoint(expected_methods=['post'])
+    def add_files_to_package(self, request, bundle, **kwargs):
```
Hwesta: Rather than having add_files and remove_files, why not have file/<uuid>/files/ that responds to DELETE or PUT? Seems more RESTful.

mistydemeo: Sure, will update.

mistydemeo: Updated!

storage_service/locations/tests/test_models.py
mistydemeo: Moved into test_package.py!

storage_service/locations/migrations/0007_file_add_new_fields.py
```diff
@@ -0,0 +1,248 @@
...
+                      keep_default=False)
+
+        # Adding field 'File.accessionid'
+        db.add_column(u'locations_file', 'accessionid',
+                      self.gf('django.db.models.fields.TextField')(default='', blank=True),
+                      keep_default=False)
+
+        # Adding field 'File.origin'
+        db.add_column(u'locations_file', 'origin',
+                      self.gf('django.db.models.fields.CharField')(default='', unique=False, max_length=36, blank=True),
```
Hwesta: I get a slightly different migration when I generate it. This column is not unique.  Perhaps this file needs to be re-generated?

storage_service/locations/models/package.py
```diff
@@ -589,6 +722,12 @@ def backlog_transfer(self, origin_location, origin_path):
             source_path=os.path.join(self.origin_location.relative_path, self.origin_path),
             destination_path=self.current_path,  # This should include Location.path
             destination_space=dest_space)
+
+        try:
+            self.index_file_data_from_transfer_mets(prefix=os.path.join(dest_space.staging_path, self.current_path))  # create File entries for every file in the transfer
+        except StorageException as e:
+            LOGGER.warning("Transfer METS data could not be read: %s", str(e))
```
Hwesta, line 730: Is this suppose to raise?

mistydemeo, line 730: No - I left this condition here to allow the newer storage service to continue serving older AMs.

Hwesta, line 730: Ahhhh, okay. Will those SIPs be arrange-able?

mistydemeo, line 730: Within that AM version, yep! After an upgrade to Archivematica 1.4, they'll need to be reindexed in the storage service using the script that's there.

Hwesta, line 730: Ah, okay. Makes sense.

Hwesta:
:+1:

storage_service/locations/migrations/0007_file_add_new_fields.py
mistydemeo: Re-generated.

PR #60, opened 2015-01-29
Improve default location creation

Hwesta:
Previously, when creating default locations, we checked if a Location with the exact same attributes (path, description, etc) already existed, and created one if it didn't.  This causes problems when a Location with that purpose already exists but has been modified, eg storage service internal, since there can only be one.  Update to only look for a Location with the same purpose.

To support disabling Locations instead of deleting them, ensure queries against Location use 'active' not 'objects'.

Add tests.

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -325,7 +325,7 @@ class LocationResource(ModelResource):
 
         # Get the object for this endpoint
         try:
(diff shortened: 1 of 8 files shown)


mistydemeo:
:+1: Really nice to see the tests here too!

Hwesta:
Added logging of Tastypie errors, so debugging this sucks less next time.

mistydemeo:
:+1: \* :100:

PR #63, opened 2015-02-03
Update packaging scripts

jhsimpson:
updating debian package scripts, to fix issue 7916, as well as update version number to 0.6.0

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -37,7 +37,7 @@ chown -R archivematica:archivematica /var/archivematica/storage-service
 chown -R archivematica:archivematica /var/archivematica/storage_service
 chown -R archivematica:archivematica /var/archivematica/.storage-service
 chown -R archivematica:archivematica /usr/share/python/archivematica-storage-service
-chown -R archivematica:archivematica /var/log/archivematica
+chown -R archivematica:archivematica /var/log/archivematica/storage-service
 
 rm -f /tmp/storage_service.log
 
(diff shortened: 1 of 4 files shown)


debian/postinst
```diff
@@ -31,12 +32,12 @@ echo "configuring django database and static files"
 /usr/share/python/archivematica-storage-service/bin/python manage.py migrate
 /usr/share/python/archivematica-storage-service/bin/python manage.py collectstatic --noinput
 
-
 echo "updating directory permissions"
 chown -R archivematica:archivematica /var/archivematica/storage-service
 chown -R archivematica:archivematica /var/archivematica/storage_service
 chown -R archivematica:archivematica /var/archivematica/.storage-service
 chown -R archivematica:archivematica /usr/share/python/archivematica-storage-service
+chown -R archivematica:archivematica /var/log/archivematica
```
mistydemeo: Do we want to pre-emptively set this to `g+ws` to prepare for artefactual/archivematica#34?

jhsimpson: I'm not sure.  There is no guarantee that Archivematica itself will get installed on the same machine as the storage service.  Set gid is probably not a good idea for the storage service, I think?  Only the one user (archivematica) should be able to write to the storage service logs..  For Archivematica the situation is a little different because of the archivematicadashboard user.  The storage service would prefer g+rs I think?  

Maybe the storage service should be logging to /var/log/archivematica/storage-service, and the dashboard to /var/log/archivematica/dashboard?  We were basing this layout on things like nginx and elasticsearch, which do not use setgid.

mistydemeo: That sounds like a good idea, actually - I like the idea of separating the log paths for the different components into subdirectories; better than setting gid.

Hwesta:
Looks good to me.

jhsimpson:
I changed the default location of storage service logs in efc6e99.  I will merge this now.

PR #64, opened 2015-02-18
Browse: Return additional info

Hwesta:
Return more information in a 'properties' entry in the browse dict about the browse entries. Currently, returns number of objects in a directory and (for local filesystems) file size.

refs #7853


storage_service/locations/models/space.py
```diff
@@ -143,16 +143,31 @@ def get_child_space(self):
...
+        Return information about the objects (files, directories) at `path`.
+
+        Attempts to call the child space's implementation.  If not found, falls
+        back to looking for the path locally.
+
+        Returns a dictionary with keys 'entries', 'directories' and 'properties'.
+
+        'entries' is a list of strings, one for each entry in that directory, both file-like and folder-like.
+        'directories' is a list of strings for each folder-like entry. Each entry should also be listed in 'entries'.
+        'properties' is a dictionary that may contain additional information for the entries.  Keys are the entry name found in 'entries', values are a dictionary containing extra information. 'properties' may not contain all values from 'entries'.
```
mistydemeo, line 156: I know it's provided in the sample below, but mind explicitly documenting what properties currently exist and when they're reported?

storage_service/locations/models/space.py
```diff
@@ -351,6 +366,15 @@ def _create_local_directory(self, path, mode=None):
         except os.error as e:
             LOGGER.warning(e)
 
+    def _count_objects_in_directory(self, path):
+        """
+        Returns all the files in a directory, including children.
+        """
+        total_files = 0
```
mistydemeo, line 407: Sorry to golf, but could reduce this to a one-liner via `reduce`:

``` python
return reduce(lambda count, (dir, dirs, files): count + len(files), os.walk(path), 0)
```

mistydemeo:
:+1:

mistydemeo:
Though, this is making me think... I wonder if it would make sense to send these transfer components as partial, or complete, bags? Bags contain a metadata file that lets you know the expected contents (numbers of files/directories, expected sizes), which are the same properties we're sending as a part of the JSON response here.

Hwesta:
Hmm, maybe. On the other hand, we may want to send other things not stored in a bag back, like the last modified timestamp. Automation tools will use that to only run new or changed transfers through.

storage_service/locations/models/space.py
Hwesta, line 156: I had anticipated 'properties' being more free-form than that, since not all Spaces will provide all properties (eg Duracloud doesn't do file size though LocalFS does, so I'm hesitant to provide a definitive list. The consuming code should handle if the data isn't there.  Do you still think it'd be helpful to provide a partial list with caveats?

mistydemeo, line 156: Sure, that'd help! I'm looking at this as a consumer of the data later, wanting to know what kind of data is actually going to be available - otherwise I wouldn't be sure what I could do with it.

If we could document in the spaces what properties they can provide that would also be helpful!

mistydemeo:
If we need extra properties that bags don't provide, then yeah, makes sense not to use that.

storage_service/locations/models/space.py
Hwesta, line 156: Added info about the properties that are/can be returned.

PR #66, opened 2015-02-24
Preserve original file timestamps

jhsimpson:
Tested this change just now, files in a transfer source location keep their original timestamps when they are copied to the processing location to start a transfer, and are maintained all the way through to the METS file in the final aip.


mistydemeo:
:+1:

Hwesta:
:+1: 

Note this will only preserve timestamps for Local Filesystem, NFS and Pipeline Local Filesystem, but will not affect Duracloud or FEDORA via SWORD2.

PR #67, opened 2015-02-27
Tests: Setup default testing config in pytest.ini

Hwesta:
Add Django settings and a useful PYTHONPATH to the pytest.ini file.

Unlike Archivematica, the Storage Service doesn't create symlinks to itself in a fixed location.  I set the python_path to /usr/lib/archivematica/storage-service/ which is where the debian package installs it, and ran `ln -s /path/to/archivematica-storage-service/storage_service storage-service` from `/usr/lib/archivematica/`  Please test this and tell me if it actually makes running tests easier.

### DIFF
--- a/storage_service/pytest.ini
+++ b/storage_service/pytest.ini
@@ -1,0 +1,4 @@
+[pytest]
+django_find_project=false
+python_paths=/usr/lib/archivematica/storage-service/
+DJANGO_SETTINGS_MODULE=storage_service.settings.local
--- a/requirements/local.txt
+++ b/requirements/local.txt
@@ -7,4 +7,5 @@ Sphinx==1.2b1
 ipython==1.1.0
 pytest
 pytest-django
+pytest-pythonpath
 vcrpy>=1.0.0


storage_service/pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+django_find_project=false
+python_paths=/usr/lib/archivematica/storage-service/
+DJANGO_SETTINGS_MODULE=storage_service.settings.local
```
Hwesta, line 4: I tried setting this to `storage_service.settings.test`, but ran into http://stackoverflow.com/questions/19958587/django-test-in-memory-sqllite-missing-tables-during-test-run because of our startup() code in urls.py being run on import.  Django 1.7 should fix that, and then we can switch this to using the test settings.

Hwesta:
This doesn't currently set environment variables, so things like SS_DB_NAME, SS_DB_USER etc need to already be set (probably by the virtualenv).  How much to we care about fixing that?

mistydemeo:
I think it's reasonable to expect the virtualenv to take care of that.

mistydemeo:
Even with the `/usr/lib/archivematica/storage-service` symlink created, running `py.test` errors out due to not being able to find the settings module:

> _pytest.config.UsageError: Could not import settings 'storage_service.settings.local' (Is it on sys.path?): No module named settings.local

mistydemeo:
Never mind, I'd symlinked the root of the repo instead of `storage_service` from the repo.

PR #69, opened 2015-03-04
Post-move fixes

Hwesta:
These have been languishing in the Arkivum branch for a while and I'd like them in qa.
- post_move_from takes the source and destination path and does the same path munging that move_from does
- Staging copy is deleted after post_move_from
- Package staging vs uploaded set better.

### DIFF
--- a/storage_service/locations/models/space.py
+++ b/storage_service/locations/models/space.py
@@ -216,6 +216,34 @@ class Space(models.Model):
             # This is optional for the child class to implement
             pass
 
+    def _move_from_path_mangling(self, staging_path, destination_path):
+        """
+        Does path pre-processing before passing to move_from_* functions.
+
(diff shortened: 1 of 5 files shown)


storage_service/locations/models/space.py
```diff
@@ -216,6 +216,22 @@ def post_move_to_storage_service(self, *args, **kwargs):
             # This is optional for the child class to implement
             pass
 
+    def _move_from_path_mangling(self, staging_path, destination_path):
```
mistydemeo, line 219: I know this is a private function, but mind adding a docstring that provides examples of what the mangling looks like?

mistydemeo:
:+1: aside from comment

storage_service/locations/models/space.py
Hwesta, line 219: Added

PR #70, opened 2015-03-09
Package.check_fixity: fix when scanning uncompressed AIP

mistydemeo:
Fixes #7777


storage_service/locations/models/package.py
```diff
@@ -642,7 +643,7 @@ def check_fixity(self, delete_after=True):
             failures = failure.details
             message = failure.message
 
-        if delete_after and (self.local_path_location != self.current_location or self.local_path != self.full_path):
+        if temp_dir is not None and delete_after and (self.local_path_location != self.current_location or self.local_path != self.full_path):
```
Hwesta: Just `if temp_dir and` ?

Hwesta:
:+1:

mistydemeo:
Updated, and added a changelog.

PR #72, opened 2015-03-23
Replace -a flag in rsync to avoid CIFS errors ref. #8124

hakamine:
The purpose of using rsync with -a was to preserve file
timestamps when copying files from/to the storage service.
rsync -a attemps to preserve timestamps and permissions of copied
files and directories. However, on a CIFS mounted volume it
seems not possible to set timestamps and permissions of
copied directories, producing a non-zero return code.
Replace -a with -t and -O. -t will preserve modification
times. -O will avoid to preserve modification times in
copied directories.

### DIFF
--- a/storage_service/locations/models/space.py
+++ b/storage_service/locations/models/space.py
@@ -321,7 +321,7 @@ class Space(models.Model):
 
         # Rsync file over
         # TODO Do this asyncronously, with restarting failed attempts
-        command = ['rsync', '-a', '--protect-args', '-vv', '--chmod=ugo+rw', '-r', source, destination]
+        command = ['rsync', '-t', '-O', '--protect-args', '-vv', '--chmod=ugo+rw', '-r', source, destination]
         LOGGER.info("rsync command: %s", command)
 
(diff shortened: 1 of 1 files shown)


mistydemeo:
Looks good to me; I think omitting directory times is a fair tradeoff, but @jhsimpson may want to comment on that.

jhsimpson:
I think this is a reasonable trade off, in the short term.  In the long term, it would be good to look at being able to have different parameters for rsync based on the protocol of the Space.

mistydemeo:
We might want to look at tracking more information related to that in the future. The CIFS mount where we encountered this issue was a `LocalFilesystem` space - there wasn't an immediate way to determine what kind of filesystem it is, and what features it supports.

jhsimpson:
Yes, fair point, there is a redmine ticket to add CIFS as a Space protocol, but it has never been implemented.  In any case, it is future, unspecified work, this PR is good, I give it :+1:

mistydemeo:
Merged into qa/0.x, and backported into stable/0.6.x and the relevant client branch.

PR #73, opened 2015-04-06
Browse: Return properties for pipeline local FS

Hwesta:
Return size and timestamp for pipeline local FS, extracted from rsync output.

Supports artefactual/automation-tools#2 and artefactual/archivematica#188


mistydemeo:
:+1:

PR #74, opened 2015-04-15
Space: use utils.coerce_str, not str, before rmtree

mistydemeo:
Fixes #8285.


Hwesta:
:+1: 

Are there other places that should be coerced too, while you're fixing it?

mistydemeo:
I did a brief grep and didn't find any others, fortunately.

PR #75, opened 2015-06-04
Browse: Limit local filesystem walking

Hwesta:
When counting the number of files in a local browse, limit the number to prevent the call from taking too long.

### DIFF
--- a/storage_service/locations/models/space.py
+++ b/storage_service/locations/models/space.py
@@ -464,6 +464,9 @@ class Space(models.Model):
         total_files = 0
         for _, _, files in os.walk(path):
             total_files += len(files)
+            # Limit the number of files counted to keep it from being too slow
+            if total_files > 5000:
+                return '5000+'
         return total_files
 
     def _browse_local(self, path):


mistydemeo:
Works for me. How long are we talking about? Is there an upper limit on the acceptable time?

In testing I found that shelling out to `find PATH -type f -print | wc -l` is twice as fast as running the `os.walk` loop or a `reduce` equivalent; if that's fast enough for large sets of files we may be able to switch to something like that instead and keep the accurate result?

Hwesta:
I like that idea, and using find is much faster (for 198K files on local hard drive it's ~0.4 seconds vs ~1.2 seconds).  However, for a very large filesysystem and/or unreliably connection (eg NFS) there's no way to stop it if it takes too long.  Because of that, I'm inclined to leave this fix the way it is for now.

mistydemeo:
Sounds good! :+1:

Hwesta:
I also merged this into qa/0.7.0

PR #76, opened 2015-06-11
Set directory permissions when creating with rsync

Hwesta:
Also a couple of fixes.


Hwesta:
This is also in the dev/issue-8540-pipefs-create-dir-perms-0.7 branch based on qa/0.7.0

mistydemeo:
:+1:

PR #77, opened 2015-07-06
AIP re-ingest

Hwesta:
Add re-ingest functionality to the storage service
- Add re-ingest button to package page
- Add endpoint for Archivematica to start a re-ingest
- Extract and copy files to Archivematica to start re-ingest
- Accept a reingested package, updates/replaces METS & metadata files, and stores
- Known bug: Does not support uncompressed AIPs.  Does support .tar.bz2 & .zip

Companion to artefactual/archivematica#257

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -852,42 +852,34 @@ class Package(models.Model):
                 'message': 'This AIP already being reingested on {}'.format(self.misc_attributes['reingest_pipeline'])}
         self.misc_attributes.update({'reingest_pipeline': pipeline.uuid})
 
-        # Copy to storage service internal location
-        # Mostly because we probably need to extract it, and can't extract from
-        # remote to SS internal location
-        ss_internal = Location.objects.get(purpose=Location.STORAGE_SERVICE_INTERNAL)
-        relative_path = os.path.basename(os.path.normpath(self.current_path))
-        self.current_location.space.move_to_storage_service(
-            source_path=os.path.join(self.current_location.relative_path, self.current_path),
-            destination_path=relative_path,
-            destination_space=ss_internal.space,
-        )
-        ss_internal.space.move_from_storage_service(
-            source_path=relative_path,
-            destination_path=os.path.join(ss_internal.relative_path, relative_path)
-        )
-        dest_path = os.path.join(ss_internal.full_path, relative_path)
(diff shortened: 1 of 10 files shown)


storage_service/locations/models/package.py
```diff
@@ -820,6 +829,409 @@ def delete_from_storage(self):
...
+            LOGGER.debug('Extract file RC: %s', rc)
+            # Get output path
+            command = ['lsar', '-ja', reingest_full_path]
+            try:
+                output = subprocess.check_output(command)
+                j = json.loads(output)
+                bname = sorted([d['XADFileName'] for d in j['lsarContents'] if d.get('XADIsDirectory', False)], key=len)[0]
+            except (subprocess.CalledProcessError, ValueError):
+                # FIXME
+                print 'I dunno what to do here'
```
mistydemeo: :scream_cat:

storage_service/locations/models/package.py
```diff
@@ -820,6 +829,409 @@ def delete_from_storage(self):
...
+                original_metadata_dir)
+
+        # Update bag payload and verify
+        bag = bagit.Bag(path)
+        bag.save(manifests=True)
+        bag.validate()  # Raises exception in case of problem
+
+        # Compress
+        # TODO update this for uncompressed AIPs
+        reingest_root = etree.parse(reingest_pointer)
```
mistydemeo, line 1048: Maybe handle the absence of a pointer file as meaning uncompressed AIP?

mistydemeo:
Do we want to keep a copy of the original AIP in case something goes wrong with the reingestion process?

storage_service/locations/models/package.py
Hwesta, line 1048: A good idea, but I think out of scope for the moment.

Hwesta:
> Do we want to keep a copy of the original AIP in case something goes wrong with the reingestion process?

I'd be more comfortable with that, but there are storage concerns. We don't currently make any changes to the original objects, and the AIP in storage isn't modified until after it's repackaged with new METS & metadata.  Perhaps an AIP validation tool, but I think that's out of scope currently.

storage_service/locations/models/package.py
Hwesta: Fixed with default value and warning log.

PR #79, opened 2015-08-04
Libffi is a runtime not build dependency

Hwesta:


mistydemeo:
:+1:

mistydemeo:
Wait - I'm confused. I thought that we installed dependencies as part of `dh_virtualenv`.

mistydemeo:
e.g. during the buildphase.

Hwesta:
Not the correct fix for this problem, see artefactual/archivematica#264

PR #80, opened 2015-08-28
Space: don't log exc_info when falling back to browse local

mistydemeo:
The traceback isn't generally useful here, since the exception being caught is very specific already. The exception is also misleading; we've encountered a number of users having problems who believed that because an exception was printed here, it was the actual cause of their problems when it was actually unrelated.


jhsimpson:
:+1:

Hwesta:
Context for this was that I had an AttributeError bubbling up from elsewhere that got caught by this, so I added the debug log so I could see what was going on.  Since it was confusing for users deleting it makes sense.

PR #81, opened 2015-08-31
recover_aip: post_move_from_storage_service params are now mandatory

mistydemeo:
The parameters to `post_move_from_storage_service` were made mandatory in 7f72eeb8eb8ed58d79fa3c29e995abd4f6a63f99; however, not all code that calls this method was updated to reflect that.

`recover_aip` calls `post_move_from_storage_service` without parameters; calling it with both parameters as `None` should behave the same as it did before the method was changed.

This was reported on the mailing list.


Hwesta:
:+1:

PR #82, opened 2015-09-02
Dev/issue 8635 fixity logging

mcantelon:
This adds:
- failure handling if the fixity check doesn't complete because of an exception
- per-AIP log model of fixity attempts (time, whether they're successful, and error/failure details)
- two columns to package model for latest fixity attempt datetime/result
- two columns in package hitlist for latest fixity attempt datetime/result
- a fixity failures page where fixity check history for a package can be viewed

I'm not sure how you folk merge (whether you squash or merge individual commits), but I can clean up the history and put in better commit messages if individual commits are merged.

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -714,6 +714,10 @@ class PackageResource(ModelResource):
             signals.failed_fixity_check.send(sender=self,
                 uuid=bundle.obj.uuid, location=bundle.obj.full_path,
                 report=report)
+        else:
+            signals.successful_fixity_check.send(sender=self,
+                uuid=bundle.obj.uuid, location=bundle.obj.full_path,
+                report=report)
 
         return http.HttpResponse(
             report,
--- a/storage_service/locations/models/fixity_log.py
+++ b/storage_service/locations/models/fixity_log.py
@@ -1,0 +1,27 @@
+# stdlib, alphabetical
+
+# Core Django, alphabetical
+from django.db import models
+
+# Third party dependencies, alphabetical
+
+# This project, alphabetical
+
+# This module, alphabetical
+
+
+class FixityLog(models.Model):
+    """ Stores fixity check success/failure and error details """
+
+    package = models.ForeignKey('Package', to_field='uuid')
+    success = models.BooleanField(default=False)
+    error_details = models.TextField(null=True)
+    datetime_reported = models.DateTimeField(auto_now=True)
+
+    class Meta:
+        verbose_name = "Fixity Log"
+        app_label = 'locations'
+
+    def __unicode__(self):
+        return u"Fixity check of {package}".format(
+            package=self.package)
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -27,6 +27,7 @@ from . import StorageException
 from location import Location
 from space import Space
 from event import File
+from fixity_log import FixityLog
 
 __all__ = ('Package', )
 
@@ -162,6 +163,22 @@ class Package(models.Model):
             else:
                 message = "{} is neither a file nor a directory".format(full_path)
             raise StorageException(message)
+
+    @property
+    def latest_fixity_check_datetime(self):
+        latest_check = self._latest_fixity_check()
+        return latest_check.datetime_reported if latest_check is not None else None
+
+    @property
+    def latest_fixity_check_result(self):
+        latest_check = self._latest_fixity_check()
+        return latest_check.success if latest_check is not None else None
+
+    def _latest_fixity_check(self):
+        try:
+            return FixityLog.objects.filter(package=self).order_by('-datetime_reported')[0]  # limit 1
+        except IndexError:
+            return None
 
     def get_download_path(self, lockss_au_number=None):
         full_path = self.fetch_local_path()
--- a/storage_service/locations/signals.py
+++ b/storage_service/locations/signals.py
@@ -1,12 +1,16 @@
+import json
 import logging
 from django.dispatch import receiver, Signal
 from django.contrib.auth.models import User
 from django.core.urlresolvers import reverse
 
+import models
+
 LOGGER = logging.getLogger(__name__)
 
 deletion_request = Signal(providing_args=["uuid", "location", "url"])
 failed_fixity_check = Signal(providing_args=["uuid", "location", "report"])
+successful_fixity_check = Signal(providing_args=["uuid", "location", "report"])
 
 
 def _notify_administrators(subject, message):
@@ -35,8 +39,16 @@ To approve this deletion request, visit: {}{}
     _notify_administrators(subject, message)
 
 
+def _log_report(uuid, success, message=None):
+    package = models.Package.objects.get(uuid=uuid)
+    models.FixityLog.objects.create(package=package, success=success, error_details=message)
+
+
 @receiver(failed_fixity_check, dispatch_uid="fixity_check")
 def report_failed_fixity_check(sender, **kwargs):
+    report_data = json.loads(kwargs['report'])
+    _log_report(kwargs['uuid'], False, report_data['message'])
+
     subject = "Fixity check failed for package {}".format(kwargs["uuid"])
     message = """
 A fixity check failed for the package with UUID {}. This package is currently stored at: {}
@@ -44,4 +56,10 @@ A fixity check failed for the package with UUID {}. This package is currently st
 Full failure report (in JSON format):
 {}
 """.format(kwargs["uuid"], kwargs["location"], kwargs["report"])
+
     _notify_administrators(subject, message)
+
+
+@receiver(successful_fixity_check, dispatch_uid="fixity_check")
+def report_successful_fixity_check(sender, **kwargs):
+    _log_report(kwargs['uuid'], True)
--- a/storage_service/locations/tests/test_fixity_log.py
+++ b/storage_service/locations/tests/test_fixity_log.py
@@ -1,0 +1,18 @@
+from django.test import TestCase
+
+from locations import models
+
+
+class TestFixityLog(TestCase):
+
+    fixtures = ['base.json', 'fixity_log.json']
+
+    def setUp(self):
+        self.fl_object = models.FixityLog.objects.all()[0]
+        #self.auth = requests.auth.HTTPBasicAuth(self.ds_object.user, self.ds_object.password)
+
+    def test_has_required_attributes(self):
+        assert self.fl_object.package
+        assert self.fl_object.success
+        assert self.fl_object.error_details
+        assert self.fl_object.datetime_reported
--- a/storage_service/locations/views.py
+++ b/storage_service/locations/views.py
@@ -14,7 +14,7 @@ from django.views.decorators.csrf import csrf_exempt
 
 from common import decorators
 from common import utils
-from .models import Callback, Space, Location, Package, Event, Pipeline, LocationPipeline, StorageException
+from .models import Callback, Space, Location, Package, Event, Pipeline, LocationPipeline, StorageException, FixityLog
 from . import forms
 from .constants import PROTOCOL
 
@@ -43,8 +43,13 @@ def get_delete_context_dict(request, model, object_uuid, default_cancel='/'):
 ########################## FILES ##########################
 
 def package_list(request):
-    packages = Package.objects.all()
-    return render(request, 'locations/package_list.html', locals())
+    context = {'packages': Package.objects.all()}
+    return render(request, 'locations/package_list.html', context)
+
+def package_fixity(request, package_uuid):
+    log_entries = FixityLog.objects.filter(package__uuid=package_uuid).order_by('-datetime_reported')
+    context = {'log_entries': log_entries}
+    return render(request, 'locations/fixity_results.html', context)
 
 class PackageRequestHandlerConfig:
     event_type = ''                # Event type being handled
--- a/storage_service/templates/locations/fixity_results.html
+++ b/storage_service/templates/locations/fixity_results.html
@@ -1,0 +1,28 @@
+{% extends "base.html" %}
+
+{% block page_title %}Fixity Checks{% endblock %}
+
+{% block content %}
+
+{% if log_entries %}
+  <table class="datatable">
+    <thead>
+      <tr>
+        <th>Date</th>
+        <th>Error</th>
+      </tr>
+    </thead>
+    <tbody>
+    {% for entry in log_entries %}
+      <tr>
+        <td>{{ entry.datetime_reported }}</td>
+        <td>{{ entry.error_details }}</td>
+      </tr>
+    {% endfor %}
+    </tbody>
+  </table>
+{% else %}
+  <p>No fixity checks found.</p>
+{% endif %}
+
+{% endblock %}
--- a/storage_service/templates/snippets/packages_table.html
+++ b/storage_service/templates/snippets/packages_table.html
@@ -9,6 +9,8 @@
         <th>Type</th>
         <th>Pointer File</th>
         <th>Status</th>
+        <th>Fixity Date</th>
+        <th>Fixity Status</th>
         <th>Actions</th>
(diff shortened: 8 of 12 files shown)


storage_service/locations/api/resources.py
```diff
@@ -670,7 +671,12 @@ def pointer_file_request(self, request, bundle, **kwargs):
 
     @_custom_endpoint(expected_methods=['get'])
     def check_fixity_request(self, request, bundle, **kwargs):
-        success, failures, message = bundle.obj.check_fixity()
+        try:
+            success, failures, message = bundle.obj.check_fixity()
+        except:
```
Hwesta: Only catch the exception(s) you're expecting here, or at a minimum `except Exception`.  Maybe a logging message with a traceback too?

storage_service/locations/models/fixity_log.py
```diff
@@ -0,0 +1,28 @@
...
+# This module, alphabetical
+
+
+class FixityLog(models.Model):
+    """ Stores fixity check success/failure and error details """
+
+    package = models.ForeignKey('Package', to_field='uuid')
+    success = models.BooleanField(default=False)
+    error_details = models.TextField()
+    time_reported = models.DateTimeField(auto_now=True)
```
Hwesta: Nitpick, but the `DateTimeField` is called `time` here and `date` in Package. For consistency, call them both the same thing. Maybe datetime?

storage_service/locations/signals.py
```diff
@@ -44,4 +47,26 @@ def report_failed_fixity_check(sender, **kwargs):
...
+        log.package = package
+        report_data = json.loads(kwargs['report'])
+        log.error_details = report_data['message']
+        log.save()
+
+        package.last_fixity_check_result = False
+        package.last_fixity_check_date = log.time_reported
+        package.save()
+    except:
+        LOGGER.exception("Error logging fixity failure.")
```
Hwesta: Is this an ERROR level log or something lesser?  Tracebacks can be added to any log message with [exc_info](https://docs.python.org/2/library/logging.html#logging.debug)

storage_service/locations/signals.py
```diff
@@ -44,4 +47,26 @@ def report_failed_fixity_check(sender, **kwargs):
...
+
     _notify_administrators(subject, message)
+
+@receiver(successful_fixity_check, dispatch_uid="fixity_check")
+def report_successful_fixity_check(sender, **kwargs):
+    package = Package.objects.get(uuid=bundle.obj.uuid)
+    package.last_fixity_check_result = True
+    package.last_fixity_check_date = datetime.datetime.now()
+    package.save()
```
Hwesta: Should a FixityLog be created here as well?

storage_service/locations/views.py
```diff
@@ -46,6 +46,11 @@ def package_list(request):
     packages = Package.objects.all()
     return render(request, 'locations/package_list.html', locals())
 
+def package_fixity(request, uuid):
+    package = Package.objects.get(uuid=uuid)
+    log_entries = FixityLog.objects.filter(package=package)
```
Hwesta: This is easier with `FixityLog.objects.filter(package__uuid=uuid)` unless you need the package as well.  See https://docs.djangoproject.com/en/dev/topics/db/queries/#lookups-that-span-relationships for details.  Objects also store [reverse-foreign-key-sets](https://docs.djangoproject.com/en/dev/topics/db/queries/#following-relationships-backward).

storage_service/locations/views.py
```diff
@@ -46,6 +46,11 @@ def package_list(request):
     packages = Package.objects.all()
     return render(request, 'locations/package_list.html', locals())
 
+def package_fixity(request, uuid):
```
Hwesta: Be wary of shadowing common libraries like `uuid`, since it can cause weird bugs later.

storage_service/locations/views.py
```diff
@@ -46,6 +46,11 @@ def package_list(request):
...
+    package = Package.objects.get(uuid=uuid)
+    log_entries = FixityLog.objects.filter(package=package)
+    return render(request, 'locations/fixity_results.html', locals())
```
Hwesta: I know it's super convenient, but we're trying to move away from using `locals()` and towards an explicit dict of the variables used.

storage_service/locations/models/package.py
```diff
@@ -91,6 +91,10 @@ class Package(models.Model):
     misc_attributes = jsonfield.JSONField(blank=True, null=True, default={},
         help_text='For storing flexible, often Space-specific, attributes')
 
+    # Provide access, without join, to last fixity check result/date
+    last_fixity_check_result = models.NullBooleanField()
+    last_fixity_check_date = models.DateTimeField(blank=True, null=True)
```
Hwesta: Are these duplicates of `FixityLog.[success|time_reported]`?  I'm concerned about them getting out of sync.  Maybe make it a foreign key and look at this performance tweak if it becomes a problem?  Are there case where we need this info but no FixityLog object exists?

Hwesta:
> I'm not sure how you folk merge (whether you squash or merge individual commits), but I can clean up the history and put in better commit messages if individual commits are merged.

Generally we do a rebase on top of qa/0.x, then merge it as a fast-forward merge. A cleaner commit history would be appreciated.  https://github.com/artefactual/archivematica/blob/code_review_guide/code_review.md maybe useful as well.

storage_service/locations/signals.py
```diff
@@ -1,8 +1,11 @@
...
 import logging
 from django.dispatch import receiver, Signal
 from django.contrib.auth.models import User
 from django.core.urlresolvers import reverse
 
+import models
+
 LOGGER = logging.getLogger(__name__)
 
 deletion_request = Signal(providing_args=["uuid", "location", "url"])
```
Hwesta, line 11: `successful_fixity_check` needs to be added here.

Hwesta:
I strongly recommend installing a linter (probably [flake8](https://flake8.readthedocs.org/en/2.4.1/)) in your text editor, since it caught the undefined variable above.

Also, we have tests now!  Run `py.test --ignore locations/tests/test_models.py` from the `storage_service` directory

storage_service/locations/api/resources.py
```diff
@@ -670,7 +671,12 @@ def pointer_file_request(self, request, bundle, **kwargs):
...
+            success = False
+            failures = []
+            message = 'Internal error'
```
mistydemeo: Probably good to a) log the error, b) include some additional information in the response. We don't need to expose too much internal app state in the API response, but it'd be good to get an idea _why_ it didn't work.

storage_service/locations/signals.py
```diff
@@ -44,4 +47,26 @@ def report_failed_fixity_check(sender, **kwargs):
...
+        log = models.FixityLog()
+        log.package = package
+        report_data = json.loads(kwargs['report'])
+        log.error_details = report_data['message']
+        log.save()
+
+        package.last_fixity_check_result = False
+        package.last_fixity_check_date = log.time_reported
+        package.save()
+    except:
```
mistydemeo: Like before, are there specific exceptions to catch here?

storage_service/locations/signals.py
```diff
@@ -44,4 +48,23 @@ def report_failed_fixity_check(sender, **kwargs):
 Full failure report (in JSON format):
 {}
 """.format(kwargs["uuid"], kwargs["location"], kwargs["report"])
+    package = models.Package.objects.get(uuid=kwargs['uuid'])
+
+    log = models.FixityLog()
+    log.package = package
+    report_data = json.loads(kwargs['report'])
```
Hwesta: Is this supposed to be used?

storage_service/locations/signals.py
```diff
@@ -44,4 +48,23 @@ def report_failed_fixity_check(sender, **kwargs):
...
+    log.error_details = report_data['message']
+    log.save()
```
Hwesta: You can also create an object in one line/atomically with [`FixityLog.objects.create()`](https://docs.djangoproject.com/en/dev/ref/models/querysets/#create)

storage_service/locations/signals.py
```diff
@@ -44,4 +48,23 @@ def report_failed_fixity_check(sender, **kwargs):
...
+
+@receiver(successful_fixity_check, dispatch_uid="fixity_check")
+def report_successful_fixity_check(sender, **kwargs):
+    package = models.Package.objects.get(uuid=kwargs['uuid'])
+
+    log = models.FixityLog()
+    log.package = package
+    report_data = json.loads(kwargs['report'])
+    log.success = True
+    log.save()
```
Hwesta: Could this be merged with `report_failed_fixity_check` above? They seem very similar.  Also, report_data is not used in this function.

storage_service/locations/views.py
```diff
@@ -44,8 +44,23 @@ def get_delete_context_dict(request, model, object_uuid, default_cancel='/'):
...
+    # Amalgate last fixity check date and result with package data
+    package_data = []
+    for package in packages:
+        fixity_checks = FixityLog.objects.filter(package=package).order_by('-datetime_reported')
+        latest_check = fixity_checks[0].datetime_reported if fixity_checks else None
+        latest_result = fixity_checks[0].success if fixity_checks else None
+        package_data.append((latest_check, latest_result, package))
+
+    context = {'package_data': package_data}
     return render(request, 'locations/package_list.html', locals())
```
Hwesta: You probably meant `return render(request, 'locations/package_list.html', {'package_data': package_data, 'package': package)`

storage_service/templates/locations/fixity_results.html
```diff
@@ -0,0 +1,28 @@
...
+    {% for entry in log_entries %}
+      <tr>
+        <td>{{ entry.datetime_reported }}</td>
+        <td>{{ entry.error_details }}</td>
+      </tr>
+    {% endfor %}
+    </tbody>
+  </table>
+{% else %}
+  <p>No fixity failures exist.</p>
```
Hwesta: This isn't just for errors anymore, is it?  Perhaps display the `entry.success` value in the table too.

storage_service/locations/views.py
```diff
@@ -44,8 +44,23 @@ def get_delete_context_dict(request, model, object_uuid, default_cancel='/'):
...
 def package_list(request):
     packages = Package.objects.all()
+
+    # Amalgate last fixity check date and result with package data
+    package_data = []
+    for package in packages:
+        fixity_checks = FixityLog.objects.filter(package=package).order_by('-datetime_reported')
+        latest_check = fixity_checks[0].datetime_reported if fixity_checks else None
+        latest_result = fixity_checks[0].success if fixity_checks else None
+        package_data.append((latest_check, latest_result, package))
```
Hwesta: Since we need this info in several places (the package list, the Locations list where it shows several packages) then we probably want to generate this information in once place.  I think it makes sense to add latest_check & latest_result to the Package as @propeties.  Eg.

``` python
class Package(models.Model):
    ...
    @property
    def latest_fixity_check(self):
        try:
            latest_check = FixityLog.objects.filter(package=self).order_by('-datetime_reported')[0]  # limit 1
        except IndexError:
            return None
        return latest_check.datetime_reported
```

Thoughts?

Hwesta:
I noticed several try/except blocks disappeared in the rebase - why?

Also, tests?

storage_service/locations/signals.py
```diff
@@ -35,8 +39,16 @@ def report_deletion_request(sender, **kwargs):
     _notify_administrators(subject, message)
 
 
+def _log_report(uuid, success, message=None):
+    package = models.Package.objects.get(uuid=uuid)
+    models.FixityLog.objects.create(package=package, error_details=message)
```
Hwesta: Shouldn't `success` be used here?

This can also be done with `models.FixityLog.objects.create(package_id=uuid, error_details=message)`.  Is it worth having a 1-line cleanup function?

storage_service/templates/snippets/packages_table.html
```diff
@@ -9,11 +9,13 @@
...
         <th>Pointer File</th>
         <th>Status</th>
+        <th>Fixity Date</th>
+        <th>Fixity Status</th>
         <th>Actions</th>
       </tr>
     </thead>
     <tbody>
-    {% for package in packages %}
+    {% for last_fixity_check_datetime, last_fixity_check_result, package in package_data %}
```
Hwesta: `last_fixity_check_datetime, last_fixity_check_result` are never provided as context anywhere - is this leftover from previous changes?

storage_service/locations/views.py
Hwesta: I see you added the fixity attributes to the `Package` class, this view should be updated to use them instead of re-generating the information.  The template also needs to be updated for the new attributes.  `render` should return the `context` that it uses, not `locals()`.

storage_service/locations/models/package.py
```diff
@@ -163,6 +163,28 @@ def is_compressed(self):
...
+    def latest_fixity_check_result(self):
+        latest_check = self._latest_fixity_check()
+        if latest_check is not None:
+            return latest_check.success
+        else:
+            return None
+
+    def _latest_fixity_check(self):
+        try:
+            return FixityLog.objects.filter(package=self).order_by('-datetime_reported')[0]  # limit 1
```
Hwesta, line 179: Please run a linter over your code before you commit it. It will find things like `FixityLog` being undefined.

storage_service/locations/views.py
```diff
@@ -43,8 +43,13 @@ def get_delete_context_dict(request, model, object_uuid, default_cancel='/'):
 ########################## FILES ##########################
 
 def package_list(request):
-    packages = Package.objects.all()
-    return render(request, 'locations/package_list.html', locals())
+    context = {'packages': Package.objects.all()}
+    return render(request, 'locations/package_list.html', context)
+
+def package_fixity(request, package_uuid):
+    log_entries = FixityLog.objects.filter(package__uuid=package_uuid).order_by('-datetime_reported')
```
Hwesta, line 50: Tip: You can filter by the value in a foreign key (instead of the object, which requires a join) by using `_id` https://docs.djangoproject.com/en/1.8/topics/db/queries/#field-lookups  So this could be `FixityLog.objects.filter(package_id=package_uuid).order_by('-datetime_reported')`

storage_service/locations/tests/test_fixity_log.py
```diff
@@ -0,0 +1,18 @@
+from django.test import TestCase
+
+from locations import models
+
+
+class TestFixityLog(TestCase):
+
+    fixtures = ['base.json', 'fixity_log.json']
```
Hwesta, line 8: This fixture fails to load with `IntegrityError: Problem installing fixtures: The row in table 'locations_fixitylog' with primary key '1' has an invalid foreign key: locations_fixitylog.package_id contains a value 'e0a41934-c1d7-45ba-9a95-a7531c063ed1' that does not have a corresponding value in locations_package.uuid.`  You need to add the 'package.json' to the fixtures list.

storage_service/locations/tests/test_fixity_log.py
```diff
@@ -0,0 +1,18 @@
...
+
+    def setUp(self):
+        self.fl_object = models.FixityLog.objects.all()[0]
+        #self.auth = requests.auth.HTTPBasicAuth(self.ds_object.user, self.ds_object.password)
```
Hwesta, line 12: This looks leftover?

Hwesta:
Glad to see tests! Maybe add a couple for the Package's new attributes (since they're actually functions) `latest_fixity_check_datetime` `latest_fixity_check_result`, especially if there's no FixityLog objects.

Also, please rebase this on qa/0.x

mcantelon:
I got rid of some of the try/except blocks because I couldn't remember what had caused them and hence couldn't be specific about what exception to check for.

sevein:
@mcantelon, it looks like it was a mistake to add this to the `0.8.0` milestone. Sorry for the trouble.

mcantelon:
Merged in qa/0.x.

PR #85, opened 2015-10-14
Logging: Cleanup config

Hwesta:
Remove duplicate handler config, setup a catch-all root logger.

### DIFF
--- a/storage_service/storage_service/settings/base.py
+++ b/storage_service/storage_service/settings/base.py
@@ -283,36 +283,27 @@ LOGGING = {
         },
     },
     'loggers': {
-        'django': {
-            'handlers': ['null'],
-            'propagate': True,
-            'level': 'INFO',
-        },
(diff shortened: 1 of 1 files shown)


sevein:
:+1:

mistydemeo:
:+1: 

Does this change the level at which `django` loggers occur?

Hwesta:
> Does this change the level at which django loggers occur?

Since 'django' doesn't have an explicit logger anymore, it will be caught by the 'root' logger, and everything `WARNING` and higher will be logged.  This is the same behavior as for any other library we don't explicitly handle.

Previously, it was somewhat confused.  Anything from 'django' was silenced (null handler), then propagated to the root logger, which probably went to apache/error.log. ERROR logs from 'django.request' or 'django.request.tastypie' had some handling.

This is similar to the AM logging config in artefactual/archivematica#303

PR #86, opened 2015-10-16
Allow new metadata files to be POSTed to storage service (re-ingest), refs #9059

sevein:
Redmine ticket: https://projects.artefactual.com/issues/9059

This pull request updates the reingest choices to:
- Full re-ingest aka `full` (**NEW**).
- Metadata re-ingest aka `metadata`.
- Partial re-ingest `partial`.

If the re-ingest is `full`, the user has the option to indicate a processing configuration different that the default.

---

A new parameter `target` is sent to the [pipeline's reingest API](https://github.com/artefactual/archivematica/blob/qa/1.x/src/dashboard/src/components/api/views.py#L395-L436) indicating whether the packages should go to `ingest` or `transfer`.

| Re-ingest | `/api/reingest`'s field `target` |
| --- | --- |
| full | transfer |
| metadata | ingest |
| partial | ingest |

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -27,6 +27,7 @@ from . import StorageException
 from location import Location
 from space import Space
 from event import File
+from pipeline import PipelineClientException
 
 __all__ = ('Package', )
 
@@ -113,9 +114,10 @@ class Package(models.Model):
     )
 
     # Reingest type options
-    METADATA_ONLY = 'metadata'
-    OBJECTS = 'objects'
-    REINGEST_CHOICES = (METADATA_ONLY, OBJECTS)
+    METADATA_ONLY = 'metadata' # Re-ingest metadata only
+    OBJECTS = 'objects'        # Re-ingest metadata and objects for DIP generation
+    FULL = 'full'              # Full re-ingest
+    REINGEST_CHOICES = (METADATA_ONLY, OBJECTS, FULL)
 
     class Meta:
         verbose_name = "Package"
@@ -372,7 +374,7 @@ class Package(models.Model):
         temp_aip.delete()
 
         # Do fixity check of AIP with recovered files
-        return self.check_fixity() 
+        return self.check_fixity()
 
     def store_aip(self, origin_location, origin_path):
         """ Stores an AIP in the correct Location.
@@ -835,13 +837,14 @@ class Package(models.Model):
 
     # REINGEST
 
-    def start_reingest(self, pipeline, reingest_type):
+    def start_reingest(self, pipeline, reingest_type, processing_config='default'):
         """
         Copies this package to `pipeline` for reingest.
 
         Fetches the AIP from storage, extracts and runs fixity on it to verify integrity.
         If reingest_type is METADATA_ONLY, sends the METS and all files in the metadata directory.
         If reingest_type is OBJECTS, sends METS, all files in metadata directory and all objects, preservation and original.
+        If reingest_type is FULL, we do like in OBJECTS but sending the package to the transfer source location.
         Calls Archivematica endpoint /api/ingest/reingest/ to start reingest.
 
         :param pipeline: Pipeline object to send reingested AIP to.
@@ -855,7 +858,7 @@ class Package(models.Model):
         # Check and set reingest pipeline
         if self.misc_attributes.get('reingest_pipeline', None):
             return {'error': True, 'status_code': 409,
-                'message': 'This AIP already being reingested on {}'.format(self.misc_attributes['reingest_pipeline'])}
+                'message': 'This AIP is already being reingested on {}'.format(self.misc_attributes['reingest_pipeline'])}
         self.misc_attributes.update({'reingest_pipeline': pipeline.uuid})
 
         # Fetch and extract if needed
@@ -880,7 +883,7 @@ class Package(models.Model):
         reingest_files = [
             os.path.join(relative_path, 'data', 'METS.' + self.uuid + '.xml')
         ]
-        if reingest_type == self.OBJECTS:
+        if reingest_type in (self.OBJECTS, self.FULL):
             # All in objects except submissionDocumentation dir
             for f in os.listdir(os.path.join(local_path, 'data', 'objects')):
                 if f in ('submissionDocumentation'):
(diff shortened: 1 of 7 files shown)


storage_service/locations/urls.py
```diff
@@ -25,7 +25,7 @@
         name='package_update_status'),
     url(r'^packages/aip_recover_request$', views.aip_recover_request,
         name='aip_recover_request'),
-    url(r'^packages/(?P<package_uuid>' + UUID + ')/reingest/$', 'aip_reingest',
+    url(r'^packages/(?P<package_uuid>' + UUID + ')/reingest/$', views.aip_reingest,
```
mistydemeo, line 28: Rebase on top of qa/0.x - there's a fix for this in now.

storage_service/locations/models/package.py
```diff
@@ -894,6 +896,25 @@ def start_reingest(self, pipeline, reingest_type):
         elif reingest_type == self.METADATA_ONLY:
             reingest_files.append(os.path.join(relative_path, 'data', 'objects', 'metadata', ''))
 
+        # Fetch processing configuration, put it in the root of the package and
+        # include the file in reingest_files.
+        if reingest_type == self.FULL and processing_config != 'default':
+            config = pipeline.get_processing_config(processing_config)
+            if config == '':
+                LOGGER.debug('Reingest: processing configuration %s could not be loaded', processing_config)
```
mistydemeo: Should this be an error condition?

storage_service/locations/models/package.py
```diff
@@ -921,31 +942,19 @@ def start_reingest(self, pipeline, reingest_type):
...
+        reingest_target = 'transfer' if self.FULL else 'ingest'
         try:
-            response = requests.post(reingest_url, data=params, allow_redirects=True)
-        except requests.exceptions.RequestException:
-            LOGGER.exception('Unable to connect to pipeline %s', pipeline)
-            return {'error': True, 'status_code': 502,
-                'message': 'Unable to connect to pipeline'}
-        LOGGER.debug('Response: %s %s', response.status_code, response.text)
+            response = pipeline.reingest(relative_path, self.uuid, reingest_target)
+        except Exception:
```
mistydemeo: Maybe catch specific exceptions instead of everything?

storage_service/locations/models/package.py
```diff
@@ -880,7 +882,7 @@ def start_reingest(self, pipeline, reingest_type):
         reingest_files = [
             os.path.join(relative_path, 'data', 'METS.' + self.uuid + '.xml')
         ]
-        if reingest_type == self.OBJECTS:
+        if reingest_type == self.OBJECTS or reingest_type == self.FULL:
```
Hwesta: Tip: Can be `if reingest_type in (self.OBJECTS, self.FULL):`

storage_service/locations/models/pipeline.py
```diff
@@ -126,3 +127,53 @@ def create_default_locations(self, shared_path=None):
...
+            raise e
+        LOGGER.debug('Response: %s %s', response.status_code, response.text)
+        return response
+
+    def get_processing_config(self, name):
+        """Obtain a processing configuration XML document given its name. The
+        content is returned as a string, which may be empty in case of errors.
+        """
+        if not name:
+            raise ValueError()
```
mistydemeo: Include a message here?

storage_service/locations/models/pipeline.py
```diff
@@ -126,3 +127,53 @@ def create_default_locations(self, shared_path=None):
...
+        url = 'processing-configuration/' + name
+        response = None
+        try:
+            response = self._request_api('GET', url)
+        except requests.exceptions.RequestException:
+            return ''
```
mistydemeo: Raise an exception instead of returning an empty string, maybe?

storage_service/locations/models/package.py
```diff
@@ -894,6 +896,25 @@ def start_reingest(self, pipeline, reingest_type):
...
+                try:
+                    # It's not expected to find an existing processingMCP.xml
+                    # file in the original AIP, but we are using the w+ mode
+                    # just in case.
+                    with open(config_path, 'w+') as f:
+                        f.write(config)
+                    LOGGER.debug('Reingest: processing configuration %s written, location: %s', processing_config, config_path)
+                    reingest_files.append(os.path.join(relative_path, 'processingMCP.xml'))
+                except IOError as e:
+                    LOGGER.error('Reingest: processing configuration %s could not be written, error: %s', processing_config, e)
```
Hwesta: Tip: `LOGGER.exception` or `LOGGER.error( ... , exc_info=True)` will attach the traceback to the log, instead of just printing the exception-string.

storage_service/locations/models/pipeline.py
```diff
@@ -126,3 +127,53 @@ def create_default_locations(self, shared_path=None):
...
+
+    def reingest(self, name, uuid, target='transfer'):
+        if not name or not uuid:
+            raise ValueError()
+        fields = {
+            'name': name,
+            'uuid': uuid,
+            'target': target
+        }
+        return self._request_api('POST', 'reingest', fields=fields)
```
mistydemeo: Newline at end of file

storage_service/locations/models/pipeline.py
```diff
@@ -126,3 +127,53 @@ def create_default_locations(self, shared_path=None):
...
+        }
+        if 'fields' in kwargs:
+            fields.update(kwargs['fields'])
+
+        LOGGER.debug('URL: %s; fields: %s', url, fields)
+        try:
+            response = requests.request(method, url, data=fields, allow_redirects=True)
+        except requests.exceptions.RequestException as e:
+            LOGGER.exception('Unable to connect to pipeline %s', self)
+            raise e
```
Hwesta: A bare `raise` in an exception will re-raise it, which looks like what you're trying to do.  Details: http://www.ianbicking.org/blog/2007/09/re-raising-exceptions.html

mistydemeo:
Regarding the "full" reingest type name - is this a "full" reingest because we're reingesting files other than metadata and obejcts, or because it's going through transfer?

storage_service/locations/models/pipeline.py
```diff
@@ -126,3 +127,53 @@ def create_default_locations(self, shared_path=None):
...
+        if not name:
+            raise ValueError()
+        url = 'processing-configuration/' + name
+        response = None
+        try:
+            response = self._request_api('GET', url)
+        except requests.exceptions.RequestException:
+            return ''
+        if response.status_code != requests.codes.ok:
+            LOGGER.error('Error in processing configuration API (status_code=%d)'.format(response.status_code))
```
Hwesta: Logging does its own % formatting, no `.format` needed

sromkey:
@mistydemeo : "Full" referring to the fact that it's going back to transfer for a complete re-ingest.

storage_service/locations/models/pipeline.py
sevein: Thanks, I keep forgetting that.

mistydemeo:
@sromkey I wonder if a different name is appropriate then? The other two refer to the kinds of files being reingested, so "full" implies that even more files than in "objects" are being reingested - at least to me.

storage_service/locations/models/pipeline.py
sevein: Ok, thanks.

sromkey:
@mistydemeo Any suggestions? How about "complete"?

PR #87, opened 2015-10-19
Tests: Update Package fixtures & tests

Hwesta:
- Pull package-related fixtures into their own package.json fixture file, reorder
- Add tests for `check_fixity`, `download_request` and `extract_file_request`
- Require `extract_file` to have a GET param of relative_path_to_file, handle error


Hwesta:
Django also supports YAML as a fixture format (as does loaddata/dumpdata that these are generated from).  Do we want to switch to YAML because it's a less picky format and allows comments?

sevein:
YAML also adds many complexities around data types, which we probably don't need. Can we just add the supporting documentation in a README file?

Hwesta:
I've found YAML to be lightweight and more flexible than JSON, but I don't feel strongly about switching. README seems overkill for the lightweight context-dependent comments I was thinking of, but could be useful, especially as we get more tests/fixtures

mistydemeo:
I could see comments as being a good argument to switch to YAML, as long as we're confident we're unlikely to introduce type issues.

storage_service/locations/tests/test_api.py
```diff
@@ -61,6 +61,25 @@ def test_cant_move_to_disabled_locations(self):
...
+        # Set up locations to point to fixtures directory
+        self.test_location.relative_path = self.fixtures_dir[1:]
+        self.test_location.save()
+        ss_int = models.Location.objects.get(purpose='SS')
+        ss_int.relative_path = self.fixtures_dir[1:]
+        ss_int.save()
+
+    def tearDown(self):
+        for entry in os.listdir(self.fixtures_dir):
+            if entry.startswith('tmp'):
```
mistydemeo, line 80: Would it be possible for us to stub the stuff that actually creates files so we inhibit them being placed on-disk in the first place?

Hwesta, line 80: Probably, but I haven't learned how yet. :) How important do you think this is, vs creating/deleting actual files?

mistydemeo, line 80: It probably doesn't need to hold this up in the short-term!

PR #88, opened 2015-10-23
Add Arkivum-specific return codes

Hwesta:
Arkivum files are NFS-mounted and have local paths but may not be immediately available locally.  Before fetching an Arkivum file (including checking if it's compressed) check if the file is locally available and return different status codes depending on Arkivum's response.

Endpoints: `/api/v1/file/<uuid>/download/` & `/api/v1/file/<uuid>/extract_file/`
- 200 package locally available (return file)
- 202 package exists but not here now (Arkivum needs to get from tape)
- 404 package doesn't exist
- 502 backend error from Arkivum (assume can't get file)

Includes #87

### DIFF
--- a/storage_service/common/utils.py
+++ b/storage_service/common/utils.py
@@ -98,20 +98,13 @@ def download_file_stream(filepath, temp_dir=None):
         return http.HttpResponseNotFound("File not found")
 
     filename = os.path.basename(filepath)
-    extension = os.path.splitext(filepath)[1].lower()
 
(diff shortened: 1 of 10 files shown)


mistydemeo:
:+1: 

Does anything that currently uses those endpoints, e.g. in Archivematica, need to be updated due to this?

Hwesta:
Replaced by #136, which also includes #95 and cleanups.

PR #89, opened 2015-10-30
Dev/issue 8895 search api

mcantelon:
Added new fields to File model and add logic to index files in stored AIPs using metsrw to parse file metadata from METS files. It's related to issue #8895.


storage_service/metsrw/exceptions.py
```diff
@@ -0,0 +1,13 @@
+"""
```
Hwesta: Why not specify this as a pip dependency instead of vendoring it? Pip can install from git, and we can start working on getting it into PyPI. See https://github.com/artefactual/automation-tools/blob/dev/dataverse/requirements/base.txt#L4

storage_service/metsrw/__init__.py
```diff
@@ -0,0 +1,13 @@
+from __future__ import absolute_import
```
Hwesta: Why not specify this as a pip dependency instead of vendoring it? Pip can install from git, and we can start working on getting it into PyPI. See https://github.com/artefactual/automation-tools/blob/dev/dataverse/requirements/base.txt#L4

docs/search.rst
```diff
@@ -0,0 +1,170 @@
+:Authors:
```
Hwesta, line 1: :sparkling_heart: Docs

storage_service/locations/migrations/0005_auto_20151026_1638.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
```
Hwesta: Rename this to something semantically useful like 0005_search_api.py

docs/search.rst
```diff
@@ -0,0 +1,170 @@
...
+  http://<storage service URL>/api/v2/search/location/
+
+Locations can be searched using the following search parameters:
+
+* uuid (location UUID)
+* space (space UUID)
+* purpose (purpose code)
+* enabled (whether the location is enabled)
+
+Adding "?format=json" when sending a GET request will return JSON results.
```
Hwesta: While this is true in TastyPie, it's more of a bug than a feature. Not requiring `?format=json` would be better, since we're making improvements.

storage_service/locations/api/resources.py
```diff
@@ -475,6 +476,10 @@ def obj_create(self, bundle, **kwargs):
         if bundle.obj.package_type in (Package.AIP, Package.AIC, Package.DIP) and bundle.obj.current_location.purpose in (Location.AIP_STORAGE, Location.DIP_STORAGE):
             # Store AIP/AIC
             bundle.obj.store_aip(origin_location, origin_path)
+
+            # Asynchronously index AIP files
+            p = Process(target=bundle.obj.index_file_data_from_aip_mets)
+            p.start()
```
Hwesta: How do we check if this succeeded or failed?

This is probably a good thing to discuss with Justin/Misty/etc in the context generally being able to do asynchronous processing in the storage service.

storage_service/locations/models/package.py
```diff
@@ -752,6 +754,104 @@ def backlog_transfer(self, origin_location, origin_path):
...
+
+        import metsrw
+
+        aip_dir_name = os.path.basename(os.path.splitext(self.full_path)[0])
+        temp_dir = tempfile.mkdtemp()
+        relative_path = os.path.join(aip_dir_name, "data", "METS." + self.uuid + ".xml")
+
+        self.extract_file(relative_path, temp_dir)
+
+        mw = metsrw.mets.METSWriter()
```
Hwesta: Note: This will need to be updated if/when artefactual-labs/mets-reader-writer#9 gets merged.

storage_service/locations/models/package.py
```diff
@@ -752,6 +754,104 @@ def backlog_transfer(self, origin_location, origin_path):
...
+        Attempts to read an Archivematica AIP METS file inside this
+        package, then uses the retrieved metadata to generate one entry in the
+        File table in the database for each file inside the package.
+
+        :raises StorageException: if the transfer METS cannot be found,
+            or if required elements are missing.
+        """
+        prefix = self.full_path
+
+        import metsrw
```
Hwesta: Imports should be [at the top of the file](https://www.python.org/dev/peps/pep-0008/#imports).

storage_service/locations/models/package.py
```diff
@@ -752,6 +754,104 @@ def backlog_transfer(self, origin_location, origin_path):
...
+
+        aip_dir_name = os.path.basename(os.path.splitext(self.full_path)[0])
+        temp_dir = tempfile.mkdtemp()
```
Hwesta: Providing a temp dir is unecessary, since `extract_file` creates a temp dir if none is provided, and returns it.  See https://github.com/artefactual/archivematica-storage-service/blob/stable/0.7.x/storage_service/locations/models/package.py#L451-L464

Also, don't forget to delete the temp dir once it's no longer needed.

storage_service/locations/models/package.py
```diff
@@ -752,6 +754,104 @@ def backlog_transfer(self, origin_location, origin_path):
...
+        Cycle through an FSEntry object's AMDsec subsections and consolidate
+        PREMIS object/event metadata.
+        """
+        metadata = None
+
+        if fsentry.path != 'None':
+            metadata = {}
+
+            for subsection in fsentry.amdsecs[0].subsections:
+                if subsection.subsection == 'techMD' and subsection.contents.mdtype == 'PREMIS:OBJECT':
```
Hwesta: Maybe update metsrw to provide easy access to techMD (and other subsection types) instead of having to iterate over them like this?  metsrw really needs an API doc...

storage_service/locations/models/event.py
```diff
@@ -142,6 +155,11 @@ class File(models.Model):
         help_text="Accession ID of originating transfer")
     origin = UUIDField(editable=False, unique=False, version=4, blank=True,
         help_text="Unique identifier of originating Archivematica dashboard")
+    normalized = models.BooleanField(default=None,
+        help_text="Whether or not file has been normalized")
+    validated = models.BooleanField(default=None,
```
Hwesta: What's the difference between `False` and `None` for `normalized` and `validated`? Should they default to `False`?

storage_service/locations/api/resources.py
mcantelon: Hrm, yeah good question.

mistydemeo: Yeah, this is something we've meant to look into but we don't necessarily have a good answer on right now. It'd be good to be able to track success on this though.

storage_service/locations/api/search/router.py
```diff
@@ -0,0 +1,159 @@
...
+    This allows users to query booleans without having to use "True" and "False"
+    """
+    def filter(self, qs, value):
+        if value is not None:
+            lc_value = value.lower()
+            if lc_value == "true":
+                value = True
+            elif lc_value == "false":
+                value = False
+            return qs.filter(**{self.name: value})
```
mistydemeo, line 22: What happens here if the string doesn't match either? It just passes the string back?

storage_service/locations/models/package.py
```diff
@@ -752,6 +754,104 @@ def backlog_transfer(self, origin_location, origin_path):
...
+
+        self.extract_file(relative_path, temp_dir)
+
+        mw = metsrw.mets.METSWriter()
+        mw.fromfile(os.path.join(temp_dir, relative_path))
+
+        for fsentry in mw.all_files():
+
+
+            if fsentry is not None and fsentry.path != 'None':
```
mistydemeo: Do you need to guard `fsenrty.path` here since that's done in `_parse_file_metadata`?

storage_service/locations/models/package.py
```diff
@@ -752,6 +754,104 @@ def backlog_transfer(self, origin_location, origin_path):
...
+        if fsentry.path != 'None':
+            metadata = {}
+
+            for subsection in fsentry.amdsecs[0].subsections:
+                if subsection.subsection == 'techMD' and subsection.contents.mdtype == 'PREMIS:OBJECT':
+                    premis_object = metsrw.Object.parse(subsection.contents.document)
+
+                    # Don't provide metadata for METS files
+                    if premis_object.characteristics.is_mets:
+                        return None
```
mistydemeo: Nitpick, but bare `return` returns `None`.

storage_service/locations/models/package.py
```diff
@@ -752,6 +754,104 @@ def backlog_transfer(self, origin_location, origin_path):
...
+                    aip_file.name = os.path.join(aip_dir_name, fsentry.path)
+                    aip_file.ingestion_time = mw.createdate
+                    if 'format_name' in metadata:
+                        aip_file.format_name = metadata['format_name']
+                    if 'size' in metadata:
+                        aip_file.size = int(metadata['size'])
+                    if 'pronom_id' in metadata:
+                        aip_file.pronom_id = metadata['pronom_id']
+                    if 'normalized' in metadata:
+                        aip_file.normalized=metadata['normalized']
```
mistydemeo: Nitpick, but spaces around `=` when it's not a keyword argument.

storage_service/locations/api/resources.py
mistydemeo: Celery looks interesting: http://docs.celeryproject.org/en/latest/django/first-steps-with-django.html

It has Django integration, including the ability to store task results in the database using the Django ORM.

storage_service/locations/api/search/router.py
mistydemeo, line 22: Just tested, looks like it's ignored if a supported value isn't passed. Maybe we want this it to raise an error instead?

mistydemeo:
One thing I'm wondering about - I'm not sure what the requirements were here. Is there any requirement to support clauses other than `AND`? E.g. to be able to specify `NOT` or `OR` clauses?

mistydemeo:
Ping - any thoughts re: my last comment?

storage_service/locations/models/package.py
mcantelon: Makes sense... I'll make it it so there's an array for each subsection type.

storage_service/locations/models/event.py
mcantelon: If the value's None then a validation check didn't occur... if it's False then it failed.

mcantelon:
@mistydemeo Sorry, didn't see this... The requirements are pretty rudimentary... here are the searches that were anticipated: https://wiki.archivematica.org/Research_data_management#METS_parsing

storage_service/metsrw/__init__.py
mcantelon: Thanks... that works slick.

storage_service/locations/api/resources.py
mcantelon: Celery sounds like a good solution.

storage_service/locations/api/search/router.py
mcantelon, line 22: Yeah, it passes the string back as it and their search query will then fail (which makes sense given they're provided the wrong values for a boolean).

mistydemeo, line 22: Okay, cool! What's the failure look like?

mcantelon, line 22: It'll pass the value back as-is and cause the query to be invalid (which is desirable given they're provided the wrong type of value).

jhsimpson:
I moved the milestone for this to 0.10.0, instead of 0.9.0, because this feature brings in changes to some dependencies, (django rest framework, mets-reader-writer) and we need a bit more time to sort those out.

storage_service/locations/models/package.py
```diff
@@ -752,6 +752,106 @@ def backlog_transfer(self, origin_location, origin_path):
...
+        Cycle through an FSEntry object's AMDsec subsections and consolidate
+        PREMIS object/event metadata.
+        """
+        metadata = None
+
+        if fsentry.path != 'None':
+            metadata = {}
+
+            # Get technical metadata
+            if len(fsentry.techmds):
```
Hwesta, line 1643: Tip: This could be just `if fsentry.techmds` since an empty list is Falsey.

storage_service/locations/models/package.py
```diff
@@ -752,6 +752,106 @@ def backlog_transfer(self, origin_location, origin_path):
...
+                techmd = fsentry.techmds[0]
+                premis_object = metsrw.premis.Object.parse(techmd.contents.document, False)
```
Hwesta, line 1645: What's the `False`? Please name the arguments, especially if they're something without context like a boolean.

storage_service/locations/models/package.py
```diff
@@ -752,6 +752,106 @@ def backlog_transfer(self, origin_location, origin_path):
...
+
+                # Don't provide metadata for METS files
+                if premis_object.characteristics[0]['is_mets']:
+                    return
+
+                metadata['filename'] = premis_object.original_name
+
+                if len(premis_object.object_identifiers[0]):
+                    if premis_object.object_identifiers[0]['type'] == 'UUID':
```
Hwesta, line 1654: This sort of usage is useful to inform what changes we should make to metsrw & plugins - it should be easier to do this.

storage_service/locations/models/package.py
```diff
@@ -752,6 +752,106 @@ def backlog_transfer(self, origin_location, origin_path):
...
+        package, then uses the retrieved metadata to generate one entry in the
+        File table in the database for each file inside the package.
+
+        :raises StorageException: if the transfer METS cannot be found,
+            or if required elements are missing.
+        """
+        aip_dir_name = os.path.basename(os.path.splitext(self.full_path)[0])
+        relative_path = os.path.join(aip_dir_name, "data", "METS." + self.uuid + ".xml")
+
+        path_to_mets, temp_dir = self.extract_file(relative_path)
```
Hwesta, line 1603: Has this been tested with compressed & uncompressed AIPs?  Does it work with AIPs not stored in the local filesystem?  Since this is happening asynchronously, what happens when the AIP storage operation completes during this thread? eg self.full_path may not be accurate, or may change partway through

Hwesta (commented):
This looks pretty good, but there are a few architectural questions to consider before merging it.

As Misty asked, do the queries need to support OR or NOT, or anything more complicated than AND? How might that be handled?

How should we handle multiprocessing? This uses the multiprocessing, and using celery has been suggested. How do we want to handle async processes?  For the async function itself, how does it handle multiple threads dealing with the same data? Since it's happing in parallel with the store AIP procedure, care needs to be taken that they're not operating on the same data. Also, what happens if only one of those processes fails?

This is also dependent on artefactual-labs/mets-reader-writer#20 which has its own architecture questions to be discussed.

There's a lot of good practical uses of metsrw in here as well - this is very helpful for deciding what the metsrw API could look like.

docs/search.rst
```diff
@@ -0,0 +1,164 @@
...
+
+::
+
+  http://<storage service URL>/api/v2/search/file/
+
+Files can be searched using the following search criteria:
+
+* uuid (file UUID)
+* package (package UUID)
+* name (enter or partial filename)
```
jrwdunham, line 113: Is this a typo? Should "enter" be "entire"?

jrwdunham: Also, I think there's a typo in the line below: "PRONUM"

docs/search.rst
```diff
@@ -0,0 +1,164 @@
...
+* pronom_id (PRONUM PUID)
+* format_name (format name)
+* min_size (minimum filesize)
+* max_size (maximum filesize)
+* normalized (boolean: whether or not file was normalized)
+* valid (boolean: whether or not file data is valid or malformed)
```
jrwdunham, line 119: Wording sounds contradictory: "whether or not file data is valid or malformed"

docs/search.rst
```diff
@@ -0,0 +1,164 @@
...
+
+For example, if you wanted to get details about files contained in the package
+ebd8cbb6-29ca-4fa3-b897-58bea057f35f you could HTTP GET:
+
+::
+
+   http://<storage service URL>/api/v2/search/file/?min_size=29965171
+
+Here is an example JSON response:
```
jrwdunham: The text above the URL doesn't appear to match what that URL actually searches for.

storage_service/locations/api/search/router.py
```diff
@@ -0,0 +1,159 @@
...
+    max_size = django_filters.NumberFilter(name='size', lookup_type='lte')
+    pipeline = django_filters.CharFilter(name='origin')
+    package = django_filters.CharFilter(name='source_package')
+    name = django_filters.CharFilter(name='name', lookup_type='icontains')
+    normalized = CaseInsensitiveBooleanFilter(name='normalized')
+    ingestion_time = django_filters.DateFilter(name='ingestion_time', lookup_type='contains')
+
+    class Meta:
+        model = models.File
+        fields = ['uuid', 'name', 'file_type', 'min_size', 'max_size', 'format_name', 'pronom_id', 'pipeline', 'source_package', 'normalized', 'validated', 'ingestion_time']
```
jrwdunham: Should `source_package` in this list be `package`?

docs/search.rst
jrwdunham: Also, I think `valid` should be `validated` here. No? If so, I think the validated attribute of File only records whether the file has been validated, not whether it is valid. (Unless the help text is wrong in `models/event.py::File`

storage_service/locations/api/search/router.py
```diff
@@ -0,0 +1,159 @@
...
+    Search API view for File model data
+
+    Custom endpoint "stats" provides total size of files searched for
+    """
+    queryset = models.File.objects.all()
+    serializer_class = FileSerializer
+    filter_backends = (filters.DjangoFilterBackend,)
+    filter_class = FileFilter
+
+    @list_route(methods=['get'])
```
jrwdunham, line 152: How is this used? TODO: find out.

storage_service/locations/migrations/0016_search_api.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('locations', '0015_gpg_encrypted_space'),
```
jrwdunham: Needs another rebase: 0016_mirror_location_aip_replication.py migration has since been introduced.

storage_service/locations/models/package.py
jrwdunham: Now that files in a METS file can have other types of ids (e.g., handle, PURL), we need a more robust way of getting the UUID. I think this needs to loop through all identifiers and find the first UUID one, if it exists.

storage_service/locations/models/package.py
```diff
@@ -820,6 +820,106 @@ def backlog_transfer(self, origin_location, origin_path):
...
+
+            # Cycle through event data to see if file has been validated and if it passed
+            for digiprovmd in fsentry.digiprovmds:
+                if digiprovmd.contents.mdtype == 'PREMIS:EVENT':
+                    # Parse PREMIS event
+                    premis_event = metsrw.premis.Event.parse(digiprovmd.contents.document)
+
+                    # Indicate whether or not a file has been validated in metadata and if it passed
+                    if premis_event.event_type == 'validation':
+                        metadata['validated'] = premis_event.outcomes[0]['outcome'] == "pass"
```
jrwdunham, line 1686: So it seems that the `validated` attribute of a file does actually mean "is valid". I think maybe the name should be change to `is_valid`. But if that's the case, then I think we want 3 possible values: `True`, `False`, and `None`...

docs/search.rst
```diff
@@ -0,0 +1,146 @@
...
+
+* uuid (location UUID)
+* space (space UUID)
+* purpose (purpose code)
+* enabled (whether the location is enabled)
+
+For example, if you wanted to get details about the transfer source location
+contained in the space 6d0b6cce-4372-4ef8-bf48-ce642761fd41 you could HTTP get::
+
+    http://<storage service URL>/api/v2/search/location/?space=7ec3d5d9-23ec-4fd5-b9fb-df82da8de630&purpose=TS
```
jrwdunham, line 32: I noticed that if you pass a nonsense param to one of these search endpoints, you get all of the resources. Is that expected/desired? For example `http://192.168.168.192:8000/api/v2/search/location/?moonbeam=figaro` returns all locations. As does `http://192.168.168.192:8000/api/v2/search/location/`

docs/search.rst
jrwdunham: Should `source_package` be listed as a search filter here? `FileFilter` seems to expose it. It's a bit confusing since it only references Transfers I believe. See the pre-existing `Package.index_file_data_from_transfer_mets`

docs/search.rst
jrwdunham: I fixed these typos in https://github.com/artefactual/archivematica-storage-service/pull/262

docs/search.rst
jrwdunham: I changed `validated` to `valid` a nullable boolean field where `None` indicates not validated, `True` valid, and `False` invalid. Migrations and help text updated accordingly in https://github.com/artefactual/archivematica-storage-service/pull/262.

jrwdunham (commented):
My comments have led to changes in the PR that supersedes this one: see https://github.com/artefactual/archivematica-storage-service/pull/262

jrwdunham:
I am closing this in favour of https://github.com/artefactual/archivematica-storage-service/pull/262

PR #90, opened 2015-10-30
WIP - Backlog transfer package deletion for SS

MikeFE:
- Generalize the storage service deletion request page so it can delete both AIPs and backlog transfer packages


Hwesta:
:+1:

PR #91, opened 2015-11-12
Package: prefix METS pointer elements

mistydemeo:


mistydemeo:
It's possibly an oversight that these weren't assigned a namespace when adding these elements before; because pointer files had no default namespace, it was possible to omit namespaces and still have them considered METS elements.

Hwesta:
`common/utils.py` also touches pointer files and namespaces.

mistydemeo:
Good catch, updated.

Hwesta:
And it looks like `lockssomatic.py` also modifies the pointer file.  No other files use `[Sub]Element` or `ElementMaker`

mistydemeo:
Arg, also good catch. Updated, and also did some cleanup since this file predates `utils.PREFIX_NS`.

Hwesta:
:+1:

PR #92, opened 2015-11-13
Fix "Re-ingest" button

mistydemeo:


Hwesta:
:+1: 

I assume you checked for other merge-mistakes like this?

mistydemeo:
At a glance, everything else in this file seems fine.

PR #94, opened 2015-11-18
Test improvements

mistydemeo:
- List test deps in requirements/test.txt
- Set null loggers for the root logger, instead of using the default loggers, so that the tests can be run without having a `/var/archivematica/sharedDirectory`
- Update python-keystoneclient, which allows Swift tests to pass


storage_service/storage_service/settings/test.py
```diff
@@ -13,3 +13,23 @@
...
+LOGGING = {
+    'version': 1,
+    'disable_existing_loggers': False,
+    'filters': {
+        'require_debug_false': {
+            '()': 'django.utils.log.RequireDebugFalse'
+        }
+    },
+    'handlers': {
+        'null': {
```
Hwesta: Have a console logger instead of a null logger so we can still see debugging output?  Does py.test still print logging output if the test fails with it configured like this?

mistydemeo: Console logger might be a good idea, yeah. I was assuming that tests were likely to raise, not silently fail for other reasons, but that might not be a fair assumption.

mistydemeo:
Added a commit which fixes working-directory sensitive things, like VCR cassette locations. This makes it possible to run the tests from the repository root.

storage_service/storage_service/settings/test.py
Hwesta: And it's useful for troubleshooting why the test failed.

mistydemeo: Switched to using `StreamHandler` at the default of stderr.

storage_service/storage_service/settings/test.py
```diff
@@ -13,3 +13,29 @@
         "PORT": "",
     },
 }
+
+LOGGING = {
```
Hwesta, line 17: Update `pytest.ini` to point at this not settings.local?

mistydemeo, line 17: Done!

Hwesta:
As discussed, this reveals a problem with `test_startup.py` where the tested function is imported before setup finished, and is run on import.  Pushed a suggested fix - what do you think?

Hwesta:
:+1:

PR #96, opened 2015-12-01
Add gunicorn to requirements/local.txt

sevein:
It is going to be used in development environments (`gunicorn --reload`), it
may be added later to other environments.


Hwesta:
:+1:

PR #98, opened 2015-12-03
Reingesting uncompressed AIPs fix. refs #8893

mcantelon:
Fixed issues with reingestion of uncompressed AIPs.

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -498,9 +498,8 @@ class Package(models.Model):
             rc = subprocess.call(command)
             LOGGER.debug('Extract file RC: %s', rc)
         else:
-            aip_path = os.path.join(full_path, basename)
-            LOGGER.info('Copying AIP from: %s to %s', aip_path, output_path)
-            shutil.copytree(aip_path, output_path)
+            LOGGER.info('Copying AIP from: %s to %s', full_path, output_path)
+            shutil.copytree(full_path, output_path)
 
         if not relative_path:
             self.local_path_location = ss_internal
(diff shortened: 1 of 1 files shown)


storage_service/locations/models/package.py
```diff
@@ -1049,30 +1050,39 @@ def finish_reingest(self, origin_location, origin_path, reingest_location, reing
...
+                if algo == 'bzip2':
+                    compression = self.COMPRESSION_7Z_BZIP
+                elif algo == 'lzma':
+                    compression = self.COMPRESSION_7Z_LZMA
+                else:
+                    compression = self.COMPRESSION_7Z_BZIP
+                    LOGGER.warning('Reingest: Unable to determine reingested compression algorithm, defaulting to bzip2.')
+            elif puid == 'x-fmt/268':  # Bzipped (probably tar)
+                compression = self.COMPRESSION_TAR_BZIP2
             else:
```
mistydemeo, line 1068: Maybe check this explicitly instead of assuming?

mistydemeo, line 1068: Oh, I see that's what the old code did.

storage_service/locations/models/package.py
```diff
@@ -1049,30 +1050,39 @@ def finish_reingest(self, origin_location, origin_path, reingest_location, reing
...
+            compress_output = ''
 
         # Delete working files
         shutil.rmtree(reingest_full_path)
+
+        if not self.is_compressed:
+            # If temp files are deleted before compression check, the wrong result is returned 
+            shutil.rmtree(temp_dir)
+            self.save()
+            return
```
Hwesta, line 1083: This skips all the code that re-stores the updated AIP.  Was this supposed to be at the end of this function?

Hwesta:
The updated AIP doesn't make it to the storage service. I checked the METS file in the pipeline at the Store AIP microservice, and it had a 'reingestion' event. The AIP in SS internal storage or the final destination didn't have any 'reingestion' events.  This is true for compressed and uncompressed AIPs.  Compressed AIPs are re-stored correctly in qa/0.x.

I tested with AM branch dev/issue-xxxx-reingest-metsrw because there's a couple reingest-related bugs in AM qa/1.x.  If you use dev/issue-xxxx-reingest-metsrw, mets-reader-writer branch dev/reingest will have to be installed.  Alternately, commit e602ac08ce0f42951e27ba376ed40ee24a74be05 in Archivematica fixes the problem.

mcantelon:
Newer version of this is being currently developed as part of https://github.com/artefactual/archivematica-storage-service/pull/110

PR #99, opened 2015-12-04
Add Dataverse Space

Hwesta:
- Transfer Source only
- Space.path should be empty
- Location.path is a search query for Dataverse
- Saves the dataset information as `dataset.json`
- Reads the `dataset.json` to fetch all the files in the dataset
  *\* Tabfiles are the access copies of the original, so fetch the original as a bundle
- Dataverse Agent info is written as an `agents.json`

### DIFF
--- a/storage_service/locations/models/dataverse.py
+++ b/storage_service/locations/models/dataverse.py
@@ -1,6 +1,7 @@
 # stdlib, alphabetical
 import json
 import logging
+import os
 
 # Core Django, alphabetical
 from django.db import models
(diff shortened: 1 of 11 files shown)


storage_service/locations/models/dataverse.py
```diff
@@ -0,0 +1,161 @@
...
+            'entries': entries,
+            'properties': properties,
+        }
+
+    def move_to_storage_service(self, src_path, dest_path, dest_space):
+        """
+        Fetch dataset with ID `src_path` to dest_space.staging_path/dest_path.
+        """
+        # TODO how to strip location path if location isn't passed in?
+        # HACK strip everything that isn't a number
```
mistydemeo, line 103: What do the location paths look like?

mistydemeo:
:+1:

storage_service/locations/models/dataverse.py
Hwesta, line 103: Location path is the search query for Dataverse. Eg `*` or `Part of Title`

Hwesta:
Added a warning to the Space creation page that Dataverse is in beta

PR #100, opened 2015-12-10
Re-ingest: add full option

sevein:
Redmine ticket: https://projects.artefactual.com/issues/9211
See also: artefactual/archivematica#384

I got some initial feedback as part of #86. I filed a new PR because the ticket and the branch of the name was wrong due to some misunderstanding on my side.

I think that what's left is to see if I got right the exception handling in `models/pipeline.py`. I tried to create a couple of meaningful exceptions but now I kind of feel that it's overcomplicated. Let me know what you think please.

### DIFF
--- a/storage_service/locations/forms.py
+++ b/storage_service/locations/forms.py
@@ -220,6 +220,7 @@ class ConfirmEventForm(forms.ModelForm):
         super(ConfirmEventForm, self).__init__(*args, **kwargs)
         self.fields['status_reason'].required = True
 
+
 class CallbackForm(forms.ModelForm):
     class Meta:
         model = models.Callback
@@ -227,9 +228,14 @@ class CallbackForm(forms.ModelForm):
 
 
 class ReingestForm(forms.Form):
-    pipeline = forms.ModelChoiceField(queryset=models.Pipeline.active.all())
     REINGEST_CHOICES = (
-        (models.Package.METADATA_ONLY, 'Re-ingest metadata only'),
-        (models.Package.OBJECTS, 'Re-ingest metadata and objects')
+        (models.Package.METADATA_ONLY, 'Metadata re-ingest'),
+        (models.Package.OBJECTS, 'Partial re-ingest'),
+        (models.Package.FULL, 'Full re-ingest'),
     )
+
+    pipeline = forms.ModelChoiceField(queryset=models.Pipeline.active.all())
     reingest_type = forms.ChoiceField(choices=REINGEST_CHOICES, widget=forms.RadioSelect)
+    processing_config = forms.CharField(required=False, initial='default',
+        help_text='Optional: The processing config is only used with full re-ingest',
+        widget=forms.TextInput(attrs={'placeholder': 'default'}))
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -1116,11 +1116,20 @@ class Package(models.Model):
                 # Remove initial copy of pointer
                 os.remove(os.path.join(ss_internal.full_path, reingest_pointer_name))
 
-        # Replace METS
+        # If there's no reingest-saved METS, copy the original METS there
+        # WARNING If submission documentation is copied later, the original METS will be overwritten.
         original_mets_path = os.path.join(path, 'data', 'METS.' + self.uuid + '.xml')
+        backup_mets_path = os.path.join(path, 'data', 'objects', 'submissionDocumentation', 'METS.' + self.uuid + '.xml')
+        if not os.path.exists(backup_mets_path):
+            LOGGER.info('Moving original METS from %s to %s', original_mets_path, backup_mets_path)
+            os.rename(original_mets_path, backup_mets_path)
+        else:
+            LOGGER.info('Removing intermediate METS %s', original_mets_path)
+            os.remove(original_mets_path)
+
+        # Replace METS
         reingest_mets_path = os.path.join(reingest_full_path, 'data', 'METS.' + self.uuid + '.xml')
-        LOGGER.info('Replacing original METS %s with reingested METS %s', original_mets_path, reingest_mets_path)
-        os.remove(original_mets_path)
+        LOGGER.info('Replacing AIP METS %s with reingested METS %s', original_mets_path, reingest_mets_path)
         os.rename(reingest_mets_path, original_mets_path)
 
         # Replace new metadata files
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -963,7 +963,6 @@ class Package(models.Model):
                     with open(config_path, 'w+') as f:
                         f.write(config)
                     LOGGER.debug('Reingest: processing configuration %s written, location: %s', processing_config, config_path)
-                    reingest_files.append(os.path.join(relative_path, 'processingMCP.xml'))
                 except IOError:
                     LOGGER.exception('Reingest: processing configuration %s could not be written', processing_config)
                     raise
@@ -991,7 +990,10 @@ class Package(models.Model):
 
         # Delete local copy of extraction
         if self.local_path != self.full_path:
-            shutil.rmtree(local_path)
+            try:
+                shutil.rmtree(local_path)
+            except OSError:  # May have been moved not copied
+                pass
         if temp_dir:
             shutil.rmtree(temp_dir)
 
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -997,16 +997,24 @@ class Package(models.Model):
 
         # Call reingest API
         reingest_target = 'transfer' if reingest_type == self.FULL else 'ingest'
+        reingest_uuid = self.uuid
         try:
             resp = pipeline.reingest(relative_path, self.uuid, reingest_target)
         except requests.exceptions.RequestException as e:
             message = 'Error in approve reingest API. {}'.format(e)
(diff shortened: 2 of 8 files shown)


storage_service/locations/models/pipeline.py
```diff
@@ -126,3 +128,76 @@ def create_default_locations(self, shared_path=None):
...
+    def _request_api(self, method, path, **kwargs):
+        url = 'http://{}/api/{}'.format(
+            self.remote_name,
+            path.lstrip('/'),
+        )
+        fields = {
+            'username': self.api_username,
+            'api_key': self.api_key
+        }
+        if 'fields' in kwargs:
```
mistydemeo: If this is the only thing you're using `kwargs` for, maybe make it a defined parameter with a default of `None`?

``` python
def _request_api(self, method, path, fields=None)
# ...

if fields is not None:
    fields.update(kwargs['fields'])
```

storage_service/locations/models/pipeline.py
```diff
@@ -126,3 +128,76 @@ def create_default_locations(self, shared_path=None):
...
+            fields.update(kwargs['fields'])
+
+        LOGGER.debug('URL: %s; fields: %s', url, fields)
+        try:
+            resp = requests.request(method, url, data=fields, allow_redirects=True)
+        except requests.exceptions.RequestException, e:
+            LOGGER.error('Unable to connect to pipeline %s.', self)
+            et, ei, tb = sys.exc_info()
```
mistydemeo: If you're not planning to use `et` or `ei`, mark that by making them `_`:

``` python
_, _, tb = sys.exc_info()
```

sevein:
Thank you! It's done now.

mistydemeo:
The exception handling looks fine to me.

sevein:
Should I merge?

storage_service/locations/models/package.py
```diff
@@ -894,6 +900,27 @@ def start_reingest(self, pipeline, reingest_type):
...
             reingest_files.append(os.path.join(relative_path, 'data', 'objects', 'metadata', ''))
 
+        # Fetch processing configuration, put it in the root of the package and
+        # include the file in reingest_files.
+        if reingest_type == self.FULL and processing_config != 'default':
+            try:
+                config = pipeline.get_processing_config(processing_config)
+            except PipelineClientException:
+                LOGGER.exception('Reingest: processing configuration %s could not be loaded, error: %s', processing_config, e)
+                pass
```
Hwesta: Nitpick: pass not needed - it's a noop useful for marking an empty block, and the log line above means the block isn't empty.

storage_service/locations/models/package.py
```diff
@@ -894,6 +900,27 @@ def start_reingest(self, pipeline, reingest_type):
...
+                try:
+                    # It's not expected to find an existing processingMCP.xml
+                    # file in the original AIP, but we are using the w+ mode
+                    # just in case.
+                    with open(config_path, 'w+') as f:
+                        f.write(config)
+                    LOGGER.debug('Reingest: processing configuration %s written, location: %s', processing_config, config_path)
+                    reingest_files.append(os.path.join(relative_path, 'processingMCP.xml'))
+                except IOError as e:
+                    LOGGER.exception('Reingest: processing configuration %s could not be written, error: %s', processing_config, e)
```
Hwesta: LOGGER.exception will automatically log the exception details including name, so it doesn't need to be written in the error message.

If this error happens, do we want to continue processing, or should this re-raise?

storage_service/locations/models/pipeline.py
```diff
@@ -126,3 +128,76 @@ def create_default_locations(self, shared_path=None):
...
+        content is returned as a string, which may be empty in case of errors.
+        """
+        if not name:
+            raise ValueError('Parameter name is empty.')
+        url = 'processing-configuration/' + name
+        resp = self._request_api('GET', url)
+        if resp.status_code != requests.codes.ok:
+            raise PipelineClientUnexpectedCodeException(self, resp.status_code)
+        if not resp.text:
+            raise PipelineClientException('Pipeline {}: empty processing configuration ({}).'.format(self, name))
```
Hwesta: This or the docstring should be updated to reflect the actual empty config behaviour.

Hwesta:
> I think that what's left is to see if I got right the exception handling in models/pipeline.py. I tried to create a couple of meaningful exceptions but now I kind of feel that it's overcomplicated. Let me know what you think please.

While not wrong, the exception handling looks odd to me. I'd prefer to see the requests.Exception in a traceback, since that's clearer to me what went wrong.  I also wouldn't check for non-Falsey parameters, since subsequent calls that use them will fail.  With such a small use case I would not create custom exceptions, especially since not doing it simplifies the code.  I like the `_request_api` cleanup.

storage_service/locations/models/package.py
sevein: Good points. I'll re-raise the exception.

storage_service/locations/models/package.py
sevein: ok

sevein:
Cool, I followed your recommendations. Let me know what you think.

storage_service/locations/forms.py
```diff
@@ -219,16 +219,22 @@ def __init__(self, *args, **kwargs):
...
+        (models.Package.METADATA_ONLY, 'Metadata re-ingest'),
+        (models.Package.OBJECTS, 'Partial re-ingest'),
+        (models.Package.FULL, 'Full re-ingest'),
     )
+
+    pipeline = forms.ModelChoiceField(queryset=models.Pipeline.active.all())
     reingest_type = forms.ChoiceField(choices=REINGEST_CHOICES, widget=forms.RadioSelect)
+    processing_config = forms.CharField(required=False, initial='default',
+        help_text='Only needed in full re-ingest',
+        widget=forms.TextInput(attrs={'placeholder': 'default'}))
```
jrwdunham, line 241: Is the help text "Only needed in full re-ingest" accurate? I believe the behaviour is that the processing config will only work in full reingest and will not work at all in the other types of ingest. Is this not true? If it is true, then maybe the help_text should be something like "Optional; the processing config will only work with full re-ingests."

Hwesta:
Looks good to me, but someone should review the changes I made.

storage_service/locations/models/package.py
```diff
@@ -941,12 +946,32 @@ def start_reingest(self, pipeline, reingest_type):
         elif reingest_type == self.METADATA_ONLY:
             reingest_files.append(os.path.join(relative_path, 'data', 'objects', 'metadata', ''))
 
+        # Fetch processing configuration, put it in the root of the package and
+        # include the file in reingest_files.
+        if reingest_type == self.FULL and processing_config != 'default':
+            try:
+                config = pipeline.get_processing_config(processing_config)
```
jrwdunham, line 954: I see that the dashboard API does not expose a GET processing-configurations/ endpoint. However, if it did, we could call it before the re-ingest form was generated and then only allow the user to select from available configs. then this get_processing_config call would rarely (never?) fail. Wishful thinking, but thought I'd mention it.

storage_service/locations/models/package.py
```diff
@@ -1112,9 +1141,34 @@ def finish_reingest(self, origin_location, origin_path, reingest_location, reing
...
+                    dupe_preservation_regex = match.group(1) + r'-\w{8}-\w{4}-\w{4}-\w{4}-\w{12}'
+                    for p in os.listdir(dest_dir):
+                        if re.match(dupe_preservation_regex, p):
+                            del_path = os.path.join(dest_dir, p)
+                            LOGGER.info('Deleting %s', del_path)
+                            os.remove(del_path)
+                    # Copy new preservation derivative
+                    LOGGER.info('Moving %s to %s', reingest_preservation_path, original_preservation_path)
+                    shutil.copy2(reingest_preservation_path, original_preservation_path)
+
```
jrwdunham, line 1171: While it's an edge case, I think this block will cause a bug if a transfer contains a file whose original name has a UUID suffix. For example, try to re-ingest a SIP whose original transfer contains a file named "Nemastylis_geminiflora_Flower-a464e051-b6db-46cf-9f45-64c00ae16171.PNG". I tried this and the resulting AIP (after re-ingest) did not contain the .tif derivative that was created during the original ingest.

Assuming I'm correct in identifying this as an error, maybe a better way of identifying preservation derivatives would be to look for files whose names are suffixed by UUIDs AND where there is another file with the same basename, minus the UUID suffix.

jrwdunham (commented):
Two minor comments and one possible edge-case bug.

storage_service/locations/models/package.py
Hwesta: I like that idea!  I think it's out of scope for right now though. The processing-config endpoint is added in artefactual/archivematica#313 so should be exposed in qa/1.x's dashboard API.

storage_service/locations/forms.py
Hwesta: Updated text

storage_service/locations/models/package.py
Hwesta: Nice catch.  Updated so the regex matches on the extension as well (so originalfile-originaluuid.ext doesn't match both originalfile-originaluuid.ext and originalfile-originaluuid-preservationuuid.ext), and added a check to not delete itself if detected as a duplicate preservation file.  I'll rebase the fix into the original commit if it looks good.

jrwdunham:
👍

PR #95, opened 2015-11-19
Fixity values from Arkivum

Hwesta:
- Add Space-specific fixity hook
- Add uncompressed AIP support to Arkivum with new `api/3/ingest-manifest/` endpoints
- Add checking if a file is local for uncompressed bags in Arkivum

Based on top of #88

PR #101, opened 2015-12-21
api: allow http head in extract_file_request

hakamine:
This is required for Binder changes in https://github.com/artefactual/client-moma-atom/pull/68 to work 
Created this PR for merging into the branch dev/issue-8859-arkivum-fixity #95 (not sure if that is correct or should have requested to go into qa/1.x instead)

### DIFF
--- a/storage_service/common/utils.py
+++ b/storage_service/common/utils.py
@@ -98,20 +98,13 @@ def download_file_stream(filepath, temp_dir=None):
         return http.HttpResponseNotFound("File not found")
 
     filename = os.path.basename(filepath)
-    extension = os.path.splitext(filepath)[1].lower()
 
     wrapper = FileWrapper(file(filepath))
     response = http.HttpResponse(wrapper)
 
-    # force download for certain filetypes
-    extensions_to_download = ['.7z', '.zip']
-    if extension in extensions_to_download:
-        response['Content-Type'] = 'application/force-download'
-        response['Content-Disposition'] = 'attachment; filename="' + filename + '"'
-    else:
-        mimetype = mimetypes.guess_type(filename)[0]
-        response['Content-type'] = mimetype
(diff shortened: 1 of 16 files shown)


Hwesta:
This allows it to respond to HEAD, but still returns the full body (ie HEAD behaves like GET).  What's the desired behavior for HEAD? Should it verify the file exists but not return it (ie fetch package from remote location and check file exists, but return empty body)?  If the caller always wants the file, can it use GET?

hakamine:
The HEAD request is used by Binder to verify the file exists. If it does, Binder then uses a GET request to download it, and if it doesn't it uses a GET request to obtain the response error code. According to the notes in https://github.com/artefactual/client-moma-atom/pull/68 the division in these steps (check and download) was done to be able to show download problems as alerts in the Binder UI.
Although it seems to be working (for relatively small files) I think it would be better if SS doesn't return the full body with the HEAD response, could it be implemented easily?

mistydemeo:
It looks like the original function, as written, only tries to verify the existence of the file after a) copying the package to locally-accessible storage, b) extracting file from the AIP (if the AIP is compressed). Looks like it'd need some major restructuring to accommodate checking for the existence of the file without performing expensive operations.

A part of the problem is that we don't store a list of the files within a given package; for compressed packages in particular, we don't actually have a way to verify the existence of a file without checking in the package itself, which requires having a locally-addressible copy available.

mistydemeo:
For uncompressed AIPs, we could probably reduce the load by replacing the call to `full_path = package.fetch_local_path()` in order to fetch only the requested path, and not the entirety of the AIP. Not sure there's a simple way to handle this for compressed AIPs without retrieving the entire AIP though.

Hwesta:
This is included in #136

PR #102, opened 2015-12-23
Mark public Space methods as public

mistydemeo:
Several instance methods of Space were marked as being public, but were primarily, or entirely, called by outside objects. This changes their names to mark them as public.


Hwesta:
My logic was that all Spaces were internal to each other (since Space should be subclassed by its protocol spaces), but this makes sense too.

Commit message uses 'public' the first time when 'private' makes more sense.

:+1:

mistydemeo:
Fixed the commit message, and merged.

PR #104, opened 2016-01-21
Space: Optionally try os.rename in move_rsync

Hwesta:
Add try_mv_local flag to move_rsync. If true, it will try to os.rename source to destination, possibly normalizing the path. This will do a move, not a copy like rsync does by default. This will fail in several cases, notably if crossing filesystem boundaries, and falls back to the normal rsync.

move_from_storage_service calls that use move_rsync locally have been updated to use try_mv_local.

### DIFF
--- a/storage_service/locations/models/arkivum.py
+++ b/storage_service/locations/models/arkivum.py
@@ -82,6 +82,7 @@ class Arkivum(models.Model):
 
     def move_from_storage_service(self, source_path, destination_path):
         """ Moves self.staging_path/src_path to dest_path. """
+        try_mv_local = False
         # Rsync to Arkivum watched directory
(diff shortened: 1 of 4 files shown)


jrwdunham:
Small question. I'm wondering why you chose to move the file (os.rename) and fail (i.e., fallback to rsync) if the destination directory does not exist, instead of mimicking the previous rsync behaviour, i.e., creating the necessary destination directory structure and retaining a copy of the original file at `source`, e.g., by using `Space.create_local_directory` and a `shutil` copy function.

Hwesta:
I think it was because of the potential speed gains; `os.rename` may just update the inode, resulting in a very fast move, while anything that copies the content is slow on large files.

PR #105, opened 2016-01-28
Package: disable smart strings for xpath() calls

mistydemeo:
lxml's xpath() method returns "smart strings" if the XPath expression returns a string, for example if the expression terminates in text(). These smart strings are extended from the core types by including getparent() methods which know their originating element:
http://lxml.de/xpathxslt.html

This has a couple of disadvantages:
1. Memory performance is worse, since it means that the tree is being unnecessarily kept in memory even if it's not being used.
2. When using PyPy and lxml before 3.5.0, PyPy will crash with a stack overflow: http://permalink.gmane.org/gmane.comp.python.lxml.devel/7572

Since we're not using the extra behaviour of these strings, it's better for us to disable it and just return a raw string.

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -660,7 +660,7 @@ class Package(models.Model):
             # If the filename has been sanitized, the path in the fileSec
             # may be outdated; check for a cleanup event and use that,
             # if present.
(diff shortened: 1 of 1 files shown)


Hwesta:
:+1:

PR #108, opened 2016-02-16
Requirements: add defused 0.4.1 (tastypie)

sevein:
As of tastypie v0.9.13, XML serialization requires both defusedxml and lxml.
See http://django-tastypie.readthedocs.org/en/latest/release_notes/v0.9.13.html.


Hwesta:
:+1:

PR #110, opened 2016-02-17
Dev/issue 8893 dip download - work in progress

mcantelon:
- Added ability to related packages to each other
- Fixed issues with reingesting uncompressed AIPs

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -416,6 +416,7 @@ class PackageResource(ModelResource):
     current_location = fields.ForeignKey(LocationResource, 'current_location')
 
     current_full_path = fields.CharField(attribute='full_path', readonly=True)
+    related_packages = fields.ManyToManyField('self', 'related_packages', null=True)
 
     class Meta:
         queryset = Package.objects.all()
@@ -431,7 +432,7 @@ class PackageResource(ModelResource):
         # that name.
         resource_name = 'file'
 
-        fields = ['current_path', 'package_type', 'size', 'status', 'uuid']
+        fields = ['current_path', 'package_type', 'size', 'status', 'uuid', 'related_packages']
         list_allowed_methods = ['get', 'post']
         detail_allowed_methods = ['get', 'put', 'patch']
         allowed_patch_fields = ['reingest']  # for customized update_in_place
@@ -443,6 +444,7 @@ class PackageResource(ModelResource):
             'path': ALL,
             'uuid': ALL,
             'status': ALL,
+            'related_packages': ALL_WITH_RELATIONS
         }
 
     def prepend_urls(self):
@@ -468,13 +470,14 @@ class PackageResource(ModelResource):
 
     def obj_create(self, bundle, **kwargs):
         bundle = super(PackageResource, self).obj_create(bundle, **kwargs)
+        related_package_uuid = bundle.data.get('related_package_uuid')
         # IDEA add custom endpoints, instead of storing all AIPS that come in?
         origin_location_uri = bundle.data.get('origin_location')
         origin_location = self.origin_location.build_related_resource(origin_location_uri, bundle.request).obj
         origin_path = bundle.data.get('origin_path')
         if bundle.obj.package_type in (Package.AIP, Package.AIC, Package.DIP) and bundle.obj.current_location.purpose in (Location.AIP_STORAGE, Location.DIP_STORAGE):
             # Store AIP/AIC
-            bundle.obj.store_aip(origin_location, origin_path)
+            bundle.obj.store_aip(origin_location, origin_path, related_package_uuid)
         elif bundle.obj.package_type in (Package.TRANSFER,) and bundle.obj.current_location.purpose in (Location.BACKLOG,):
             # Move transfer to backlog
             bundle.obj.backlog_transfer(origin_location, origin_path)
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -63,6 +63,7 @@ class Package(models.Model):
         (DEPOSIT, 'FEDORA Deposit')
     )
     package_type = models.CharField(max_length=8, choices=PACKAGE_TYPE_CHOICES)
+    related_packages = models.ManyToManyField('self', related_name='related')
 
     PENDING = 'PENDING'
     STAGING = 'STAGING'
@@ -374,7 +375,7 @@ class Package(models.Model):
         # Do fixity check of AIP with recovered files
         return self.check_fixity()
 
-    def store_aip(self, origin_location, origin_path):
+    def store_aip(self, origin_location, origin_path, related_package_uuid=None):
         """ Stores an AIP in the correct Location.
 
         Invokes different transfer mechanisms depending on what the source and
@@ -437,6 +438,9 @@ class Package(models.Model):
         # Update package status once transferred to SS
         if dest_space.access_protocol not in (Space.LOM, Space.ARKIVUM):
             self.status = Package.UPLOADED
+        if related_package_uuid is not None:
+            related_package = Package.objects.get(uuid=related_package_uuid)
+            self.related_packages.add(related_package)
         self.save()
         dest_space.post_move_from_storage_service(
             staging_path=self.current_path,
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -372,7 +372,7 @@ class Package(models.Model):
         temp_aip.delete()
 
         # Do fixity check of AIP with recovered files
-        return self.check_fixity() 
+        return self.check_fixity()
 
     def store_aip(self, origin_location, origin_path):
         """ Stores an AIP in the correct Location.
@@ -498,9 +498,8 @@ class Package(models.Model):
             rc = subprocess.call(command)
             LOGGER.debug('Extract file RC: %s', rc)
         else:
-            aip_path = os.path.join(full_path, basename)
-            LOGGER.info('Copying AIP from: %s to %s', aip_path, output_path)
-            shutil.copytree(aip_path, output_path)
+            LOGGER.info('Copying AIP from: %s to %s', full_path, output_path)
+            shutil.copytree(full_path, output_path)
 
         if not relative_path:
             self.local_path_location = ss_internal
@@ -980,6 +979,9 @@ class Package(models.Model):
         internal_space = ss_internal.space
         dest_space = reingest_location.space
 
+        # Take note of whether old version of AIP was compressed
+        was_compressed = self.is_compressed
+
         # Copy actual AIP to ss_internal, extract if needed
         path, temp_dir = self.extract_file()  # AIP always copied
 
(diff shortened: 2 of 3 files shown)


storage_service/locations/models/package.py
```diff
@@ -1021,11 +1027,12 @@ def finish_reingest(self, origin_location, origin_path, reingest_location, reing
         # TODO what do if LOCKSS & uncompressed?
         if self.package_type in (Package.AIP, Package.AIC):
             reingest_pointer_src = os.path.join(origin_location.relative_path, os.path.dirname(origin_path), 'pointer.xml')
-            reingest_pointer_name = 'pointer.' + self.uuid + '.reingest.xml'
-            reingest_pointer_dst = os.path.join(ss_internal.relative_path, reingest_pointer_name)
-            src_space.move_to_storage_service(reingest_pointer_src, reingest_pointer_name, internal_space)
-            internal_space.move_from_storage_service(reingest_pointer_name, reingest_pointer_dst)
-            reingest_pointer = os.path.join(ss_internal.full_path, reingest_pointer_name)
+            if os.path.isfile(os.path.join(origin_location.full_path, os.path.dirname(origin_path), 'pointer.xml')):
```
Hwesta: `origin_location.full_path` may not be local to this machine.  For example, if the origin_location's Space is a PipelineFS, only accessible via SSH. This change should probably be removed.

storage_service/locations/models/package.py
```diff
@@ -63,6 +63,7 @@ class Package(models.Model):
         (DEPOSIT, 'FEDORA Deposit')
     )
     package_type = models.CharField(max_length=8, choices=PACKAGE_TYPE_CHOICES)
+    related_packages = models.ManyToManyField('self')
```
Hwesta: This line should probably be in the other commit. Look at git rebase, or ask if you aren't sure how to move it.

It would be helpful to define a [related name](https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.ForeignKey.related_name) and think about if this should be [symmetrical](https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.ManyToManyField.symmetrical) if we leave it as a ManyToMany.

storage_service/locations/api/resources.py
```diff
@@ -467,6 +469,7 @@ def prepend_urls(self):
         ]
 
     def obj_create(self, bundle, **kwargs):
+        related_package_uuid = bundle.data.get('related_package_uuid')
         bundle = super(PackageResource, self).obj_create(bundle, **kwargs)
```
Hwesta, line 472: I don't think the bundle is populated until after the `super` call.  This should be moved afterwards.

storage_service/locations/models/package.py
```diff
@@ -437,6 +438,9 @@ def store_aip(self, origin_location, origin_path):
         # Update package status once transferred to SS
         if dest_space.access_protocol not in (Space.LOM, Space.ARKIVUM):
             self.status = Package.UPLOADED
+        if related_package_uuid is not None:
+            related_package = Package.objects.get(uuid=related_package_uuid)
+            self.related_packages.add(related_package)
```
Hwesta, line 443: What if the related package doesn't exist yet? Eg when a DIP is uploaded (and knows its AIP's UUID), but the AIP hasn't been stored yet? I'm pretty sure Django enforces referential integrity here.

sromkey, line 443: Archivematica does allow you to upload a DIP before storing the AIP, but that is poor practice. We've considered changing this in a future release, by enforcing successful AIP storage before a DIP is allowed to be uploaded.

storage_service/locations/models/package.py
```diff
@@ -1020,12 +1026,13 @@ def finish_reingest(self, origin_location, origin_path, reingest_location, reing
         # Copy pointer file if exists
         # TODO what do if LOCKSS & uncompressed?
         if self.package_type in (Package.AIP, Package.AIC):
-            reingest_pointer_src = os.path.join(origin_location.relative_path, os.path.dirname(origin_path), 'pointer.xml')
-            reingest_pointer_name = 'pointer.' + self.uuid + '.reingest.xml'
-            reingest_pointer_dst = os.path.join(ss_internal.relative_path, reingest_pointer_name)
-            src_space.move_to_storage_service(reingest_pointer_src, reingest_pointer_name, internal_space)
-            internal_space.move_from_storage_service(reingest_pointer_name, reingest_pointer_dst)
-            reingest_pointer = os.path.join(ss_internal.full_path, reingest_pointer_name)
+            if self.pointer_file_path is not None and os.path.isfile(self.pointer_file_path):
```
Hwesta: `self.pointer_file_path` may not exist on this machine.  The pointer file may be on a different machine, and `os.path.isfile` will always return false.  For example, if the package comes from a PipelineFS which is only accessible via SSH. This change should be removed.

storage_service/locations/models/package.py
mcantelon, line 443: I've also got an Archivematica branch that deals with this possibility in the storeAIP microservice... here's the commit:

https://github.com/artefactual/archivematica/commit/e0455b97d8422adf67c36ecbec94f4a7f630daf2

storage_service/locations/api/resources.py
```diff
@@ -128,8 +128,8 @@ def dehydrate(self, bundle):
         return bundle
 
     def obj_create(self, bundle, **kwargs):
-        bundle = super(PipelineResource, self).obj_create(bundle, **kwargs)
         bundle.obj.enabled = not utils.get_setting('pipelines_disabled', False)
+        bundle = super(PipelineResource, self).obj_create(bundle, **kwargs)
```
Hwesta, line 131: Why move this? What problem was it causing?

storage_service/locations/models/package.py
Hwesta, line 443: Thanks

storage_service/locations/models/package.py
```diff
@@ -1019,13 +1028,27 @@ def finish_reingest(self, origin_location, origin_path, reingest_location, reing
...
+            try:
+                src_space.move_to_storage_service(reingest_pointer_src, reingest_pointer_name, internal_space)
+                internal_space.move_from_storage_service(reingest_pointer_name, reingest_pointer_dst)
+
+                self.pointer_file_location = Location.active.get(purpose=Location.STORAGE_SERVICE_INTERNAL)
+                self.pointer_file_path = os.path.join(utils.uuid_to_path(self.uuid), 'pointer.{}.xml'.format(self.uuid))
+
+                reingest_pointer = os.path.join(ss_internal.full_path, utils.uuid_to_path(self.uuid), reingest_pointer_name)
+            except:
+                pass
```
Hwesta: Please log an error or warning.  `self.pointer_file_location` and `self.pointer_file_path` may also need to be updated.

Edit: If `to_be_compressed` is true, then there should always be a pointer file to copy. This should be a real error.

storage_service/locations/models/package.py
```diff
@@ -1097,39 +1130,44 @@ def finish_reingest(self, origin_location, origin_path, reingest_location, reing
...
+                'compression',
+                event_detail=event_detail,
+                event_outcome_detail_note=compress_output,
+            )
 
-        self.update_pointer_file(compression, root=root, path=out_path)
+            self.update_pointer_file(compression, root=root, path=out_path)
+        elif was_compressed:
+            # AIP used to be compressed, but is no longer so delete pointer file
+            os.remove(self.full_pointer_file_path)
```
Hwesta, line 1167: `full_pointer_file_path` is a read only property function. This should update `self.pointer_file_location` and `self.pointer_file_path`

mcantelon, line 1167: I _think_ I've now addressed your concern? If not, let me know.

storage_service/locations/models/package.py
```diff
@@ -994,6 +996,9 @@ def finish_reingest(self, origin_location, origin_path, reingest_location, reing
         )
         reingest_full_path = os.path.join(ss_internal.full_path, reingest_path)
 
+        # Take note of whether new version of AIP should be compressed
+        to_be_compressed = os.path.isfile(os.path.join(ss_internal.full_path, reingest_path))
```
Hwesta: Nitpick: You can reuse reingest_full_path from the line above instead of doing the os.path.join

storage_service/locations/models/package.py
```diff
@@ -1018,14 +1023,25 @@ def finish_reingest(self, origin_location, origin_path, reingest_location, reing
...
-            reingest_pointer_name = 'pointer.' + self.uuid + '.reingest.xml'
-            reingest_pointer_dst = os.path.join(ss_internal.relative_path, reingest_pointer_name)
+
+            # If reingesting a previously compressed AIP, make a temporary "reingest" pointer (otherwise make a normal one)
+            if was_compressed:
+                reingest_pointer_name = 'pointer.' + self.uuid + '.reingest.xml'
+            else:
+                reingest_pointer_name = 'pointer.' + self.uuid + '.xml'
+
+            reingest_pointer_dst = os.path.join(ss_internal.relative_path, utils.uuid_to_path(self.uuid), reingest_pointer_name)
```
Hwesta: Any reason you changed the reingest_pointer_dst path to include the UUID in quads?  Also below with reingest_pointer.

storage_service/locations/models/package.py
Hwesta, line 1167: Mostly!  We need to reset both the pointer_file_path and pointer_file_location

```
# AIP used to be compressed, but is no longer so delete pointer file
os.remove(self.full_pointer_file_path)
self.pointer_file_location = None
self.pointer_file_path = None
```

mcantelon, line 1167: Ah cool... fixed.

storage_service/locations/models/package.py
```diff
@@ -1018,14 +1027,31 @@ def finish_reingest(self, origin_location, origin_path, reingest_location, reing
...
-            reingest_pointer_name = 'pointer.' + self.uuid + '.reingest.xml'
-            reingest_pointer_dst = os.path.join(ss_internal.relative_path, reingest_pointer_name)
+
+            # If reingesting a previously compressed AIP, make a temporary "reingest" pointer (otherwise make a normal one)
+            if was_compressed:
+                reingest_pointer_name = 'pointer.' + self.uuid + '.reingest.xml'
+                reingest_pointer_dst = os.path.join(ss_internal.relative_path, reingest_pointer_name)
+            else:
+                reingest_pointer_name = 'pointer.' + self.uuid + '.xml'
+                reingest_pointer_dst = os.path.join(ss_internal.relative_path, utils.uuid_to_path(self.uuid), reingest_pointer_name)
```
mcantelon, line 1040: This pointer gets created at this location in this case because the old AIP wasn't compressed and so doesn't yet have a pointer stored at that location (and the update_pointer_file function needs a pointer file to update).

Hwesta:
Note: this PR also includes work originally in #98

storage_service/locations/models/package.py
Hwesta, line 1040: That makes sense. Can you add it as a comment to the code?

Hwesta:
Looks good! :+1:

mcantelon:
Merged!

PR #112, opened 2016-02-18
Upgrade to Django 1.8

Hwesta:
Changes to work with Django 1.8
- Use new _meta API calls
- Use new TEMPLATES setting


sevein:
Things seem to be working smoothly, test passes too.

PR #113, opened 2016-03-03
Add gunicorn to production.txt

sevein:
We are using it in RPM installations and Docker. You may still not need it in today's production environments but the alternatives that we are building are also built off `production.txt`.

Still, it's harmless if you don't use it. So environments built with Ansible and others should be fine even if they end up installing gunicorn.

Related: https://github.com/artefactual-labs/am-packbuild/pull/2

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -8,6 +8,7 @@ django-extensions==1.1.1
 django-jsonfield==0.9.12
 django-model-utils==1.3.1
 django-tastypie==0.13.1
+gunicorn==19.4.5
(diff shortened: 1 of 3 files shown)


Hwesta:
If gunicorn is being used in Docker & RPM, then we'll probably put it in the debian/ansible installs.

It should go in base.txt, though, so we don't have mismatched versions if only one gets updated, and so production doesn't have unique installs.

sevein:
Updated.

Hwesta:
:+1:

PR #114, opened 2016-03-16
utils: serve files with http.FileResponse

sevein:
FileResponse is a subclass of StreamingHttpResponse optimized for binary files.
It uses wsgi.file_wrapper if provided by the wsgi server, otherwise it streams
the file out in small chunks.

This class was added in Django 1.7. stable/0.7.x users will have to upgrade.

This may solve a problem where uWSGI processes crash with a timeout error.

### DIFF
--- a/storage_service/common/utils.py
+++ b/storage_service/common/utils.py
@@ -10,7 +10,6 @@ import shutil
 import uuid
 
 from django.core.exceptions import ObjectDoesNotExist
-from django.core.servers.basehttp import FileWrapper
 from django import http
 
 from administration import models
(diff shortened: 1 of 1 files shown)


Hwesta:
Does this cause problems with setting the `content-type`, `content-length` etc below?

If not :+1:

sevein:
I had the same question and double-checked using `curl -v`, all good.

```
< HTTP/1.1 200 OK
* Server nginx is not blacklisted
< Server: nginx
< Date: Thu, 17 Mar 2016 18:22:13 GMT
< Content-Type: application/force-download
< Content-Length: 3159699041
< Connection: keep-alive
< Content-Disposition: attachment; filename="Prueba001-553611ff-7c81-4469-b684-9d18d84ad14c.7z"
< Vary: Accept, Cookie
< Cache-Control: no-cache
< X-Frame-Options: SAMEORIGIN
```

Thank you, Holly!

PR #119, opened 2016-04-04
Locations: Associated with all enabled pipelines by default

Hwesta:
This makes all pipelines selected by default when creating a Location.

It was unclear when there was only one item in the MultiSelectField if it was selected or not.  Select all pipelines by default, so even if it's unclear the expected behavior still happens.


sevein:
Makes sense! :+1:

Hwesta:
Merged to qa/0.x and qa/0.8.x as 038814b8a6c69eaaadf45e44cd5793243789a707

PR #122, opened 2016-04-26
Delete docs/

sevein:
Should this also remove the sphinx dependency from requirements?


Hwesta:
Yes, sphinx can be removed.

:+1:

PR #126, opened 2016-05-18
Report fixity check extraction error. refs #9810

mcantelon:
Add logic to raise an exception when extraction fails during fixity checking.
Return indication that extraction failed from fixity check function.

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -501,6 +501,8 @@ class Package(models.Model):
             LOGGER.info('Extracting file with: %s to %s', command, output_path)
             rc = subprocess.call(command)
             LOGGER.debug('Extract file RC: %s', rc)
+            if rc:
+                raise StorageException('Extraction error')
         else:
             LOGGER.info('Copying AIP from: %s to %s', full_path, output_path)
             shutil.copytree(full_path, output_path)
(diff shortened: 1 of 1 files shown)


storage_service/locations/models/package.py
```diff
@@ -783,7 +785,10 @@ def check_fixity(self, delete_after=True):
         if self.is_compressed:
             # bagit can't deal with compressed files, so extract before
             # starting the fixity check.
-            path, temp_dir = self.extract_file()
+            try:
+                 path, temp_dir = self.extract_file()
+            except StorageException:
+                 return (False, [], 'Error extracting file')
```
Hwesta: I think this should be `None` not `False` - the docstring says `False` is for a verification failure, and `None` is for a scan that couldn't start.

mcantelon: I'm a bit confused aboout the failures return parameter. It seems like it's supposed to be a list, but it gets set to the "details" property of BagValidationError which doesn't seem like it would be a list.

Hwesta: The `BagValidationError` does have a details attribute which is a list - see https://github.com/LibraryOfCongress/bagit-python/blob/master/bagit.py#L558-L566  However, I think for `None`-errors there isn't a list of different validation errors that failed (because no validation was run), and only the `message` is populated.

Hwesta:
:+1:

mcantelon:
Merged.

PR #130, opened 2016-06-14
Dev/issue 9909 move endpoint

mcantelon:
Added package move logic and endpoint (based on stable/0.7.x).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,1 +1,5 @@
 This is the Artefactual Storage Service for Archivematica.
+
+Start worker:
+
+celery -A storage_service worker -l info
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -16,4 +16,5 @@ httplib2
 sword2
 #python-keystoneclient==1.3.0
 #python-swiftclient==2.4.0
-
+django-celery>=3.0,<4.0
+redis>=2.4.4
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -4,6 +4,7 @@
 # stdlib, alphabetical
 import json
 import logging
+from multiprocessing import Process
 import os
 import shutil
 import urllib
@@ -449,6 +450,7 @@ class PackageResource(ModelResource):
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/delete_aip%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('delete_aip_request'), name="delete_aip_request"),
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/recover_aip%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('recover_aip_request'), name="recover_aip_request"),
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/extract_file%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('extract_file_request'), name="extract_file_request"),
+            url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/move%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('move_request'), name="move_request"),
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/download/(?P<chunk_number>\d+)%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('download_request'), name="download_lockss"),
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/download%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('download_request'), name="download_request"),
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/pointer_file%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('pointer_file_request'), name="pointer_file_request"),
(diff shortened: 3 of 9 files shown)


storage_service/locations/models/package.py
```diff
@@ -289,6 +291,61 @@ def _update_quotas(self, space, location):
...
+                package=self)
+
+            # Update location
+            self.current_location = destination_location
+            self.status = Package.UPLOADED
+            self.save()
+            self.current_location.space.update_package_status(self)
+
+        except Exception, e:
+            LOGGER.info('Attempt to move package %s to location %s failed: %s', self.uuid, destination_location_uuid, str(e))
```
Hwesta: `except Exception, e` syntax is deprecated, please use `except Exception as e` if you need it. However, in this case you can use the logging module's `exc_info=True` to automatically add a traceback to the log, which is even better.

storage_service/locations/models/package.py
```diff
@@ -289,6 +291,61 @@ def _update_quotas(self, space, location):
         location.used += self.size
         location.save()
 
+    def move_to_location(self, destination_location_uuid):
```
Hwesta, line 294: Please add a docstring summarizing what this function is doing. If you're not sure of formatting, https://www.python.org/dev/peps/pep-0257/ is a good description.

storage_service/locations/models/package.py
```diff
@@ -289,6 +291,61 @@ def _update_quotas(self, space, location):
...
+            ss_internal = Location.active.get(purpose=Location.STORAGE_SERVICE_INTERNAL)
+
+            source_path = os.path.join(
+                self.current_location.relative_path,
+                self.current_path)
+
+            origin_space = self.current_location.space
+            origin_space.move_to_storage_service(
+                source_path=source_path,
+                destination_path='move',
```
Hwesta, line 323: Does this work if the endpoint is called multiple times simultaneously?

storage_service/storage_service/urls.py
```diff
@@ -125,5 +125,11 @@ def startup():
             utils.set_setting(loc_info['default_setting'], [new_loc.uuid])
             LOGGER.info('Set %s as %s', new_loc, loc_info['default_setting'])
 
+    # Set any interrupted package moves to failed
+    packages = locations_models.Package.objects.filter(status=locations_models.Package.MOVING)
+    for package in packages:
+        package.status = locations_models.Package.MOVE_FAILED
+        package.save()
```
Hwesta: This can be done in one line (and one query), using [`.update()`](https://docs.djangoproject.com/en/dev/ref/models/querysets/#update)  There's a lot of useful things on the queryset reference page, check it out.

storage_service/locations/models/package.py
```diff
@@ -341,6 +341,21 @@ def move_to_location(self, destination_location_uuid):
             self.save()
             self.current_location.space.update_package_status(self)
 
+            # Update pointer file's location information
+            if self.pointer_file_path and self.package_type in (Package.AIP, Package.AIC):
+                root = etree.parse(self.full_pointer_file_path)
+                element = root.find('.//mets:file', namespaces=utils.NSMAP)
+                flocat = element.find('mets:FLocat', namespaces=utils.NSMAP)
+                if self.uuid in element.get('ID', '') and flocat is not None:
+                    flocat.set('{{{ns}}}href'.format(ns=utils.NSMAP['xlink']), self.full_path)
```
Hwesta, line 357: Tip: `utils.PREFIX_NS` has all the namespaces wrapped in `{}` to make this sort of thing easier.  So this could be `flocat.set('{ns}href'.format(ns=utils.PREFIX_NS['xlink']), self.full_path)` or `flocat.set(utils.PREFIX_NS['xlink'] + 'href', self.full_path)`

storage_service/locations/models/package.py
```diff
@@ -341,6 +341,21 @@ def move_to_location(self, destination_location_uuid):
...
+                # Add USE="Archival Information Package" to fileGrp.  Required for
+                # LOCKSS, and not provided in Archivematica <=1.1
+                if root.find('.//mets:fileGrp[@USE="Archival Information Package"]', namespaces=utils.NSMAP) is not None:
```
Hwesta: Should this be `is None`, since if it's not None then it already found something with `@USE="Archival Information Package"` ?

storage_service/locations/models/package.py
```diff
@@ -289,6 +291,76 @@ def _update_quotas(self, space, location):
...
+                    root.find('.//mets:fileGrp', namespaces=utils.NSMAP).set('USE', 'Archival Information Package')
+
+                with open(self.full_pointer_file_path, 'w') as f:
+                    f.write(etree.tostring(root, pretty_print=True))
+
+        except Exception, e:
+            LOGGER.info('Attempt to move package %s to location %s failed: %s', self.uuid, destination_location_uuid, str(e))
```
Hwesta: `except Exception, e` syntax is deprecated, please use `except Exception as e` if you need it. However, in this case you can use the logging module's `exc_info=True` to automatically add a traceback to the log, which is even better.

requirements/base.txt
```diff
@@ -16,4 +16,4 @@ httplib2
 sword2
 #python-keystoneclient==1.3.0
 #python-swiftclient==2.4.0
-
+django-celery-with-redis>=3.0
```
Hwesta: I think it makes more sense to install celery and redis separately, to be more explicit about what versions of things we're installing ind if we change the backend.  Looks like it's `django-celery>=3.0,<4.0` and `redis>=2.4.4`

Hwesta:
Did you forget to commit `celery.py`? It's mentioned in [the docs](http://docs.celeryproject.org/en/latest/django/first-steps-with-django.html) and imported in this code.

mcantelon:
Ah cool, yeah I didn't reglect to add celery.py. I've committed it and separated django-celery and redis in the requirements.

storage_service/locations/api/resources.py
```diff
@@ -529,6 +532,30 @@ def recover_aip_request(self, request, bundle, **kwargs):
...
+            message = "Move already in progress."
+            success = False
+        else:
+            move_package_to_location_task.delay(package.uuid, request_info['location_uuid'])
+            message = "Move initiated."
+            success = True
+
+        response_data = {'success': success, 'message': message}
+
+        return http.HttpResponse(
```
Hwesta, line 554: Since this starts an async process, should it return 201 now?

README.rst
```diff
@@ -1 +1,5 @@
 This is the Artefactual Storage Service for Archivematica.
+
+Start worker:
+
+celery -A storage_service worker -l info
```
Hwesta, line 5: Should https://github.com/artefactual-labs/ansible-role-archivematica-src be updated to start celery too?

storage_service/locations/models/package.py
mcantelon, line 323: Yeah it won't initiate a move twice.

Hwesta:
Duplicated by #143, closing.

PR #131, opened 2016-06-15
Reingest: Recalculate size on reingest

Hwesta:
When storing an AIP after reingest, calculate size instead of trusting what AM provided. For metadata-only reingest, this is incorrect.

refs https://projects.artefactual.com/issues/10030


sevein:
👍

PR #132, opened 2016-06-22
Version and git commit are now displayed under Administration > Version.

jrwdunham:
Redmine: https://projects.artefactual.com/issues/7586

### DIFF
--- a/storage_service/administration/views.py
+++ b/storage_service/administration/views.py
@@ -1,3 +1,4 @@
+import subprocess
 
 from django.contrib import messages
 from django.contrib.auth import get_user_model
@@ -5,6 +6,7 @@ from django.contrib.auth.forms import SetPasswordForm
 from django.shortcuts import render, redirect, get_object_or_404
 
 from common import utils
+from storage_service import __version__ as ss_version
 from . import forms as settings_forms
 
 
@@ -23,6 +25,20 @@ def settings_edit(request):
         messages.success(request, "Setting saved.")
         return redirect('settings_edit')
     return render(request, 'administration/settings_form.html', locals())
+
+
+########################## VERSION ########################
+
+def get_git_commit():
+    try:
+        return subprocess.check_output(['git', 'rev-parse', 'HEAD'])
+    except subprocess.CalledProcessError:
+        return None
+
+def version_view(request):
+    return render(request, 'administration/version.html',
+        {'version': ss_version,
+         'git_commit': get_git_commit()})
 
 
 ########################## USERS ##########################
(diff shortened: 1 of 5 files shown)


storage_service/administration/views.py
```diff
@@ -25,6 +26,29 @@ def settings_edit(request):
     return render(request, 'administration/settings_form.html', locals())
 
 
+########################## VERSION ########################
+
+def get_ss_version():
+    try:
+        from storage_service import __version__ as ss_version
```
Hwesta: When would this import fail?

storage_service/administration/views.py
```diff
@@ -25,6 +26,29 @@ def settings_edit(request):
...
+    except:
+        return 'unknown'
+    if not ss_version:
+        return 'unknown'
+    return ss_version
+
+def get_git_commit():
+    try:
+        return subprocess.check_output(['git', 'rev-parse', 'HEAD'])
+    except:
```
Hwesta: What exception is this expecting to catch?  It's better to catch the most specific exception expected, so unknown errors don't get by, and we don't accidentally silence exceptions like `KeyboardInterrupt` and `SyntaxError`.  http://stackoverflow.com/questions/21553327/why-is-except-pass-a-bad-programming-practice explains more.

storage_service/templates/administration/version.html
```diff
@@ -0,0 +1,26 @@
...
+
+    <tr>
+      <td>Version:</td>
+      <td>{{ version }}</td>
+    </tr>
+
+    {% if git_commit %}
+    <tr>
+      <td>Git commit:</td>
+      <td><a href="https://github.com/artefactual/archivematica-storage-service/commit/{{ git_commit }}"
```
Hwesta, line 17: Nice touch with the link.

storage_service/administration/views.py
jrwdunham: If someone deletes the **version** var.

storage_service/administration/views.py
jrwdunham: Yeah, you're right. I guess I'll change it to catch an `ImportError`.

storage_service/administration/views.py
Hwesta: If someone deletes `__version__` then that's probably a bug we want to be aware of by letting the exception bubble up.  This import can probably be moved to the top of the file (which may mean this function gets folded into `version_view`).

Hwesta: (I assume the ImportError comment below was supposed to be here).  I'd remove the try/catch here - we want the ImportError to bubble up, since if that variable has disappeared that's a bug we don't want to silence.  Also, since it's within our codebase, it seem safe to assume that it will be there (or any change removing it will fix anything that depends on it).

Hwesta:
:+1: Looks good. I'll suggest we wait to merge this until @jhsimpson can look at it as well, since he filed the bug originally.  As you suggested, we should check the behavior on a package install as well.

PR #133, opened 2016-06-29
sword: use shutil.move instead of os.rename

scollazo:
rename fails if src and dst are on different filesystems

https://docs.python.org/3/library/os.html#os.rename


Hwesta:
:+1:

PR #135, opened 2016-07-07
Basename extraction error handling. refs #9810

mcantelon:
Catch error getting basname and raise StorageException so error can be handled by REST endpoint.

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -486,7 +486,10 @@ class Package(models.Model):
 
         # The basename is the base directory containing a package
         # like an AIP inside the compressed file.
-        basename = self.get_base_directory()
+        try:
+            basename = self.get_base_directory()
+        except subprocess.CalledProcessError:
+            raise StorageException('Error determining basename during extraction')
 
         if relative_path:
             output_path = os.path.join(extract_path, relative_path)


storage_service/locations/models/package.py
```diff
@@ -486,7 +486,10 @@ def extract_file(self, relative_path='', extract_path=None):
 
         # The basename is the base directory containing a package
         # like an AIP inside the compressed file.
-        basename = self.get_base_directory()
+        try:
+            basename = self.get_base_directory()
+        except CalledProcessError:
```
Hwesta: Since this is a problem generated inside `get_base_directory`, it makes more sense to handle the exception there.

Hwesta:
This handles the exception, but doesn't address the problem that caused `lsar` to fail. Are we expecting this to run on files that are invalid packages?

mcantelon:
In the issue Hector says "It looks like this is exception is not handled in the PR fix?" so this is to handle it gracefully (while still informing the end user). The lsar issue is likely caused by an invalid/corrupt archive file (I'm just copying it right now to take a look at it).

mcantelon:
Oh I see what you're saying... so we could conceivably change the logic so if it fails extraction it'll still do a fixity check on the archive.

mcantelon:
Sounds like, from talking to Justin, that just failing gracefully is all that's needed.

Hwesta:
Ah, so we are expecting this to run on 'bad' files and fail gracefully. Ok.

Since the actual problem happens inside `get_base_directory`, which is called multiple places, it makes sense to handle all those cases at once (instead of just extract_file) by putting the error handling inside `get_base_directory`.

hakamine:
This PR has been qa'ed and verified to work (for the case of having a corrupted .7z file) in a different branch (ref. 9810).

mcantelon:
Merged into qa/0.x.

PR #136, opened 2016-07-12
Arkivum enhancements

Hwesta:
Replaces #88 and #95  

Arkivum files are NFS-mounted and have local paths but may not be immediately available locally.  Before fetching an Arkivum file check if the file is locally available and return different status codes depending on Arkivum's response.

Endpoints: `/api/v1/file/<uuid>/download/` & `/api/v1/file/<uuid>/extract_file/`
- 200 package locally available (return file)
- 202 package exists but not here now (Arkivum needs to get from tape)
- 404 package doesn't exist
- 502 backend error from Arkivum (assume can't get file)

Changes in this PR:
- Pull package-related fixtures into their own package.json fixture file, reorder
- Add tests for `check_fixity`, `download_request` and `extract_file_request`
- Require `extract_file` to have a GET param of relative_path_to_file, handle error
- Add Space-specific fixity checking, including the remote's timestamp, but allow for forcing a local check
- Add Arkivum fixity checking
- Check if an Arkivum file or package is local before fetching
- Update package status in Arkivum based on replication & fixity results
- `download_request` and `extract_file_request` respond to `HEAD`
- Unlike #95, do not disable SSL checking

TODO
- [ ] Fix download file & download package `HEAD` requests not to behave just like a GET (fetch, return & delete the file)  Some discussion on this in #101


storage_service/locations/fixtures/vcr_cassettes/arkivum_post_move_from_ss.yaml
```diff
@@ -1,16 +1,14 @@
...
       "md5", "size": "10"}'
     headers:
       Accept: ['*/*']
       Accept-Encoding: ['gzip, deflate']
-      Content-Length: ['207']
       Content-Type: [application/json]
       User-Agent: [python-requests/2.3.0 CPython/2.7.3 Linux/3.5.0-52-generic]
     method: POST
-    uri: https://198.50.244.148:8443/api/2/files/release/aips/c0f8/498f/b92e/4a8b/8941/1b34/ba06/2ed8/ep9-c0f8498f-b92e-4a8b-8941-1b34ba062ed8.7z
+    uri: https://198.50.244.148:8443/api/2/files/release/working_bag.zip
```
jhsimpson: I think it would be good to rewrite this fixture, to show localhost instead of the ip, since that is not going to work for anyone else anyways, and it is a private ip for Artefactual, that would be better not to have on github.

storage_service/locations/fixtures/vcr_cassettes/arkivum_update_package_status_uncompressed.yaml
```diff
@@ -0,0 +1,71 @@
+interactions:
+- request:
+    body: null
+    headers:
+      Accept: ['*/*']
+      Connection: [keep-alive]
+      User-Agent: [python-requests/2.4.3 CPython/2.7.3 Linux/3.5.0-54-generic]
+    method: GET
+    uri: https://198.50.244.148:8443/api/3/ingest-manifest/status/5afe9428-c6d6-4d0f-9196-5e7fd028726d
```
jhsimpson: couple other ip addresses that could be replaced with localhost in this fixture

jhsimpson:
:+1: 

Other than a couple minor changes to some fixtures, this looks good to me.  I know it has been tested in one installation.  Merging this to qa/0.x will make it possible for Arkivum to do more testing with it, so I think merging now is a good idea, even with a todo left about HEAD/GET

storage_service/locations/fixtures/vcr_cassettes/arkivum_post_move_from_ss.yaml
Hwesta: Fixed.

PR #137, opened 2016-07-22
SWORD: Save files in object ID directory

Hwesta:
Save files submitted via the FEDORA SWORD API in a directory named after the object ID.  This prevents collisions when multiple METS are submitted with the same object names.

refs #9940 in Redmine

Also, PEP8 & misc cleanup


sevein:
👍

PR #138, opened 2016-08-03
Python3: modernize

Hwesta:
Command: futurize --stage1 --nobackups --write

Also, fix places that use print to use logging instead.


jrwdunham:
👍

PR #140, opened 2016-08-17
Use explicit sword2 version

scollazo:
This fixes an issue with logging in CentOS , as there are some changes in the sword2_logging.py that are not avaliable in the pypi version.

Refs #10187


requirements/base.txt
```diff
@@ -20,5 +20,5 @@ python-dateutil==2.4.2
 python-keystoneclient==1.8.1
 python-swiftclient==2.4.0
 requests>=2.3.0
-sword2
+-e git://github.com/swordapp/python-client-sword2.git@0c241438f2ced13ec3364caa03edf28d5ed5c02c#egg=sword2
```
sevein: Do we need to make it editable?

sevein:
I think we should wait for @jhsimpson and @Hwesta. If we are doing this perhaps we should also update `debian/rules` (using `--find-links`) and delete `lib/sword2-0.1.tar.gz`?

requirements/base.txt
sevein: Should we use the `https` transport instead of `git` to avoid issues when installing SS behind a firewall?

scollazo:
Requirements are avalible in git, because in the past,  canonical/launchpad was the one that built our packages, and they don't allow internet access during builds. Now that we build them, this restriction doesn't exist.

To avoid breaking the workflow, but also refresh the pip packages each time, now we download python requirements as a step of [packbuild.py](https://github.com/artefactual-labs/am-packbuild/blob/dev/rpm-packaging/packbuild.py#L222)

sevein:
Ok I get it, so we still need `--find-links` in `debian/rules` but `lib/sword2-0.1.tar.gz` can be deleted?

scollazo:
Yes, the debian/rules doesn't need any update. I think that the sword2-0.1.tar.gz can be safely removed, but also want to hear from @jhsimpson and @Hwesta about this.

scollazo:
cherry-picked as 9c0e51b

scollazo:
@sevein , I think this addresses all your concerns:
- we use git over https, to avoid firewall issues
- sword2-0.10.tar.gz has been removed
- the dependency is not editable

scollazo:
One doubt about the naming convention, once I merge this, should I tag the release as v0.9.1 or as v0.9.0-2 (being v0.9.0-1 the first one)?

sevein:
👍

sevein:
> One doubt about the naming convention, once I merge this, should I tag the release as v0.9.1 or as v0.9.0-2 (being v0.9.0-1 the first one)?

Let's discuss that somewhere else.

PR #142, opened 2016-08-31
Use shutil.move instead of os.rename

scollazo:
Avoid problems when moving files between different devices


Hwesta:
I don't think this is necessary.  The docs for `shutil.move` say "If the destination is on the current filesystem, then os.rename() is used. Otherwise, src is copied (using shutil.copy2()) to dst and then removed."  However, we handle the case of it being on a different filesystem by falling back to using rsync if rename fails. Using rsync also allows us to copy to/from other servers.

PR #143, opened 2016-09-02
Add new api endpoint for AIP Migration

jhsimpson:
This PR is meant to replace https://github.com/artefactual/archivematica-storage-service/pull/130 which is based on 0.7.x.

This PR brings in a new feature, allowing AIPs to be moved from one Location to another.  This is done though a new API endpoint.

This also brings in celery and redis as new dependencies.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,1 +1,5 @@
 This is the Artefactual Storage Service for Archivematica.
+
+Start worker:
+
+celery -A storage_service worker -l info
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -427,13 +427,14 @@ class PackageResource(ModelResource):
         detail_uri_name = 'uuid'
         always_return_data = True
         filtering = {
-            'location': ALL_WITH_RELATIONS,
+            'current_location': ALL_WITH_RELATIONS,
             'package_type': ALL,
             'path': ALL,
             'uuid': ALL,
             'status': ALL,
             'related_packages': ALL_WITH_RELATIONS
         }
+        ordering = ['uuid']
 
     def prepend_urls(self):
         return [
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -4,7 +4,6 @@
 # stdlib, alphabetical
 import json
 import logging
-from multiprocessing import Process
 import os
 import shutil
 import urllib
@@ -35,6 +34,7 @@ from ..models import (Callback, CallbackError, Event, File, Package, Location, S
 from ..forms import LocationForm, SpaceForm
 from ..constants import PROTOCOL
 from locations import signals
+from locations.tasks import move_package_to_location_task
 
 LOGGER = logging.getLogger(__name__)
 
@@ -608,8 +608,7 @@ class PackageResource(ModelResource):
             message = "Move already in progress."
             success = False
         else:
-            p = Process(target=package.move_to_location, args=(request_info['location_uuid'], ))
-            p.start()
+            move_package_to_location_task.delay(package.uuid, request_info['location_uuid'])
             message = "Move initiated."
             success = True
 
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -4,6 +4,7 @@
 # stdlib, alphabetical
 import json
 import logging
+from multiprocessing import Process
 import os
 import shutil
 import urllib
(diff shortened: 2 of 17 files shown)


storage_service/locations/models/package.py
```diff
@@ -317,6 +319,67 @@ def _update_quotas(self, space, location):
...
+            ss_internal = Location.active.get(purpose=Location.STORAGE_SERVICE_INTERNAL)
+
+            source_path = os.path.join(
+                self.current_location.relative_path,
+                self.current_path)
+
+            origin_space = self.current_location.space
+            origin_space.move_to_storage_service(
+                source_path=source_path,
+                destination_path='move',
```
Hwesta: This should probably be a temp directory, so multiple AIPs being moved don't interfere with each other.

storage_service/locations/models/package.py
```diff
@@ -317,6 +319,67 @@ def _update_quotas(self, space, location):
...
+                source_path=source_path,
+                destination_path=destination_path)
+            destination_space.post_move_from_storage_service(
+                staging_path=source_path,
+                destination_path=destination_path,
+                package=self)
+
+            # Update location
+            self.current_location = destination_location
+            self.status = Package.UPLOADED
```
Hwesta: Do we know this is the correct status for all Spaces?

storage_service/locations/models/package.py
```diff
@@ -317,6 +319,83 @@ def _update_quotas(self, space, location):
...
+                # LOCKSS, and not provided in Archivematica <=1.1
+                if root.find('.//mets:fileGrp[@USE="Archival Information Package"]', namespaces=utils.NSMAP) is None:
+                    root.find('.//mets:fileGrp', namespaces=utils.NSMAP).set('USE', 'Archival Information Package')
+
+                with open(self.full_pointer_file_path, 'w') as f:
+                    f.write(etree.tostring(root, pretty_print=True))
+
+        except Exception:
+            LOGGER.info('Attempt to move package %s to location %s failed', self.uuid, destination_location_uuid, exc_info=True)
+            self.status = Package.MOVE_FAILED
```
Hwesta: Do we need a `MOVE_FAILED` status or can we just display an error message in the GUI? Should it be set back to the previous state? Do we do different behaviour if the package failed move?

storage_service/locations/api/resources.py
```diff
@@ -592,6 +595,30 @@ def recover_aip_request(self, request, bundle, **kwargs):
...
+        """
+        Move package to another location
+        """
+        request_info = bundle.data
+        package = bundle.obj
+
+        if package.status == Package.MOVING:
+            message = "Move already in progress."
+            success = False
+        else:
```
Hwesta: Do we want to move on every package status?  Probably not for Failed, Deleted, Pending (not yet on SS), maybe not for Delete requested or recovery request.

storage_service/locations/models/package.py
```diff
@@ -317,6 +319,83 @@ def _update_quotas(self, space, location):
         location.used += self.size
         location.save()
 
+    def move_to_location(self, destination_location_uuid):
+        """
+        Move a package to another location (that has the same purpose as package's current location).
+        """
+        destination_location = Location.active.get(uuid=destination_location_uuid)
```
Hwesta, line 326: Should this be checked in `resources.move_request`? It's user-submitted data, we should verify the location exists before trying to start a move and immediately erroring.

storage_service/locations/api/resources.py
```diff
@@ -592,6 +595,30 @@ def recover_aip_request(self, request, bundle, **kwargs):
...
+            move_package_to_location_task.delay(package.uuid, request_info['location_uuid'])
+            message = "Move initiated."
+            success = True
+
+        response_data = {'success': success, 'message': message}
+
+        return http.HttpResponse(
```
Hwesta, line 620: Since this starts an async process, should it return 201 now?

README.rst
```diff
@@ -1 +1,5 @@
 This is the Artefactual Storage Service for Archivematica.
+
+Start worker:
+
+celery -A storage_service worker -l info
```
Hwesta, line 5: Has the ansible role been updated to start celery too?

jhsimpson: Not yet.  Updating the archivematica ansible role is going to be addressed in this ticket:
https://projects.artefactual.com/issues/10285, which is assigned to hakamine.  

That ticket will blocks merging this PR, I think.

storage_service/locations/models/package.py
hakamine: also the destination path 'move' seems not being created elsewhere, so the rsync triggered `move_to_location()` produces an error if the directory `/var/archivematica/storage_service/move` was not manually created beforehand.

README.rst
hakamine: This is being done in this branch of the ansible role (not merged to master yet) https://github.com/artefactual-labs/ansible-archivematica-src/tree/dev/ss-celery-worker

storage_service/locations/api/resources.py
hakamine: @Hwesta @jhsimpson would it be too restrictive to allow only moves of packages in the UPLOADED state? Asking because the code seems to be setting the state to UPLOADED if the move was successful, so probably this state was assumed? ( https://github.com/artefactual/archivematica-storage-service/pull/143/files#diff-5f2ead45b8622f27e9b0bf6333334fffR380 )

sevein:
I'm trying to think on not so ideal scenarios where web or task workers hang forever or exit unexpectedly, or race conditions in general in the context of this feature. Is the code safe enough? There are things like [database transactions](http://celery.readthedocs.io/en/latest/userguide/tasks.html#database-transactions), Celery's task [result/state backend](http://celery.readthedocs.io/en/latest/userguide/tasks.html#result-backends), custom task identification (e.g. `task_id` named after package UUID for easy discovery), use of `CELERY_TRACK_STARTED`, handlers ([on_success](http://celery.readthedocs.io/en/latest/userguide/tasks.html#on_success), [on_failure](http://celery.readthedocs.io/en/latest/userguide/tasks.html#on_failure)), timeouts, retries, [global locking](http://loose-bits.com/2010/10/distributed-task-locking-in-celery.html) and a few of other things that we should probably consider before we move ahead to guarantee data integrity. I find it can be overwhelming but should we at least try to cover the basics?

There is a list with links to projects using Celery at their wiki: https://github.com/celery/celery/wiki#projects--applications-usingfor-celery. It's hard to find examples that are not too basic, on the other hand the [pulp](https://github.com/pulp/pulp) project looks very sophisticated.

README.rst
hakamine: ansible role updated, merged to master branch  https://github.com/artefactual-labs/ansible-archivematica-src/pull/92

PR #145, opened 2016-09-26
Update to support 4.5.0 Version of Dataverse

jhsimpson:
refs #8693 dataverse support

This is part of the OCUL sponsored Dataverse Integration Project.  Initial support for Dataverse as a Storage Service Space was merged into qa/0.x already.  This change is to update that support, to work with the new metadata.json files produced by Dataverse.

Dataverse >= 4.4.x made a backwards incompatible change in the way it produces metadata.json files.  The word 'datafile' was replaced with 'dataFile', and within the dataFile element, the 'name' property was changed to 'filename'.

### DIFF
--- a/storage_service/locations/models/dataverse.py
+++ b/storage_service/locations/models/dataverse.py
@@ -133,9 +133,9 @@ class Dataverse(models.Model):
 
         # Fetch all files in dataset.json
         for file_entry in dataset['latestVersion']['files']:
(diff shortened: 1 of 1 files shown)


Hwesta (approved):
Looks good! :+1:

Hwesta:
Oops, missed that this needs a tests update too.

Hwesta:
Note: This was originally added in #99

PR #146, opened 2016-09-26
Add DSpace Space

Hwesta:
Add DSpace Space to the storage service
- POST AIP to DSpace using SWORD2 API
- Parse AIP DublinCore metadata and send with SWORD2
- Split package into two bitstreams: objects and metadata
- Set custom permissions on the metadata bitstream
- Return misc attributes on a Package in metadata
- move_from_storage_service accepts a Package parameter - wants discussion

This branch needs some cleanup to remove experiments that didn't work, and clean up commit history.


storage_service/locations/models/dspace.py
```diff
@@ -0,0 +1,359 @@
...
+# This module, alphabetical
+from . import StorageException
+from common import utils
+from location import Location
+
+LOGGER = logging.getLogger(__name__)
+
+
+class DSpace(models.Model):
+    """ Spaces found in the local filesystem of the storage service."""
```
jrwdunham: Docstring should be changed here, no? This is just copy-pasted from models/local_filesystem.py.

storage_service/locations/models/dspace.py
```diff
@@ -0,0 +1,359 @@
...
+# Third party dependencies, alphabetical
+from lxml import etree
+import requests
+import sword2
+import jsonfield
+
+# This project, alphabetical
+
+# This module, alphabetical
+from . import StorageException
```
jrwdunham: Imported but not used.

jrwdunham (commented):
Aside from 2 very minor comments regarding dspace.py, 👍 from me.

storage_service/locations/models/dspace.py
Hwesta: Fixed.

storage_service/locations/models/dspace.py
Hwesta: Good catch, fixed.

PR #148, opened 2016-10-06
iRODS plugin

orzel:
Hello,

We at Ymagis/Eclair (http://www.ymagis.com, http://www.eclairgroup.com/) have developped a plugin to connect the Archivematica Storage Service to an irods backend. It was tested using irods 4.1.8. Here's a patch against ass-0.8. It contains documentation and (basic) tests. You'll need a working irods instance for those tests.

We use this plugin exclusively to 'deliver' AIPs to iRODS.

iRods provides some python code to access irods, and this plugin is based on this. Unfortunately, as explained in the embedded documentation, this python wrapper misses most performance-related features.
We think that those performance issues should impact very few people. You really need to handle very big files or AIPs with a huge number of files to be hit by those issues.

The plugin is released under the same licence as ass (GNU Affero GPL). We would be happy for any feedback and even more if it was included in storage-service proper.

best regards,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 *.swp
 *.orig
 *~
+sword2_logging.conf
 
 # Sqlite DBs
 *.db
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 *.py[co]
 *.swp
+*.orig
 *~
 
 # Sqlite DBs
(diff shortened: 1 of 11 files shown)


scollazo:
test this please

jhsimpson:
test this please

jhsimpson:
Hi @orzel - thanks for contributing this. I know this is a very belated response, but hopefully late is better than never.

I have a couple comments and questions which I will add inline in other comments, but I think the first item is to verify that you are okay with signing a contributor licence agreement, as outlined here: https://wiki.archivematica.org/Contributor_Agreement .

orzel:
Sorry, i kinda gave up..

PR #150, opened 2016-11-01
Fixity: None is logged separately from True/False

Hwesta:
If a fixity check returns None (because it didn't run because of an external condition), log that differently in the FixityLog. Update the database to handle NULL in the BooleanField, and remove email alert.

Refs Redmine #10326


sevein:
👍

jrwdunham:
👍

PR #152, opened 2016-11-18
Combine user edit and password change forms

helenst:
They are now controlled by a single button - this is to prevent users
from accidentally not saving their user data due to not realising saves
must be done separately.

The two forms remain separate (as they are based on Django forms) but
the view will treat them atomically - data will be saved on both forms
if and only if they are both correctly filled in. (Note that in the
case of the password form, correctly filled in can mean empty - the user
may choose not to change their password).

Tests also added.

### DIFF
--- a/storage_service/administration/forms.py
+++ b/storage_service/administration/forms.py
@@ -2,6 +2,7 @@
 from django import forms
 from django.contrib import auth
 from django.utils.translation import ugettext as _, ugettext_lazy as _l
+from django.core.exceptions import ValidationError
 
 from common import utils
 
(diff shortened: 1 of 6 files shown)


Hwesta (approved):
Thank you for the contribution, this look great!  Especially pleased to see tests for the form & view, since I've been unsure how to do this and this example is very clear.

scollazo:
Please Jenkins, build this. Thanks

scollazo:
test this please

sevein:
Thank you so much!

PR #159, opened 2016-12-01
Storage service i18n

helenst:
Have marked up the following for translation:
* templates
* messages generated by python code
* help strings, label text etc.
* strings from js (with translation catalog).

Set up with transifex, and there's a language switching form.


helenst:
Yes @qubot I have! :)

storage_service/locations/api/sword/views.py
```diff
@@ -146,27 +147,27 @@ def collection(request, location):
                         else:
                             return _deposit_receipt_response(request, deposit, 200)
                     else:
-                        return helpers.sword_error_response(request, 412, 'Error parsing XML')
+                        return helpers.sword_error_response(request, 412, _('Error parsing XML'))
```
sevein: We have a similar string that ends with a dot: `Error parsing XML.`.

sevein: Can we use one of the two only?

storage_service/templates/locations/package_request.html
```diff
@@ -65,7 +69,7 @@
         <td>{{ req.package.get_package_type_display }}</td>
         <td>{{ req.event_reason }}</td>
         <td>{{ req.pipeline }}</td>
-        <td>{{ req.user_email }} (ID: {{ req.user_id}})</td>
+        <td>{{ req.user_email }} ({% trans "ID" %}: {{ req.user_id}})</td>
```
sevein: Should we convert this one into `{email} ({id}): {user_id}`?

storage_service/templates/locations/pipeline_form.html
```diff
@@ -1,6 +1,7 @@
 {% extends "base.html" %}
+{% load i18n %}
 
-{% block page_title %}{{action}} Pipeline{% endblock %}
+{% block page_title %}{% blocktrans %}{{action}} Pipeline{% endblocktrans %}{% endblock %}
```
sevein: I ran into a similar case while I was working on the archivematica-fpr-admin project and I wasn't totally sure if every language is going to be fully covered with just one translation for all the permutations of "{{ action }} Pipeline" - perhaps would be more convenient for translators to have multiple strings, e.g. "Edit pipeline", "Open pipeline", "Delete pipeline". I'm not suggesting you to change it but I figured I'd raise the point for further discussion.

storage_service/templates/login.html
```diff
@@ -21,7 +22,7 @@
 </tr>
 </table>
 
-<input type="submit" value="login" class='btn btn-primary' />
+<input type="submit" value="{% trans "login" %}" class='btn btn-primary' />
```
sevein: Is this necessary?

storage_service/locations/views.py
```diff
@@ -128,14 +129,18 @@ def _handle_package_request(request, config, view_name):
                     event.package.status = config.approved_status
                     success, err_msg = config.execution_logic(event.package)
                     if not success:
-                        error_message = "{}: {}. Please contact an administrator or see logs for details.".format(
-                            config.execution_fail_message, err_msg)
+                        error_message = "{}: {}. {}".format(
```
sevein, line 132: I think `{}: {}. {}` should be the literal that we want to annotate.

storage_service/locations/models/package.py
```diff
@@ -1220,7 +1224,7 @@ def finish_reingest(self, origin_location, origin_path, reingest_location, reing
         # If AIP is a directory, calculate size recursively
         if os.path.isdir(out_path):
             size = 0
-            for dirpath, _, filenames in os.walk(out_path):
+            for dirpath, dirs, filenames in os.walk(out_path):
```
sevein, line 1234: I think that we use `_` when it's unused. But I like it both ways.

storage_service/static/js/project.js
```diff
@@ -1,5 +1,30 @@
 
-
 $(document).ready(function() {
-    $('.datatable').dataTable();
+    $('.datatable').dataTable({
+      // List of language strings from https://datatables.net/reference/option/language
+      oLanguage: {
```
sevein, line 5: I wasn't familiar with the `oLanguage` setting, I've been using just `language`. Are the equivalent? I followed https://datatables.net/manual/i18n.

sevein: Oh I see, that's the way it was done in v1.9? Ok. I upgraded to v1.10 in other places but this is good!

storage_service/administration/views.py
```diff
@@ -76,3 +77,24 @@ def user_create(request):
...
+
+
+########################## LANGUAGE ##########################
+
+def change_language(request):
+    current_language = get_language()
+    language_selection = None
+
+    # Fall back to a generic language if the selected language
+    # is too specific (e.g. en-us can fall back to en)
```
sevein, line 89: Did you have to do this because `DEFAULT_LANGUAGE` defaults to `en-us`? I think it's ok to change it to `en`. For some reason I didn't run into any issues related to this in the Dashboard.

storage_service/storage_service/urls.py
```diff
@@ -26,6 +26,7 @@
     url(r'^api/', include(locations.api.urls)),
 
     url(r'^jsi18n/$', views.cached_javascript_catalog, {'domain': 'djangojs'}, name='javascript-catalog'),
+    url(r'^i18n/', include('django.conf.urls.i18n', namespace='i18n')),
```
sevein, line 29: I didn't know about URL namespaces, that's handy.

sevein:
CC @Hwesta

storage_service/administration/tests/test_languages.py
```diff
@@ -0,0 +1,38 @@
+from django.contrib.auth.models import User
+from django.test import TestCase, override_settings
+
+
+class TestLanguageSwitching(TestCase):
```
sevein, line 5: 🎉  tests!

storage_service/templates/locations/pipeline_form.html
helenst: I'm not sure about this either. I suspect the answer will be to pass the full string from the view to the template rather than attempt to construct it.

storage_service/templates/login.html
helenst: I figured so, since it's a user facing string. I guess it should be Log In (the verb) though - and similarly for the page title and log out button This would also match the main dashboard.

storage_service/static/js/project.js
helenst: Exactly! :sunglasses:

storage_service/locations/models/package.py
helenst: This was to prevent shadowing as `gettext` is imported as `_` now. Unlikely to cause actual problems with the tiny scope that it applies to, but linters like to complain. Alternatives would be to keep the linter quiet about it (I think it's pylint giving me this warning), or import gettext as something else

storage_service/static/js/project.js
helenst: I guess Storage could be upgraded to 1.10 here too, but I'd feel happier treating it as a separate task :)

storage_service/administration/views.py
helenst: Yes... perhaps I should try it as just `en`. The issue here was no default selection on the languages switcher form when the user hadn't yet chosen anything for themselves.

sevein: Interesting, it didn't happen to me. Perhaps you tried before `locale/en` was populated with `makemessages`?

storage_service/templates/login.html
sevein: Sorry, I was thinking on a `<button>`. That makes sense and it looks like Django doesnt' really care about what value you're passing as long as it is a `POST`.

Hwesta (approved):
Looks good to me!  I see you already responded to Jesús' suggestions, and I'm happy to see tests.

sevein:
Replaced by #174.

PR #161, opened 2016-12-04
Add coverage requirement / config

helenst:
Thinking about Python 3 compatibility, I wanted to see what coverage was like so set up the pytest-cov plugin. (It's reporting 47%)

At the moment, it can just be run manually (`py.test --cov`) but could be part of a future CI.


Hwesta:
This looks great, thanks for setting it up! It's missing the `coverage` dependency itself, and I think the `.coveragerc` should be in the same directory as `manage.py`, since that's where the tests are run from.

helenst:
Thanks! I'll add the coverage requirement to the other PR too. (https://github.com/artefactual/archivematica/pull/539)

Hwesta (approved):

sevein:
Thanks @helenst, this is now in qa/0.x as fc9193a.

PR #162, opened 2016-12-06
Fixes metadata-only AIP re-ingest of __MACOSX-zipped files

jrwdunham:
Fixes issue #10045 where AIP metadata-only reingest was failing because python-bagit was registering a different number of bytes on the re-ingest than were documented in the bagit oxum file.

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -104,18 +104,6 @@ class Package(models.Model):
     PACKAGE_TYPE_CAN_RECOVER = (AIP, AIC)
     PACKAGE_TYPE_CAN_REINGEST = (AIP, AIC)
 
-    # Compression options
-    COMPRESSION_7Z_BZIP = '7z with bzip'
-    COMPRESSION_7Z_LZMA = '7z with lzma'
-    COMPRESSION_TAR = 'tar'
-    COMPRESSION_TAR_BZIP2 = 'tar bz2'
-    COMPRESSION_ALGORITHMS = (
-        COMPRESSION_7Z_BZIP,
-        COMPRESSION_7Z_LZMA,
-        COMPRESSION_TAR,
-        COMPRESSION_TAR_BZIP2,
-    )
-
     # Reingest type options
     METADATA_ONLY = 'metadata' # Re-ingest metadata only
     OBJECTS = 'objects'        # Re-ingest metadata and objects for DIP generation
@@ -522,10 +510,30 @@ class Package(models.Model):
             output_path = os.path.join(extract_path, basename)
 
         if self.is_compressed:
-            command = ['unar', '-force-overwrite', '-o', extract_path, full_path]
+            # The command used to extract the compressed file at
+            # full_path was, previously, universally::
+            #
+            #     $ unar -force-overwrite -o extract_path full_path
+            #
+            # The problem with this command is that unar treats __MACOSX .rsrc
+            # ("resource fork") files differently than 7z and tar do. 7z and
+            # tar convert these .rsrc files to ._-prefixed files. Similar
+            # behaviour with unar can be achieved by passing `-k hidden`.
+            # However, while a command like::
+            #
+            #     $ unar -force-overwrite -k hidden -o extract_path full_path
+            #
+            # preserves the .rsrc MACOSX files as ._-prefixed files, it does so
+            # differently than 7z/tar do: the resulting .-prefixed files have
+            # different sizes than those created via unar. This makes
+            # ``bag.validate`` choke.
+            if self.full_pointer_file_path:
+                compression = utils._get_compression(self.full_pointer_file_path)
+            else:
+                compression = None  # no pointer file :. command will be unar
+            command = self._get_decompr_cmd(compression, extract_path, full_path)
             if relative_path:
                 command.append(relative_path)
-
             LOGGER.info('Extracting file with: %s to %s', command, output_path)
             rc = subprocess.call(command)
             LOGGER.debug('Extract file RC: %s', rc)
@@ -546,7 +554,7 @@ class Package(models.Model):
         Produces a compressed copy of the package.
 
         :param algorithm: Compression algorithm to use. Should be one of
-            :const:`Package.COMPRESSION_ALGORITHMS`
+            :const:`utils.COMPRESSION_ALGORITHMS`
         :param str extract_path: Path to compress to. If not provided, will
             compress to a temp directory in the SS internal location.
         :return: Tuple with (path to the compressed file, parent directory of
@@ -558,8 +566,8 @@ class Package(models.Model):
         if extract_path is None:
             ss_internal = Location.active.get(purpose=Location.STORAGE_SERVICE_INTERNAL)
             extract_path = tempfile.mkdtemp(dir=ss_internal.full_path)
-        if algorithm not in self.COMPRESSION_ALGORITHMS:
-            raise ValueError('Algorithm %s not in %s' % algorithm, self.COMPRESSION_ALGORITHMS)
+        if algorithm not in utils.COMPRESSION_ALGORITHMS:
+            raise ValueError('Algorithm %s not in %s' % algorithm, utils.COMPRESSION_ALGORITHMS)
 
         full_path = self.fetch_local_path()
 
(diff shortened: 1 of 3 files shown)


storage_service/locations/models/package.py
```diff
@@ -845,6 +878,8 @@ def check_fixity(self, force_local=False, delete_after=True):
             failures = []
             message = ""
         except bagit.BagValidationError as failure:
+            LOGGER.info('bagit.BagValidationError on %s:\n%s', path, failure.message)
```
sevein: Should we use `LOGGER.error` or `LOGGER.exception`?

storage_service/locations/models/package.py
```diff
@@ -522,10 +522,43 @@ def extract_file(self, relative_path='', extract_path=None):
...
+            #                    '8f41dc6f-05eb-46d4-9b22-0a0d74673510',
+            #                    'b8864e2c-87a0-4aec-a8cf-a747a073be28');
+            #
+            # TODOs/Questions:
+            # 1. Will the relative/subpath final argument to 7z work in the
+            #    same way that it does in unar?
+            # 2. Should/can the extractContents client script be used here so
+            #    that extraction is uniform across AM/AM-SS?
+            # 3. Why was unar chosen here originally instead of 7z? Was there a
+            #    good reason?
```
Hwesta: Unar was chosen because it handles a variety of packages - 7z can only open .7z files, zip can only open .zip files, tar can only handle .tar.bz files etc, while unar can handle [several formats](https://unarchiver.c3.cx/formats)

storage_service/locations/models/package.py
```diff
@@ -522,10 +522,43 @@ def extract_file(self, relative_path='', extract_path=None):
...
+            command = ['7z', 'x', '-bd', '-y', '-o{}'.format(extract_path),
```
Hwesta: Minor: Since `command` is the tokenized list of what's getting passed to subprocess, ` '-o', extract_path,` should work instead of ` '-o{}'.format(extract_path),`

storage_service/locations/models/package.py
```diff
@@ -845,6 +878,8 @@ def check_fixity(self, force_local=False, delete_after=True):
             failures = []
             message = ""
         except bagit.BagValidationError as failure:
+            LOGGER.info('bagit.BagValidationError on %s:\n%s', path, failure.message)
+            LOGGER.debug(subprocess.check_output(['tree', '-a', '--du', path]))
```
Hwesta: I like the idea of checking the output worked, but if the subprocess fails it will raise an exception and halt the fixity check for a debug message, which seems problematic. Remove this or wrap it in exception handling?

storage_service/locations/models/package.py
```diff
@@ -522,10 +522,43 @@ def extract_file(self, relative_path='', extract_path=None):
...
+            #   FROM fpr_fpcommand
+            #     WHERE uuid IN ('a03a3a6d-5695-4800-a16d-edd7143cc438',
+            #                    '8f41dc6f-05eb-46d4-9b22-0a0d74673510',
+            #                    'b8864e2c-87a0-4aec-a8cf-a747a073be28');
+            #
+            # TODOs/Questions:
+            # 1. Will the relative/subpath final argument to 7z work in the
+            #    same way that it does in unar?
+            # 2. Should/can the extractContents client script be used here so
+            #    that extraction is uniform across AM/AM-SS?
```
Hwesta: I like that idea of using the FPR to handle the different extraction commands! It's been discussed that the storage service should have workers & FPR access for this sort of use case, but hasn't happened yet. I don't know if it's out of scope for this.

storage_service/locations/models/package.py
```diff
@@ -522,10 +522,43 @@ def extract_file(self, relative_path='', extract_path=None):
...
+            # To get FPR commands used for extraction in extractContents_v0.0:
+            # SELECT command
+            #   FROM fpr_fpcommand
+            #     WHERE uuid IN ('a03a3a6d-5695-4800-a16d-edd7143cc438',
+            #                    '8f41dc6f-05eb-46d4-9b22-0a0d74673510',
+            #                    'b8864e2c-87a0-4aec-a8cf-a747a073be28');
+            #
+            # TODOs/Questions:
+            # 1. Will the relative/subpath final argument to 7z work in the
+            #    same way that it does in unar?
```
Hwesta: This should be fixed before this can be merged, since this is the same function used to extract single files.

Hwesta (commented):
Good sleuthing to find the problem! Unfortunately this change doesn't work for `.tar.bz2` files (created with the parallel bzip option in AM) since 7z doesn't know how to extract them.  (`7z x -bd -y -o outputdir /path/to/aip/aipname.tar.bz2` gives me the error `Error: Incorrect command line`)  Handling multiple package formats was the original motivation for using `unar` instead of another specific tool.

storage_service/locations/models/package.py
jrwdunham: I get the following error when I run `7z x -bd -y -o extract_path full_path.7z`:

    Command Line Error:
    Too short switch:
    -o

Hwesta (approved):
Looks good!  

Can you copy or move the explanation for the change into the commit message?  I usually check the commit history for an explanation of why a particular change was made, though in-code comments are also useful.

storage_service/locations/models/package.py
```diff
@@ -1024,6 +1047,45 @@ def start_reingest(self, pipeline, reingest_type, processing_config='default'):
             'reingest_uuid': reingest_uuid,
         }
 
+    def _get_compression(self, pointer_path):
```
Hwesta: `common.utils` has several other pointer file-related functions - this may be a better fit there.

jrwdunham:
@Hwesta can you take another look at this? I've changed things so that pbzip2-compressed AIPs should be re-ingest-ible too (i.e., with weird Mac resource fork stuff in them).

storage_service/locations/models/package.py
```diff
@@ -522,10 +522,30 @@ def extract_file(self, relative_path='', extract_path=None):
...
+            #     $ unar -force-overwrite -k hidden -o extract_path full_path
+            #
+            # preserves the .rsrc MACOSX files as ._-prefixed files, it does so
+            # differently than 7z/tar do: the resulting .-prefixed files have
+            # different sizes than those created via unar. This makes
+            # ``bag.validate`` choke.
+            if self.full_pointer_file_path:
+                compression = self._get_compression(self.full_pointer_file_path)
+            else:
+                compression = None  # no pointer file :. command will be unar
```
Hwesta, line 533: This is inside the `is_compressed` block, so it should always be compressed.

storage_service/locations/models/package.py
```diff
@@ -522,10 +522,30 @@ def extract_file(self, relative_path='', extract_path=None):
...
+            command = self._get_decompr_cmd(compression, extract_path, full_path)
             if relative_path:
                 command.append(relative_path)
```
Hwesta, line 536: Does this still work with all commands?

storage_service/locations/models/package.py
```diff
@@ -1024,6 +1050,45 @@ def start_reingest(self, pipeline, reingest_type, processing_config='default'):
...
+            else:
+                LOGGER.warning('Unable to determine reingested compression'
+                               ' algorithm, defaulting to bzip2.')
+                return self.COMPRESSION_7Z_BZIP
+        elif puid == 'x-fmt/268':  # Bzipped (probably tar)
+            return self.COMPRESSION_TAR_BZIP2
+        else:
+            LOGGER.warning('Unable to determine reingested file format,'
+                           ' defaulting recompression algorithm to %s.',
+                           compression)
```
Hwesta: Copy-paste error - compression is undefined.

Hwesta (commented):
I really like the changes the separate fetching the compression type & compression command into their own functions.

storage_service/locations/models/package.py
jrwdunham: Actually, this can happen. A travis CI fail alerted me to it. Sometimes `extract_file` is called on something that is not an AIP, something without a pointer file. So in those cases we set compression to None so that _get_decompr_cmd will return the elsewhere unar command.

storage_service/locations/models/package.py
jrwdunham: Yes, my investigation concluded  that 7z, unar and tar all allow for a final argument that specifies a relative path within the compressed directory. Please tell me if you think this is not the case.

storage_service/locations/models/package.py
jrwdunham: Yes. Thanks.

storage_service/locations/models/package.py
jrwdunham: Ok.

storage_service/locations/models/package.py
Hwesta: Great - wanted to make sure you'd looked into this.

storage_service/locations/models/package.py
Hwesta: Ah, reasonable.

Hwesta (approved):
New changes look good.

PR #163, opened 2016-12-06
Settings: Set ALLOWED_HOSTS in base.py

Hwesta:
`ALLOWED_HOSTS` was only specified in production.py, so when running with any other settings file it generated an error.


sevein:
:+1:

PR #165, opened 2016-12-10
Fixity: Fix typo

Hwesta:
Fixity returns 4 things - status, error list, error message, and timestamp.


sevein:
It'd be cool to have type annotations in our code and use github.com/python/mypy to detect these errors early, right? 👍

PR #166, opened 2016-12-14
Version display should work even if git not installed

jrwdunham:
Run `sudo apt-get purge git` in your VM and then be amazed that you can still see the SS version page!


sevein:
mindblown    <sup><sup>:+1:</sup></sup>

PR #167, opened 2016-12-15
Reingest: Don't keep original METS

Hwesta:
Only keep the updated METS file, don't keep the original METs file in submission documentation.

refs redmine 10669


jhsimpson:
I tested this and it works as expected.  code change seems pretty simple too.

:+1:

jrwdunham:
👍 works as expected, code clear, SS tests pass

PR #168, opened 2016-12-16
Reingest: Fix uncompressed AIP directories

Hwesta:
Get the local path of an uncompressed AIP as a directory, with a trailing /. Otherwise, we end up sending the directory nested inside a directory of the same name (e.g. aipname-uuid/aipname-uuid) which causes problems.

Requires artefactual/archivematica#546

refs redmine 10080


jhsimpson:
I tested this and it is working.  

:+1:

jhsimpson:
:+1: 

tested latest version with local filesystem, works.

PR #169, opened 2016-12-19
DSpace: stop creating .cache directory

Hwesta:
Remove the .cache directory that SWORD2 creates, because it is often the source of permissions errors.

Update the version of sword2 to include a fix that allows disabling the cache directory.

Edit: Updated sword2 version to install from the new pip release. It includes one extra commit from the install-from-git previously used, which is release related things and a license change (MIT->Apache)

These 2 commits may be squashed before merging.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -20,5 +20,5 @@ python-dateutil==2.4.2
 python-keystoneclient==1.8.1
 python-swiftclient==2.4.0
 requests>=2.3.0
-git+https://github.com/swordapp/python-client-sword2.git@0c241438f2ced13ec3364caa03edf28d5ed5c02c#egg=sword2
+sword2==0.2.0
 msgpack-python>=0.4.0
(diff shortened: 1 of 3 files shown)


jhsimpson:
This looks pretty straightforward.  The ability to stop creating the .cache directory doesn't require upgrading the version of sword2, I don't think?  The cache_deposit_receipts variable exists in both commits.

I am going to test this with DSpace, I assume if that is tested we don't really have to test with Locks-o-matic as well, which might take longer to set up a test bed for?

Hwesta:
Regardless of `cache_deposit_receipts`, sword2 still tries to create a `.cache` directory if specified.  Specifying `cache_dir=None` prevents the `.cache` directory from being created, but sword2 had a bug (https://github.com/swordapp/python-client-sword2/pull/6) that prevented it from working correctly until recently.

I updated LOCKSS because it also uses sword2, but I don't think we have access to a service to test against for it.

jrwdunham:
👍 from me: confirmed that these changes allow an AIP to be stored in our DSpace test instance.

PR #171, opened 2017-01-12
Dev/issue 10748 sword name slash

jhsimpson:
Fixes a bug in the FEDORA deposit Space, where files with a / in the name are now sanitized.


jhsimpson:
I ran the tests after @jrwdunham added one for this change, looks good. 👍

jrwdunham:
👍

PR #173, opened 2017-02-15
DSpace: archive format option

sevein:
Redmine: https://projects.artefactual.com/issues/10808

7z used to be the default. This commit makes ZIP the new default and updates
the DSpace model so the user can choose between the two.

### DIFF
--- a/storage_service/locations/migrations/0012_dspace_archive_format.py
+++ b/storage_service/locations/migrations/0012_dspace_archive_format.py
@@ -1,0 +1,19 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('locations', '0011_fixitylog_status'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='dspace',
+            name='archive_format',
+            field=models.CharField(default=b'ZIP', max_length=3, verbose_name=b'Archive format', choices=[(b'ZIP', b'ZIP'), (b'7Z', b'7z')]),
+        ),
+    ]
--- a/storage_service/locations/models/dspace.py
+++ b/storage_service/locations/models/dspace.py
@@ -50,6 +50,14 @@ class DSpace(models.Model):
             'This will override existing policies. '
             'Example: [{"action":"READ","groupId":"5","rpType":"TYPE_CUSTOM"}]'))
 
+    ARCHIVE_FORMAT_ZIP = 'ZIP'
+    ARCHIVE_FORMAT_7Z = '7Z'
+    ARCHIVE_FORMAT_CHOICES = (
+        (ARCHIVE_FORMAT_ZIP, 'ZIP'),
+        (ARCHIVE_FORMAT_7Z, '7z'),
+    )
+    archive_format = models.CharField(max_length=3, choices=ARCHIVE_FORMAT_CHOICES, default=ARCHIVE_FORMAT_ZIP, verbose_name=_l('Archive format'))
+
     sword_connection = None
 
     class Meta:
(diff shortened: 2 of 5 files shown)


storage_service/locations/migrations/0012_dspace_archive_format.py
```diff
@@ -0,0 +1,27 @@
...
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='dspace',
+            name='archive_format',
+            field=models.CharField(default=b'ZIP', max_length=3, verbose_name=b'Archive format', choices=[(b'ZIP', b'ZIP'), (b'7Z', b'7z')]),
+        ),
+
+        migrations.RunPython(data_migration),
```
sevein: I realize this is probably not necessary, right?

sevein: Should I use `preserve_default=True`?

storage_service/locations/models/dspace.py
```diff
@@ -121,6 +128,58 @@ def _get_metadata(self, input_path, aip_uuid):
...
+
+    def _archive_zip(self, src, dst):
+        """ Return the command that creates the ZIP archive file. """
+        if not dst.endswith('.zip'):
+            dst += '.zip'
+
+        return (dst, [
+            'zip',
+            '-r',
+            dst,  # Destination (without extension)
```
Hwesta: Should `dst` have the extension here? The comment says no but the current behaviour is yes.

storage_service/locations/models/dspace.py
```diff
@@ -121,6 +128,58 @@ def _get_metadata(self, input_path, aip_uuid):
         os.remove(mets_path)
         return kwargs
 
+    def _archive(self, src, dst):
+        """
+        Run the archiver in order to create the archive file based on the
+        preferred format. `dst` is the path of the archive file. The file
+        extension must not be included, instead this function will return
+        the final destination with the extension on it according to the
+        archive format preferred.
```
Hwesta: Tip: [PEP257](https://www.python.org/dev/peps/pep-0257/) for docstrings suggests having a commit-message-like docstring: single line summary, optionally followed by a multi-line explanation.

storage_service/locations/models/dspace.py
```diff
@@ -152,43 +211,11 @@ def _split_package(self, input_path):
...
-            '-mmt=on',  # Multithreaded
-            metadata_zip,  # Destination
-            metadata_dir,  # Source
-        ]
-        try:
-            subprocess.check_call(command)
-        except subprocess.CalledProcessError:
-            LOGGER.error('Could not compress %s', metadata_dir)
-            raise
+        metadata_zip = self._archive(metadata_dir, os.path.join(output_dir, 'metadata'))
```
Hwesta, line 231: Nice deduplication of code here.

storage_service/locations/tests/test_dspace.py
```diff
@@ -74,6 +76,15 @@ def test_split_package(self):
         split_paths = self.dspace_object._split_package(path)
         # Verify
         assert len(split_paths) == 2
+        assert os.path.join(FIXTURES_DIR, 'objects.zip') in split_paths
```
Hwesta, line 79: Maybe rename this to `test_split_package_zip` ?

storage_service/locations/migrations/0012_dspace_archive_format.py
Hwesta: I suspect it's not necessary, but you can verify that.

Hwesta (approved):
This is more configurable than I was expecting for this change, but looks like it works.  

We're looking at implementing `move_to_storage_service` (redmine 10862) for DSpace, which will require dealing with what happens if the package format changes between storing and retrieving.

storage_service/locations/models/dspace.py
sevein: ZIP doesn't really care but the comment was definitely wrong, will fix.

sevein:
I've addressed your comments, thank you!

Hwesta (approved):

sevein:
Merged. Accidentally used cherry-pick: a3bc8dba07b80c41af5794fa75c5fd0313ce60cf.

PR #175, opened 2017-03-01
osdeps: put OS dependencies data in repository

hakamine:
Add a directory osdeps/ with files containing information of OS
dependencies. To be used in deployment (ansible, etc.)
( Currently, files in osdeps/ directory are used by [PR104 of the ansible role](https://github.com/artefactual-labs/ansible-archivematica-src/pull/104) to be able to deploy qa/0.x on CentOS/RH 7 in addition to Ubuntu trusty )


Hwesta:
This looks good.  Same questions as on artefactual/archivematica#578 (Are we planning to replace the debian control files with this, or is this intended to be just for Ansible? How do we plan to keep the dependencies & versions in sync? The RedHat and CentOS json lists are identical - do we need both?)

hakamine:
Merged to qa/0.x and also cherry-picked tp stable/0.10.x

PR #176, opened 2017-03-10
Package: add transfer backlog reindex

sevein:
Redmine: https://projects.artefactual.com/issues/10967

It also deletes a couple of unused imports.

`Package.index_file_data_from_transfer_mets` has been updated to use `update_or_create` instead of `create` to ensure the files are only created once.

```
$ curl \
  -s \
  -XPOST \
  -H "Authorization: ApiKey test:f00b4r" \
    http://127.0.0.1:8000/api/v2/file/930f58eb-4722-4f49-b8a4-d110cfb1c6bd/reindex/
```

```json
{"message": "Files indexed: 2", "error": false}
```

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -446,6 +446,7 @@ class PackageResource(ModelResource):
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/send_callback/post_store%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('aip_store_callback_request'), name="aip_store_callback_request"),
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/contents%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view("manage_contents"), name="manage_contents"),
             url(r"^(?P<resource_name>%s)/metadata%s$" % (self._meta.resource_name, trailing_slash()), self.wrap_view("file_data"), name="file_data"),
+            url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/reindex%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('reindex_request'), name="reindex_request"),
             # Reingest
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/reingest%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('reingest_request'), name="reingest_request"),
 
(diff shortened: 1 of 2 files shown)


storage_service/locations/api/resources.py
```diff
@@ -31,7 +31,7 @@
 from locations.api.sword import views as sword_views
 
 from ..models import (Callback, CallbackError, Event, File, Package, Location, Space, Pipeline, StorageException)
-from ..forms import LocationForm, SpaceForm
```
jhsimpson: this is just an extra cleanup you are doing?

storage_service/locations/api/resources.py
```diff
@@ -839,6 +840,22 @@ def aip_store_callback_request(self, request, bundle, **kwargs):
         else:
             return http.HttpNoContent()
 
+    @_custom_endpoint(expected_methods=['post'])
+    def reindex_request(self, request, bundle, **kwargs):
```
jhsimpson, line 844: can you add a doc string explaining what this method is doing?

jhsimpson (approved):
:+1: 

I tested and this is working
root@questionable:/usr/lib/archivematica/storage-service# curl -s -XPOST -H "Authorization: ApiKey test:secretapikey" http://127.0.0.1:8000/api/v2/file/930f58eb-4722-4f49-b8a4-d110cfb1c6bd/reindex/
{"message": "Files indexed: 2", "error": false}

storage_service/locations/api/resources.py
```diff
@@ -839,6 +840,22 @@ def aip_store_callback_request(self, request, bundle, **kwargs):
...
+        package = bundle.obj
+        if package.package_type != Package.TRANSFER:
+            return http.HttpBadRequest(json.dumps({"error": True, "message": "This package is not a transfer."}), content_type='application/json')
+        if package.current_location.purpose != Location.BACKLOG:
+            return http.HttpBadRequest(json.dumps({"error": True, "message": "This package is not part of a transfer backlog location."}), content_type='application/json')
```
jrwdunham: Nitpicky point but packages are not really "part of" locations. Maybe error message should be "This package is not in transfer backlog."

jrwdunham: Also, is it possible for SS to know about a transfer and yet not have that transfer be in backlog? I tried to make a request that would trigger returning of the "This package is not part of a transfer backlog location." message but was not able to. I'm not saying this check should be removed; I'm just wondering if and how it's possible to trigger this response.

storage_service/locations/api/resources.py
```diff
@@ -839,6 +840,23 @@ def aip_store_callback_request(self, request, bundle, **kwargs):
...
+        try:
+            package.index_file_data_from_transfer_mets()  # Create File entries for every file in the transfer
+        except StorageException as e:
+            LOGGER.warning("Transfer METS data could not be read: %s", str(e))
+            return http.HttpApplicationError(json.dumps({"error": True, "message": "Transfer METS data could not be read."}), content_type='application/json')
```
Hwesta: Is this the only thing that can cause an exception?  Should the log message be made more generic?

jrwdunham (commented):
Confirmed that this works as expected and was able to get correct error responses that I expected. However, was not able to trigger "not part of a transfer" response.

storage_service/locations/api/resources.py
sevein: Are you suggestion to catch `Exception` or reword the error message?

storage_service/locations/api/resources.py
sevein: I honestly didn't even know if a different state was reachable by the user. I tried to be explicit about the state I needed the package to be in order to be indexed.

jrwdunham:
👍

storage_service/locations/api/resources.py
Hwesta: Currently, Transfers only go into Backlog locations, but that may not always be true, so I'd vote to keep the extra check.

storage_service/locations/api/resources.py
Hwesta: Both.  I normally recommend catching the smallest exception we know of, but since we don't know what exceptions to expect and getting a 500 error in response to an API call sucks, I think changing it to catch `Exception` and the error message be more generic ("An error occurred while reindexing the Transfer" maybe?) would be good.

It's not well documented, but you can pass `exc_info=True` to `LOGGER.warning` (or any of the other logging functions), and it will print the traceback (at least inside an except block), which is super useful.

storage_service/locations/api/resources.py
Hwesta: I left the "unused" form imports here because they're used in commented out validation code ([example](https://github.com/artefactual/archivematica-storage-service/blob/stable/0.9.x/storage_service/locations/api/resources.py#L241)) that I left as an example of how it could work. See [this]( https://github.com/artefactual/archivematica-storage-service/blob/stable/0.9.x/storage_service/locations/api/resources.py#L40-L45) for an explanation of the bug.

jhsimpson:
cherry-picked to stable/0.10.x as ebf6f6e4af69ca14676f733b74a64428c0c7164d

PR #170, opened 2017-01-05
package.py : Use parallel bag validation processes

tw4l:
Suggested change to increase speed of bag validation. Parallel validation processes would help immensely in speeding up, e.g., scanall Fixity app scans. It's worth double-checking but my understanding of the way bagit-python implements multithreading means that if the machine is unable to dedicate 4 cores to the task, it will simply allocate what is available.

processes=4 is arbitrary (instead of, say, 8) but seemed like a reasonable number. The ideal number would likely depend on local setup.

PR #177, opened 2017-03-28
package.py : Use parallel bag validation processes

tw4l:
Rebasing based on comments in [this PR](https://github.com/artefactual/archivematica-storage-service/pull/170)


tw4l:
Yes, I have signed a CLA

jhsimpson (approved):
This looks like a good change to me, the only comment I have about making this change configurable.  The default before this change is to run a single process, the default after is to run 4 processes.  

I am wondering about how best to make this something that can be configured by a user or at install time, on some machines you might want a different number.

The storage service doesn't have an easy place to add this kind of config option.  Some options include A config file that reads into an environment variable, read from settings.py at run time, or a command line parameter passed in via the gunicorn conf file, or a way to change the number in the admin section of the gui.

I think it is okay to merge this for the 0.11.0 release, and look at adding a config option in the future.

sevein:
Thank you! 49e8c47

Issue #187, opened 2017-05-10
Problem: Not using tox


jhsimpson:
we should be using tox.

jhsimpson:
:) merged now.

PR #188, opened 2017-05-10
Set up TravisCI and tox

sevein:
Two tox environments: `py27` (to run our test suite) and `flake8` (for linting).

Also, `py36`. We know it's not going to work but good to remember the why.

Added `.travis.yml` too and a commit fixing linting issues.

It solves #187.


sevein:
This is ready now.

jrwdunham:
👍

PR #192, opened 2017-05-13
#10662 GPG encrypted storage

jrwdunham:
Allows for creation of a GPG-encrypted space with AIP Storage and Transfer Backlog-purposed locations such that all packages placed in those locations are encrypted upon entry and decrypted on exit. Also includes an interface for managing (creating, importing, viewing, deleting) GPG keys.

See also

- https://github.com/artefactual/archivematica/pull/616
- https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/9
- https://github.com/artefactual-labs/ansible-archivematica-src/pull/109

PR #198, opened 2017-06-06
GPG-encrypted spaces

jrwdunham:
Implements the following features for GPG-encrypted spaces:

- GPG keys with passphrases cannot be imported
- GPG key associated to a space or needed to decrypt an AIP cannot be deleted
- changing the GPG key of an encrypted space and re-ingesting an AIP correctly re-encrypts that AIP with the new GPG key

Crucially, it also fixes a **bug** in SS:package.py where finish_reingest on an encrypted AIP was resulting in a doubly-compressed AIP.


jhsimpson:
This should replace #192

sevein (changes requested):
Please rebase!

sevein (changes requested):
@jrwdunham I think this needs to be rebase again (sorry!), #230 seems to be the PR that caused the conflict.

storage_service/common/gpgutils.py
```diff
@@ -0,0 +1,198 @@
...
+    expiration date, the key never expires.
+    Warning: generating this key for the first time can take several minutes.
+    At present, this default key is generated the first time that
+    ``get_gpg_key_list`` is called, which is typically when the "Encryption
+    keys" link is clicked under the "Administration" tab. Waiting on a long
+    request like this is not a very good user experience. It may be desirable
+    to do this in a separate thread and poll for completion, or install
+    rng-tools, or else at least warn the user that this will take a long time,
+    or use a smaller key size. The same considerations apply for
+    ``generate_gpg_key``. TODO/QUESTION ^.
```
jraddaoui: Could we generate the default key in the install/upgrade process?

jraddaoui (commented):
WOW, great feature! And huge work Joel, congrats!

It looks perfect to me, but I feel like I don't know the project enough to approve the changes, I think it would be great if @sevein could double check.

@jrwdunham, for the next time you're planing to introduce a feature as big as this one, please, ask for a reviewer earlier in the process.

Thanks.

storage_service/common/gpgutils.py
jrwdunham: Yes, that is an option. Although what I ended up doing was having Ansible install rng-tools: see https://github.com/artefactual-labs/ansible-archivematica-src/pull/109. With rng-tools installed, GPG key creation time is reduced to the point that this is no longer a usability concern. I think I will leave the code as is and modify the docstring accordingly.

jraddaoui: Great! Thanks!

jraddaoui (approved):
Thanks @jrwdunham! Recent changes look good to me and I think @jhsimpson approved this PR in the Slack channel.

PR #200, opened 2017-06-09
pipeline local FS: add rsync daemon support

hakamine:
Refs. #11244

### DIFF
--- a/storage_service/locations/models/pipeline_local.py
+++ b/storage_service/locations/models/pipeline_local.py
@@ -36,6 +36,11 @@ class PipelineLocalFS(models.Model):
         help_text=_l("Name or IP of the remote machine."))
     # Space.path is the path on the remote machine
 
+    assume_rsync_daemon = models.BooleanField(default=False,
+        verbose_name=_l("Assume remote host serving files with rsync daemon"),
+        help_text=_l("If checked, will use rsync daemon-style commands instead of the default rsync with remote shell transport"))
+    rsync_password = models.CharField(max_length=64, blank=True, default="", help_text="RSYNC_PASSWORD value (rsync daemon)")
+
     class Meta:
         verbose_name = _l("Pipeline Local FS")
         app_label = 'locations'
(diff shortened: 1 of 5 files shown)


storage_service/locations/models/space.py
```diff
@@ -461,7 +463,11 @@ def move_rsync(self, source, destination, try_mv_local=False):
         command = ['rsync', '-t', '-O', '--protect-args', '-vv', '--chmod=ugo+rw', '-r', source, destination]
         LOGGER.info("rsync command: %s", command)
 
-        p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
+        if assume_rsync_daemon:
+            p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, env={'RSYNC_PASSWORD': rsync_password})
+        else:
+            p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
sevein: A better way to write this could be:

```python
kwargs = {'stdout': subprocess.PIPE, 'stderr': subprocess.STDOUT}
if assume_rsync_daemon:
    kwargs['env'] = {'RSYNC_PASSWORD': rsync_password}
p = subprocess.Popen(command, **kwargs)
```

I almost prefer this one because it's shorter and more readable:

```python
env = {'RSYNC_PASSWORD': rsync_password} if assume_rsync_daemon else None
p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT, env=env)
```

hakamine: Thanks! fixed this and also other parts to reduce repetition.

hakamine:
Tested and seems to be working fine (can browse transfer source locations served with rsync daemon) Tested the migration from a stable/0.10.x database.

storage_service/locations/models/pipeline_local.py
```diff
@@ -36,6 +36,11 @@ class PipelineLocalFS(models.Model):
         help_text=_l("Name or IP of the remote machine."))
     # Space.path is the path on the remote machine
 
+    assume_rsync_daemon = models.BooleanField(default=False,
+        verbose_name="Assume remote host serving files with rsync daemon",
+        help_text="If checked, will use rsync daemon-style commands instead of the default rsync with remote shell transport")
+    rsync_password = models.CharField(max_length=64, blank=True, default="", help_text="RSYNC_PASSWORD value (rsync daemon)")
```
sevein: The values of `verbose_name` and `help_text` should be translation strings, e.g.:

```python
verbose_name=_l("Assume remote host serving files with rsync daemon")  # This is 
```

sevein (changes requested):
I haven't tested yet but I've seen you're not using translation strings - it's an easy fix.

storage_service/locations/models/pipeline_local.py
hakamine: Thank you. Fixed!

storage_service/locations/models/space.py
```diff
@@ -587,23 +591,29 @@ def browse_rsync(self, path, ssh_key=None):
...
-            '--exclude', '.*',  # Ignore hidden files
-            '--rsh', 'ssh -i ' + ssh_key,  # Specify identify file
-            path]
+        # Form command string used to get entries
+        command = ['rsync',
+                   '--protect-args',
+                   '--list-only',
+                   '--exclude', '.*',  # Ignore hidden files
+                   ]
+        if not(assume_rsync_daemon):
```
sevein: `not` is a keyword, not a function.
I think it's better to type it as: `if not assume_rsync_daemon:`

sevein (changes requested):
Small comment!

storage_service/locations/models/space.py
```diff
@@ -587,23 +591,29 @@ def browse_rsync(self, path, ssh_key=None):
...
+            # Specify identity file (TODO: do we really need to add this?)
```
sevein: I'd assume so, it's better to be explicit for those cases when the key is not somewhere ssh expects?

sevein (approved):

hakamine:
Thank you, I fixed to address the comments. Tested and working good.

Issue #191, opened 2017-05-11
Problem: 7zip and unar deps are overlapping


sevein:
We use unar once in the code, lsar is used twice.

Can we achieve the same with 7zip?
Best to keep the number of deps small, can we use 7zip where unar is used today?

Also, unar is not available in Alpine Linux our fav dist when we build containers.

sevein:
Also, [p7zip-full](https://github.com/travis-ci/apt-package-whitelist/blob/master/ubuntu-trusty#L3824) is whitelisted by TravisCI but not unar. That forces us to create a VM-based Travis environment instead of a container-based build environment, which is slower.

PR #204, opened 2017-06-18
osdeps: add p7zip

sevein:
SS uses `7z` in a few places but it's not listed as a dependency in `debian/control`. I've added it to `Ubuntu-14.json`, that was easy. But I don't know what is the equivalent in CentOS - this may help but I'm not sure: https://gist.github.com/P7h/9fcccc54596ad05764128dec6f6cf78d.

Help, please! CC @scollazo @hakamine @jhsimpson 

Related: #191.

### DIFF
--- a/osdeps/CentOS-7.json
+++ b/osdeps/CentOS-7.json
@@ -21,7 +21,8 @@
    { "name": "zlib-devel", "state": "latest"},
    { "name": "libffi-devel", "state": "latest"},
    { "name": "openssl-devel", "state": "latest"},
-   { "name": "gcc", "state": "latest"}
+   { "name": "gcc", "state": "latest"},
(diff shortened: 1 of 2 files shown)


hakamine:
For CentOS/RH the required packages are probably `p7zip` and `p7zip-plugins` ? These are included in the MCPClient dependencies: 
 https://github.com/artefactual/archivematica/blob/stable/1.6.x/src/MCPClient/osdeps/CentOS-7.json#L50-L51

sevein:
@hakamine thank you! So I can just add these two entries here and we're ready to go?

hakamine:
Yes, I think just copying these entries to the SS osdeps should be good

sevein:
@hakamine ok done, please take a look!

hakamine (approved):
Looks good!

Issue #195, opened 2017-05-30
Problem: there are missing migrations, refs #174


sevein:
In #174 we localized a bunch of field properties in our models but we did not include the corresponding migration.

sevein:
Fixed in #206.

sromkey:
@sevein Does this need to be QA'd?

sevein:
Nope! In a different PR we added a check in Travis CI so it's not going to happen again!

PR #206, opened 2017-06-28
#195: Added migration for locations app i18n changes

cole:
@sevein I finally got everything sorted out to make this change. Hope this is the right target branch?


sevein (approved):

sevein:
Fantastic! I had to cherry-pick (22ab063) because (long story:) your branch didn't have the most recent commit which shouldn't be a problem but we can't use the "Rebase and merge" button in this repo because we're using GH as a mirror. Thank you!

cole:
Cool cool, glad it worked out!

Issue #211, opened 2017-07-07
Add default location setting (merge from Jisc fork)


helenst:
Merge work from https://github.com/JiscRDSS/archivematica-storage-service/pull/8 and https://github.com/JiscRDSS/archivematica-storage-service/pull/8/ into core.

This provides a default location setting in the storage service.

Related AM issue: https://github.com/artefactual/archivematica/issues/669

lower29:
I have a dev branch based on `qa/0.x` ready to go for this change, but I don't have permission to push.

The branch includes the changes from https://github.com/JiscRDSS/archivematica-storage-service/pull/8 and https://github.com/JiscRDSS/archivematica-storage-service/pull/11 (which was a fix to 8) as well as a minor fix to a typo I spotted when I was reviewing the commits. Since the change for 11 is actually a fix for 8, I've squashed the three commits into one.

sevein:
Richard, you can fork the repo via GitHub, push the branch to your own remote and then submit the pull request.

lower29:
Yeah, Justin has suggested I use the JiscRDSS fork to prepare the branch.

sevein:
Fixed in #212.

sallain:
I've only tested this with a default AIP storage location, but it seems to be working well - extensive testing on our servers with "Store AIP location" set to "Default location" hasn't resulted in any errors, including when the default location is changed in the Storage Service.

PR #212, opened 2017-07-11
Add default location setting (merge from JiscRDSS fork)

lower29:
This merges changes from https://github.com/JiscRDSS/archivematica-storage-service/pull/8 and https://github.com/JiscRDSS/archivematica-storage-service/pull/11 into the Artefactual fork for issue #211.

Code is unchanged from @sevein's original commits - the only thing I've done is combine the two PRs and fix a minor typo in the location template and a flake8 whitespace error in `storage_service/locations/models/location.py`.

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -5,6 +5,7 @@
 import json
 import logging
 import os
+import re
 import shutil
 import urllib
 
@@ -12,6 +13,8 @@ import urllib
 from django.conf import settings
 from django.conf.urls import url
 from django.core.exceptions import ObjectDoesNotExist, MultipleObjectsReturned
+from django.core.urlresolvers import reverse
+from django.http import HttpResponseRedirect
 from django.forms.models import model_to_dict
 from django.utils.translation import ugettext as _
 
@@ -28,6 +31,7 @@ from tastypie.validation import CleanedDataFormValidation
 from tastypie.utils import trailing_slash
 
 # This project, alphabetical
+from administration.models import Settings
 from common import utils
 from locations.api.sword import views as sword_views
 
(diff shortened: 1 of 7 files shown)


helenst:
I've rebased this against the latest qa/1.x

storage_service/locations/api/resources.py
```diff
@@ -272,6 +276,26 @@ def get_objects(self, space, path):
         message = _('This method should be accessed via a versioned subclass')
         raise NotImplementedError(message)
 
+    def default(self, request, **kwargs):
```
jrwdunham, line 279: Is there a reason that you can not using the `_custom_endpoint` decorator? If not, then maybe you should for module-internal consistency. If yes, could you add a brief comment indicating why?

storage_service/locations/api/resources.py
```diff
@@ -465,6 +491,34 @@ def dehydrate_misc_attributes(self, bundle):
         # Serialize JSONField as dict, not as repr of a dict
         return bundle.obj.misc_attributes
 
+    def hydrate_current_location(self, bundle):
```
jrwdunham, line 498: I think this should have a docstring to reduce the amount of tastypie knowledge that is required to understand this functionality. Also, I think the `LOGGER.debug` calls should be to `.info` and `.warning` in production, no?

storage_service/locations/models/location.py
```diff
@@ -81,6 +83,12 @@ class Meta:
     objects = models.Manager()
     active = Enabled()
 
+    # Whether this location is the default location which is a global
+    # application setting stored in administration.models.Settings. See
+    # signal receivers unset_default_locaiton and set_default_location for
```
jrwdunham, line 88: typo

storage_service/locations/forms.py
```diff
@@ -163,9 +163,11 @@ class Meta:
 
 
 class LocationForm(forms.ModelForm):
+    default = forms.BooleanField(required=False)
```
jrwdunham: Can this have a help string, e.g., like the "Enabled" attribute has "True if space can be accessed"? 

Speaking of which, it seems confusing that it's not "True if location can be accessed" at https://github.com/artefactual/archivematica-storage-service/blob/qa/0.x/storage_service/locations/models/location.py#L75, no?

jrwdunham (changes requested):
I've played around with this and it works as I expect. I have a couple of change requests related to user and developer documentation. But in terms of functionality it looks good to me.

storage_service/locations/forms.py
sevein: True but that'd need a new migration, right?

sevein:
Ok, I've made the changes suggested!

helenst (commented):
I'm happy with this, nothing to add beyond what @jrwdunham has already pointed out.

storage_service/locations/forms.py
```diff
@@ -163,9 +163,11 @@ class Meta:
 
 
 class LocationForm(forms.ModelForm):
+    default = forms.BooleanField(required=False, help=_l("Set as global default location for its purpose"))
```
helenst: This should be `label` rather than `help`

helenst (changes requested):

sevein:
I'll take a look when I'm back on Tuesday, unless you want to do it.

helenst:
@sevein no worries, I will sort it out! :)

helenst (approved):

Issue #209, opened 2017-07-07
Add create_user command (merge from Jisc fork)


helenst:
Merge work from https://github.com/JiscRDSS/archivematica-storage-service/pull/2

This adds a create_user management command to the storage service to help with automated deployment.

Related AM issue: https://github.com/artefactual/archivematica/issues/665

lower29:
I have a dev branch based on `qa/0.x` ready to go for this. https://github.com/JiscRDSS/archivematica-storage-service/pull/2 is all new code and just a single commit so is straightforward to merge.

My only query is whether there is additional work required (e.g. to tests or ansible) that I'm not aware of.

sevein:
Thank you guys!

PR #213, opened 2017-07-11
Add create_user command with api key support (merge from JiscRDSS fork)

lower29:
This merges changes from https://github.com/JiscRDSS/archivematica-storage-service/pull/2 into the upstream Artefactual repo for issue #209.

The code changes remain identical to those in @sevein's original commit, except that they've been rebased against the latest `qa/0.x` branch.

### DIFF
--- a/storage_service/common/management/commands/create_user.py
+++ b/storage_service/common/management/commands/create_user.py
@@ -1,0 +1,40 @@
+from __future__ import print_function
+from __future__ import unicode_literals
+
+from django.contrib.auth import get_user_model
+from django.core.management.base import BaseCommand, CommandError
+
+from tastypie.models import ApiKey
+
+
+class Command(BaseCommand):
+    help = 'Create user with all the arguments specified. ' \
+        'If the user already exists its attributes will be updated.'
+    args = ('username', 'email', 'password', 'api_key')
+
+    def add_arguments(self, parser):
+        parser.add_argument('--username', required=True)
+        parser.add_argument('--email', required=True)
+        parser.add_argument('--password', required=True)
+        parser.add_argument('--api-key', required=True)
+
+    def handle(self, *args, **options):
+        UserModel = get_user_model()
+        user = None
+        try:
+            user = UserModel._default_manager.using('default').get(**{UserModel.USERNAME_FIELD: options['username']})
+            print('User found.')
+        except UserModel.DoesNotExist:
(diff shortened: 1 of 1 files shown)


storage_service/common/management/commands/create_user.py
```diff
@@ -0,0 +1,40 @@
...
+        parser.add_argument('--api-key', required=True)
+
+    def handle(self, *args, **options):
+        UserModel = get_user_model()
+        user = None
+        try:
+            user = UserModel._default_manager.get(**{UserModel.USERNAME_FIELD: options['username']})
+            print('User found.')
+        except UserModel.DoesNotExist:
+            user = UserModel._default_manager.create_superuser(options['username'], options['email'], options['password'])
```
helenst: I removed the `using('default')` part of this as it was causing `create_superuser` to fail (it can't be called on a `Queryset`, only a `Manager`). I guess the Jisc work never triggered this (perhaps because we just call this on the test user, and that already exists?). As far as I can tell it's fine to remove that, as it just tells Django to use the default database, and it will do that anyway. I also removed that part from the `get` line too.

storage_service/common/management/commands/create_user.py
```diff
@@ -0,0 +1,40 @@
...
+class Command(BaseCommand):
+    help = 'Create user with all the arguments specified. ' \
+        'If the user already exists its attributes will be updated.'
+    args = ('username', 'email', 'password', 'api_key')
+
+    def add_arguments(self, parser):
+        parser.add_argument('--username', required=True)
+        parser.add_argument('--email', required=True)
+        parser.add_argument('--password', required=True)
+        parser.add_argument('--api-key', required=True)
```
helenst, line 19: I wonder if we should have a `--superuser` flag rather than always creating a superuser.

lower29: I've looked at making this change; I want to test before pushing but I'm having problems getting my development environment to work. Hopefully will get this sorted out today.

helenst: Awesome! That will need consideration when merging back into jisc, since the command will no longer create superuser by default. Which I think is the right thing to do.

helenst:
This looks good to me - I've tested it out and it works. :+1:

storage_service/common/management/commands/create_user.py
lower29: Good point. The change to Jisc will be minimal, however, so I think this is okay.

PR #203, opened 2017-06-16
Package: move pointer file removal

qubot:
Per testing for CVA Archivematica dev site (cvan105) discussed in https://projects.artefactual.com/issues/11205

Looked at with Justin, related to Kelly's Re-ingest test (Uncompressed AIP created, then Re-ingest to compressed as 7zip), getting errors not finding compression pointer file (due to it being removed in all conditions). Making this change, re-ingest Kelly did the other day now worked.

I note (and based the name off) branch dev/issue-8893-uncompressed-aip-reingest by mcantelon, seems to be some divergence there that might be worth looking into along with this change?

- David H.

PR #214, opened 2017-07-13
Re-ingest: don't delete pointer file, refs #11356

sevein:
In #203, a change was introduced in finish_reingest() to stop deleting the
pointer file if the AIP was uncompressed (see #11205 for more details).
The change was made so the file was instead deleted when the AIP was
compressed which is causing a different issue as that file is used later in
the same function so this commit reverts that part of #203.

This should be included in `qa/0.x` and `stable/0.10.x`.


jhsimpson (approved):
This change makes sense to me.  In #203 - the fix was only tested for aips that were originally uncompressed, this fix works for aips whether they were originally compressed or not.

sevein:
stable/0.10.x bd2cc48bbfb71b11f8d6d346ef6325a5a4268e28

PR #215, opened 2017-07-14
Add pipeline UUID to deletion request emails

jraddaoui:
https://projects.artefactual.com/issues/10655


sevein (approved):
It looks good to me. My only comment is about the line breaks you've removed. Do you think it looks better without them? If it wasn't raised as an issue I don't see any reason to change them.

jraddaoui:
It's just my personal preference, but you're right, I'll change it back ;)

PR #216, opened 2017-07-19
Getting checksum for Arkivum from pointer file

jrwdunham:
Fixes RM issue 11138 where the SS was attempting to calculate a checksum for a compressed AIP in the staging path of an Arkivum space after the AIP was moved to the Arkivum AIP Storage location, which resulted in a `No such file or directory` `IOError`. The new behaviour is to retrieve the size, checksum and checksum algorithm of the compressed AIP from its pointer file instead of looking for it on disk.

### DIFF
--- a/storage_service/locations/tests/test_arkivum.py
+++ b/storage_service/locations/tests/test_arkivum.py
@@ -12,6 +12,11 @@ FIXTURES_DIR = os.path.abspath(os.path.join(THIS_DIR, '..', 'fixtures'))
 ARKIVUM_DIR = os.path.abspath(os.path.join(FIXTURES_DIR, 'arkivum'))
 
 
+def get_pkg_uuid_path(package_uuid):
+    tmp = package_uuid.replace('-', '')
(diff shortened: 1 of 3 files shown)


storage_service/locations/tests/test_arkivum.py
```diff
@@ -12,6 +12,11 @@
 ARKIVUM_DIR = os.path.abspath(os.path.join(FIXTURES_DIR, 'arkivum'))
 
 
+def get_pkg_uuid_path(package_uuid):
```
sevein, line 15: Would you mind adding a comment here for the reader with an example? E.g.:
`b9a07040-4f08-4a9c-b88a-5cc325e17e7e => b9a0/7040/4f08/4a9c/b88a/5cc3/25e1/7e7e`

sevein (approved):
👍

storage_service/locations/tests/test_arkivum.py
jrwdunham: sure

sevein (approved):

PR #223, opened 2017-08-04
move db dependencies out of base requirements

### LINKED ISSUE #222: Problem:  I can't run tests without installing database packages
This recent commit https://github.com/artefactual/archivematica-storage-service/pull/217/commits/215cf1870cc4b4d3888a4d4f70092286c584b2f3

brought in support for mysql and postgresql as db backends, instead of just sqlite3.  

When running tests on this repo, sqlite3 is still used, and so it should not be necessary to install mysqlclient and other db related dependencies in the test environment.

This is important for running automated tests (travis) as well as for  testing locally (only source code checked out, not testing on a full deployed archivematica server).

jhsimpson:
fixes #222

It should not be necessary to install the database pip packages
such as mysqlclient and psycopg2 in a test environment.

This commit duplicates some deps, in local.txt and production.txt
but it allows test environments to exclude these packages.

This helps reduce the size of the test environment and the time it
takes to run tests, since some of these pip packages require
operating system packages to be installed as well.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -8,7 +8,6 @@ bagit==1.5.4
 brotli==0.5.2  # Better compression library for WhiteNoise
 defusedxml==0.5.0
 Django>=1.8,<1.9
-dj-database-url==0.4.2
 django-annoying==0.10.3
 django-braces==1.11.0
 django-extensions==1.7.9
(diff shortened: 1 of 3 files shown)


sevein (approved):
This wouldn't be a good idea if we had integration tests hitting the database, but maybe we should.
A different way to think about it, I like the idea of merging base, prod and test requirements because I think we should be building artifacts that we can test, e.g. we build our Docker image, then we run the tests in it, then we deploy... our Dockerfile installs production.txt and test.txt.

Still, I think this is 👍  :)

PR #224, opened 2017-08-04
Updates to package.extract_file

### LINKED ISSUE #221: Problem: there is no error raised when extract_file() fails to extract a file
Scenario:  Trying to get a single file from a 7zipped AIP.

When package.extract_file is called, with a relative_path argument (i.e., extract only one file), and the package was compressed with 7zip, then 7z is used to extract the a file from within the package.

If the file doesn't exist (there is no file inside the package at 'relative_path') then 7z exits, with a message that says:

```
No files to process

Files: 0
Size:       0
```

but it returns exit code 0.

The problem appears to be here: 
https://github.com/artefactual/archivematica-storage-service/blob/stable/0.10.x/storage_service/locations/models/package.py#L538-L541

This code assumes exit code 0 means the file was extracted.  

This was uncovered when trying to extract a manifest-sha512.txt file from an AIP that did not have that file, it has a manifest-sha256.txt.

jhsimpson:
fixes #221

This commit adds 4 tests and fixes to make the tests pass.

There are 2 specific bugs that this commit fixes:

1) when calling extract_file with a relative_path (i.e request
to extract a single file, not an aip)  on an
uncompressed aip, a single file is now returned.  Previously the
entire aip was returned

2) when calling extract_file with a relative_path and the requested
file does not exist, an error is raised.

Happy for comments/updates on these tests, I'm not sure they are optimal.


helenst (approved):
This looks good to me. I think the tests are fine, and cover the cases well. I added one more.

sevein:
@helenst, we lost your commit because it was pushed to github.com but this repo is a mirror from our internal git server. Justin is going to fix that soon. In the meantime, can you share a gist or similar with the contents of your commit, please? Thank you.

helenst:
@sevein I've added a gist here: https://gist.github.com/helenst/6f8d3d0871980e1bf84bc29c6d53e3a6 Also in the commit was a small whitespace correction (last line) to fix the flake8 checks on travis.

sevein:
Thanks @helenst, I've incorporated your test to his commit.

storage_service/locations/models/package.py
```diff
@@ -553,17 +554,21 @@ def extract_file(self, relative_path='', extract_path=None):
...
+                src = os.path.join(head, relative_path)
+                os.mkdir(os.path.join(extract_path, basename))
+                shutil.copy(src, output_path)
+            else:
+                src = full_path
+                shutil.copytree(full_path, output_path)
 
-        if not relative_path:
-            self.local_path_location = ss_internal
-            self.local_path = output_path
```
jrwdunham, line 566: @jhsimpson (@sevein @helenst) the deletion of these three lines seems to have introduced a bug, which was erroneously identified with [Redmine 11419](https://projects.artefactual.com/issues/11419#note-5). Because `self.local_path_location` is no longer being set after extraction, all re-ingest is failing. The chain of events is quite convoluted, but essentially all of the paths in `reingest_files` in `start_reingest()` end up being wonky, i.e., they end up looking like the impossible `var/archivematica/sharedDirectory/www/AIPsStore/var/archivematica/storage_service/...`

Long story short, I'm wondering if there is a reason that lines 564-566 were removed. If not, I'd like to revert this part of this PR.

jrwdunham: See https://github.com/artefactual/archivematica-storage-service/pull/235

PR #227, opened 2017-08-09
Pipeline local filesystem: x permission on folders

### LINKED ISSUE #226: Pipeline local filesystem: rsynced folders don't allow x permission for group
I have this issue:
I use a pipeline local filesystem to link the storage with the dashboard. In the dashboard, the services are run with the `archivematica` user.

For some reasons, when the storage sends files to the dashboard, it uses another user to ssh in the dashboard, called `archivematica-storage`.

Both `archivematica` and `archivematica-storage` are in the group `archivematica`, so the folders created by the storage on the dashboard look like this:

```bash
drwxrw-rw-. 1 archivematica-storage archivematica 0 Aug  9 15:26 archive
```

As you can see, groups can't execute the folder (so they can't see what's in). So, when the dashboard wants to use it, via the `archivematica` user, it doesn't work.

For me, it makes sense that the created folders have the execute permission as long as they already have read and write, thus I suggest the following PR: #227

remileduc:
- add the "execute" permission on folders so users can see what is inside.

closes #226

### DIFF
--- a/storage_service/locations/models/space.py
+++ b/storage_service/locations/models/space.py
@@ -445,7 +445,7 @@ class Space(models.Model):
 
         # Rsync file over
         # TODO Do this asyncronously, with restarting failed attempts
-        command = ['rsync', '-t', '-O', '--protect-args', '-vv', '--chmod=ugo+rw,Da+x', '--perms', '-r', source, destination]
+        command = ['rsync', '-t', '-O', '--protect-args', '-vv', '--chmod=F660,D770', '-r', source, destination]
         LOGGER.info("rsync command: %s", command)
 
         p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
@@ -513,7 +513,7 @@ class Space(models.Model):
         for directory in directories:
             path = os.path.join(os.path.dirname(directory), '')
             path = "{}@{}:{}".format(user, host, utils.coerce_str(path))
-            cmd = ['rsync', '-vv', '--protect-args', '--chmod=ugo+rwx', '--recursive', temp_dir, path]
+            cmd = ['rsync', '-vv', '--protect-args', '--chmod=ug=rwx,o=rx', '--recursive', temp_dir, path]
             LOGGER.info("rsync path creation command: %s", cmd)
             try:
                 subprocess.check_call(cmd)
--- a/storage_service/locations/models/space.py
+++ b/storage_service/locations/models/space.py
@@ -445,7 +445,7 @@ class Space(models.Model):
 
         # Rsync file over
         # TODO Do this asyncronously, with restarting failed attempts
-        command = ['rsync', '-t', '-O', '--protect-args', '-vv', '--chmod=ugo+rw', '-r', source, destination]
+        command = ['rsync', '-t', '-O', '--protect-args', '-vv', '--chmod=ugo+rw,Da+x', '--perms', '-r', source, destination]
         LOGGER.info("rsync command: %s", command)
 
         p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
@@ -513,7 +513,7 @@ class Space(models.Model):
         for directory in directories:
             path = os.path.join(os.path.dirname(directory), '')
             path = "{}@{}:{}".format(user, host, utils.coerce_str(path))
-            cmd = ['rsync', '-vv', '--protect-args', '--chmod=ugo+rw', '--recursive', temp_dir, path]
+            cmd = ['rsync', '-vv', '--protect-args', '--chmod=ugo+rwx', '--recursive', temp_dir, path]
             LOGGER.info("rsync path creation command: %s", cmd)
             try:
                 subprocess.check_call(cmd)


storage_service/locations/models/space.py
```diff
@@ -445,7 +445,7 @@ def move_rsync(self, source, destination, try_mv_local=False):
 
         # Rsync file over
         # TODO Do this asyncronously, with restarting failed attempts
-        command = ['rsync', '-t', '-O', '--protect-args', '-vv', '--chmod=ugo+rw', '-r', source, destination]
+        command = ['rsync', '-t', '-O', '--protect-args', '-vv', '--chmod=ugo+rw,Da+x', '--perms', '-r', source, destination]
```
remileduc: note that here, the `x` permission is added to folders only

storage_service/locations/models/space.py
```diff
@@ -513,7 +513,7 @@ def create_rsync_directory(self, destination_path, user, host):
         for directory in directories:
             path = os.path.join(os.path.dirname(directory), '')
             path = "{}@{}:{}".format(user, host, utils.coerce_str(path))
-            cmd = ['rsync', '-vv', '--protect-args', '--chmod=ugo+rw', '--recursive', temp_dir, path]
+            cmd = ['rsync', '-vv', '--protect-args', '--chmod=ugo+rwx', '--recursive', temp_dir, path]
```
remileduc: Here, we KNOW we write a folder, so we can put the `x` permission on it.

As far as I'm concerned, this whole function may be replaced by the following code (but maybe I missed some features here):

```python
    def create_rsync_directory(self, destination_path, user, host):
        # we know we have a paswordless SSH connection
        path = os.path.dirname(destination_path)  # thus we keep the previous behavior where we need a trailing slash or the folder is not created
        cmd = ['ssh', '{}@{}'.format(user, host), '"mkdir -p --mode=a+rwx " + path']
        
        LOGGER.info("path creation command: %s", cmd)
        try:
            subprocess.check_call(cmd)
        except subprocess.CalledProcessError as e:
            LOGGER.warning("rsync path creation failed: %s", e)
            raise
```

jhsimpson: I don't think you missed anything the function you suggest looks good.  Do you want to update your PR to include it?

storage_service/locations/models/space.py
jhsimpson: Thanks for the PR!  I am not familiar with that chmod syntax, is 'Da+x' the same as 'a+X'?

I think this is a good change.  
👍

sevein: @remileduc, what's the purpose of adding `--perms` when combined with `--chmod`?

I'm also wondering if we should use `--chmod=ug+rw,ug+x` instead, we don't need to give permissions to `others` right? Perhaps easier to read in octal mode: `--chmod=F660,D770`.

storage_service/locations/models/space.py
remileduc: (updated the code so we give a path to mkdir ^^)

Here, I assume there's a passwordless SSH configuration between the storage and the dashboard, as required by your documentation. However, from what I saw, only RSYNC is used to communicate between these 2 servers. Thus, somebody could have setup RSYNC to use another protocol (rsh? rsyncd?). I'm not sure it is possible, but using directly SSH like this would break it.
However, your documentation states that it needs SSH to be configured, this may be enough. What do you think?

I wait for your answer before editing the PR...

storage_service/locations/models/space.py
remileduc: @jhsimpson you have an explanation of what the `D` stands for here: https://explainshell.com/explain?cmd=rsync+-p+--chmod%3D%2Brwx+-e+%22ssh+-i+userserver.pem%22+--copy-unsafe-links+-rz+user%40server%3A~%2F+%2Fdb_backups%2Fwww%2F
It does the same as `(upper) X`, but I wasn't aware of it ^^

@sevein you're right, `--perms` would only be useful for already existing files, which shouldn't be the case here.

remileduc:
I've updated the permissions with @sevein comments.
Let me know if you want me to put the SSH version for the second method.

storage_service/locations/models/space.py
jrwdunham: @remileduc I like the idea of using ssh and `mkdir -p` to create the rsync directory. However, the third line of the snippet above has an error and will include the string `'path'` instead of the value of the `path` var. It should be something like `cmd = ['ssh', '{}@{}'.format(user, host), '"mkdir -p --mode=a+rwx {}"'.format(path)]`.

Also, in the snippet above `--mode=a+rwx` is being used while different permissions (`--chmod=ug=rwx,o=rx`) are being used in the current state of the PR.

Finally, are you sure that you want to call `path = os.path.dirname(destination_path)`? If `destination_path` has no trailing slash, then this will return the parent directory, which is not what you want, I think. If we want to always have a trailing slash we should continue to join on the empty string:

    >>> import os
    >>> os.path.dirname('a/b/c/')
    'a/b/c'
    >>> os.path.dirname('a/b/c')
    'a/b'
    >>> os.path.join('a/b/c/', '')
    'a/b/c/'
    >>> os.path.join('a/b/c', '')
    'a/b/c/'

remileduc: You're totally right! I did the snippet above very quickly to show the main idea...

Should I update the PR with the `mkdir` version?

jrwdunham: I think so. Your strategy is a lot cleaner than the current one which needlessly makes a separate rsync call for each ancestor directory.

jhsimpson (approved):

jrwdunham:
Replaced by https://github.com/artefactual/archivematica-storage-service/pull/281

PR #230, opened 2017-08-11
Configurable BagIt validation processes & warnings re Gunicorn gevent

jrwdunham:
Fixes artefactual/archivematica#708
    
Allows user to configure number of processes to use when calling BagIt-python's ``validate`` method. Using more than one when the Gunicorn worker class is 'gevent' causes BagIt validate to hang. This change warns the user/installer of Archivematica about that in two places.


sevein (approved):
I think that we should convert the output of `environ.get` into an `int` because the environment variable, if available, is always going to be a `str`. `int()` will throw a `ValueError` in some cases but I think that's ok.

Other than that, 👍

PR #235, opened 2017-09-17
Fixes re-ingest (issue #234)

### LINKED ISSUE #234: Problem: All re-ingest is failing
When attempting to re-ingest an AIP (using any re-ingest type) the Django error `Error re-ingesting package: An unknown error occurred` shows up. (Originally filed under [RM #11419 Multi-process bagit.validate breaks AIP re-ingest](https://projects.artefactual.com/issues/11419#note-5) but this is a distinct issue.)

Inspection of the SS logs shows Rsync failures when trying to move source files at paths like `var/archivematica/sharedDirectory/www/AIPsStore/var/archivematica/storage_service/tmp4q3zkU/test1-09dc6190-1909-4cb9-8134-3081739b1f12/data`. Such paths are impossible. The bad call is happening at `move_to_storage_service`, called from `package.py::start_reingest`.

Further investigation reveals that the call to `self.extract_file()` in `start_reingest` is failing to set `self.local_path_location` to the SS-internal location. The result of this is that the `current_location` var in `start_reingest` never gets set to `self.local_path_location`, which results in `relative_path` (cf. `relative_path = local_path.replace(current_location.full_path, '', 1).lstrip('/')`) failing to have the correct path prefix removed. Finally, the result of all that is that `reingest_files` ends up containing a bunch of nonsense paths that have `relative_path` prefixed to them.

The fix would seem to be restoring the following three lines at the end of `package.py::extract_file`.

    if not relative_path:
        self.local_path_location = ss_internal
        self.local_path = output_path

These lines were removed by https://github.com/artefactual/archivematica-storage-service/pull/224, although it is not clear to me if their removal is essential to the goal of that PR.

jrwdunham:
Fixes #234 by making sure that `Package().local_path_location` is set to
`ss_internal` in `extract_file`. Without this, `start_reingest` tries to
copy non-existent files.

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -570,6 +570,9 @@ class Package(models.Model):
 
             LOGGER.info('Copying from: %s to %s', src, output_path)
 
+        if not relative_path:
+            self.local_path_location = ss_internal
+            self.local_path = output_path
         return (output_path, extract_path)
 
     def compress_package(self, algorithm, extract_path=None):


sevein (approved):
@jrwdunham, this is working for me. I had a couple of issues in my local environment but they were unrelated.

PR #241, opened 2017-09-27
Mirror locations and AIP replicas

jrwdunham:
This PR brings AIP replication to the Storage Service. Users can now create locations with purpose `Replicator` (`RP`). A replicator location can be configured as a replicator location of any number of AIP Storage (`AS`) locations. When an AIP is stored (`Package.store_aip` is called), that AIP is now replicated to all of the replicator locations belonging to the AIP's `AS` location. The pointer files of both replica and the replicated AIPs are modified to document the replication that has occurred.

This PR uses [metsrw branch dev/issue-10662-write-pointer-files](https://github.com/artefactual-labs/mets-reader-writer/pull/27) which introduces a new way to create pointer files, including PREMIS elements. This is informed by [this draft Architectural Decision Record for unifying METS interactions in Archivematica](https://docs.google.com/a/artefactual.com/document/d/1mAOtwCxyj5i3uF3fufoBZXn2F0MYNYg-nfjUBFm6EwY/edit?usp=sharing).

# Warnings

1. This does not currently work with AIP re-ingest. That is, a re-ingested AIP will (probably) not be replicated in the cases where it should be. In addition, it will be necessary to implement logic such that when an AIP is re-ingested deletion requests are issued against its outdated replicas. See [Artefactual-internal "Estimate: AIP Encryption via Mirror Locations (and Replica Packages)"](https://docs.google.com/document/d/1PQqwgTHYEV4XXe_9et9CNTznSyb34Nk26rls-PHjaYs/edit)
2. More unit tests should be written for the new methods/functions in package.py. Related to this, these methods/functions should be refactored where possible to make them pure/side-effect-less so that unit tests creation can be facilitated.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -20,7 +20,7 @@ gunicorn==19.7.1
 jsonfield==2.0.1
 logutils==0.3.4.1
 lxml==3.7.3
-git+https://github.com/artefactual-labs/mets-reader-writer.git@dev/issue-10662-write-pointer-files
+metsrw==0.2.0
 ndg-httpsclient==0.4.2
 pyasn1==0.2.3
 python-gnupg==0.4.0
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -20,6 +20,7 @@ gunicorn==19.7.1
 jsonfield==2.0.1
 logutils==0.3.4.1
 lxml==3.7.3
+git+https://github.com/artefactual-labs/mets-reader-writer.git@dev/issue-10662-write-pointer-files
 ndg-httpsclient==0.4.2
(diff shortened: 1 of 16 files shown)


requirements/base.txt
```diff
@@ -20,6 +20,7 @@ gunicorn==19.7.1
 jsonfield==2.0.1
 logutils==0.3.4.1
 lxml==3.7.3
+git+https://github.com/artefactual-labs/mets-reader-writer.git@dev/issue-10662-write-pointer-files
```
sevein: Friendly reminder to update this before it's merged!

sevein (approved):

PR #244, opened 2017-10-10
Disable automatic lang selection

sevein:
This closes https://github.com/artefactual/archivematica/issues/723.
Related: https://github.com/artefactual/archivematica/pull/748.

### DIFF
--- a/storage_service/common/middleware.py
+++ b/storage_service/common/middleware.py
@@ -61,3 +61,21 @@ class CustomShibbolethRemoteUserMiddleware(ShibbolethRemoteUserMiddleware):
         entitlements = shib_meta['entitlement'].split(';')
         user.is_superuser = settings.SHIBBOLETH_ADMIN_ENTITLEMENT in entitlements
         user.save()
+
+
+class ForceDefaultLanguageMiddleware(object):
+    """
+    Ignore Accept-Language HTTP headers
+
+    This will force the I18N machinery to always choose settings.LANGUAGE_CODE
+    as the default initial language, unless another one is set via sessions or
+    cookies.
+
+    Should be installed *before* any middleware that checks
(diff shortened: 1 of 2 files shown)


jraddaoui (approved):
Not a big deal as I guess it's not being used anywhere else, but we could set the settings language in the request instead of removing it.

Thanks @sevein

sevein:
> Not a big deal as I guess it's not being used anywhere else, but we could set the settings language in the request instead of removing it.

I agree, it feels like a hack. Removing the header from the request is not ideal but it seems to be the only way that doesn't require us to bundle our own version of [LocaleMiddleware](https://github.com/django/django/blob/stable/1.8.x/django/middleware/locale.py#L29-L34).

Perhaps take a look at the link and see if you can find a better way. This is in any case a temporary solution that we're including in AM 1.7. But it'd be good to see if it has any unexpected impact.

sevein:
@helenst, I'm curious to hear your opinion :)

helenst:
@sevein so by doing it this way (rather than removing LocaleMiddleware) the user can still select their language preference explicitly in the admin, it just won't use the header value (i.e. the browser setting) - is that right? (I'm just running the ansible stuff so I can try it out). It seems like a reasonable way to deal with it.

helenst (approved):

sevein:
@helenst, that's right! The current state of translations in Spanish and French is so poor right now that we decided it doesn't make sense to offer it unless the user explicitly wants to choose it.

You may want to try https://github.com/artefactual-labs/am instead (Compose) - it's so much faster!

helenst:
@sevein Thanks, I'll give that a try!

sevein:
9954f12a92b5c408eb26c5e82c736ef68e173ed7

PR #247, opened 2017-10-13
 Enable re-login after Shibboleth logout

### LINKED ISSUE #238: Shibboleth user cannot login after the first logout
In testing the archivematica/1.x and archivematica-storage-service/0.x branches we encounter the following issue when a user authenticates with the external identity provider method Shibboleth:

1. The user has never connected with the username
2. The user connects to the dashboard\storage service
3. The user is directed to the Shibboleth Identity Provider
4. The user authenticates
5. The user is directed to the dashboard\storage service
6. A local account is created for the user with the correct roles
7. The user operated the UI
8. The user does a logout
9. The Shibboleth service provider confirms the end of the session
10. The user closes the browser.
11 . The user opens the browser
12. The user repeats 2, 3, 4 and 5 but then the expected 7 does not occur.... in stead the user is presented with the native login screen of the dashboard\service provider and cannot proceed.

helenst:
django-shibboleth-user sets a session flag to prevent immediate bounce back to logged in state after logout. This is useful in some setups but is unnecessary here, and it prevents logging back in again until cookies are cleared.

I've added a custom logout which unsets that session flag after the standard logout. This should enable users to log back in.

Fixes #238

See also artefactual/archivematica#756 (AM PR)

### DIFF
--- a/storage_service/storage_service/settings/base.py
+++ b/storage_service/storage_service/settings/base.py
@@ -251,7 +251,6 @@ LOGIN_REDIRECT_URL = '/'
 LOGIN_EXEMPT_URLS = (
     r'^api/',
     r'^admin/',
-    r'^logged-out',
     r'^Shibboleth.sso/',
     r'^login/',
 )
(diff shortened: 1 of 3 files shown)


sevein:
:+1:

PR #252, opened 2017-10-24
Travis: small optimizations

sevein:
- Cache pip dependencies
- Override `install` which was running `pip install -r requirements.txt` - not needed as that's managed internally by tox.

---

Build are significantly faster now:

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,8 +11,13 @@ addons:
       - p7zip-full
       - unar
 
+cache:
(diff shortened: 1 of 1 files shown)


jrwdunham (approved):

sevein:
Ops, wrong order. But this is the commit: efce4de78cd47c50270f45cff7a95db86c5e7d5c.

PR #256, opened 2017-11-04
Backward-compatible logging

sevein:
Optionally read logging configuration from `/etc/archivematica/storageService.logging.json`. See the included [install/README.md](https://github.com/artefactual/archivematica-storage-service/blob/dev/backward-compatible-logging/install/README.md) for more details. See also https://github.com/artefactual/archivematica/pull/782.

This closes https://github.com/artefactual/archivematica/issues/776.

### DIFF
--- a/install/logging.json
+++ b/install/logging.json
@@ -1,0 +1,83 @@
+{
+  "disable_existing_loggers": false,
+  "filters": {
+    "require_debug_false": {
+      "()": "django.utils.log.RequireDebugFalse"
+    }
+  },
+  "formatters": {
+    "detailed": {
+      "datefmt": "%Y-%m-%d %H:%M:%S",
+      "format": "%(levelname)-8s  %(asctime)s  %(name)s:%(module)s:%(funcName)s:%(lineno)d:  %(message)s"
+    },
+    "simple": {
+      "format": "%(levelname)-8s  %(name)s.%(funcName)s:  %(message)s"
+    }
+  },
+  "handlers": {
+    "console": {
+      "class": "logging.StreamHandler",
+      "formatter": "simple",
+      "level": "DEBUG"
+    },
+    "logfile": {
+      "backupCount": 5,
+      "class": "logging.handlers.RotatingFileHandler",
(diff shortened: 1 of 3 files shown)


storage_service/storage_service/settings/base.py
```diff
@@ -258,12 +261,16 @@ def get_env_variable(var_name):
...
-# the site admins on every HTTP 500 error when DEBUG=False.
 # See http://docs.djangoproject.com/en/dev/topics/logging for
 # more details on how to customize your logging configuration.
+
+# Configure logging manually
+LOGGING_CONFIG = None
+
+# Location of the logging configuration file that we're going to pass to
+# `logging.config.fileConfig` unless it doesn't exist.
+LOGGING_CONFIG_FILE = '/etc/archivematica/storageService.logging.json'
```
scollazo, line 273: Instead of hardcoding it, can we use an environment variable that points to the config file, and use /etc/archivematica/storageService.logging.json as the default value (in case the variable is not setted)?

install/logging.json
```diff
@@ -0,0 +1,83 @@
+{
```
scollazo, line 1: Renaming this file to storageService.logging.conf would make sense?

scollazo (changes requested):

scollazo:
I tested this in a centos environment, and logs continue to go into journalctl, not sure if I'm missing something.

sevein:
> I tested this in a centos environment, and logs continue to go into journalctl, not sure if I'm missing something.

@scollazo did you update the package so it copies `install/logging.json` into `/etc/archivematica/storageService.logging.json`?

storage_service/storage_service/settings/base.py
sevein: That's a good idea. Can we do this in 1.8? https://github.com/artefactual/archivematica/pull/782 has been merged already and it does the same thing.

install/logging.json
sevein: Done.

scollazo (approved):

Issue #266, opened 2017-11-18
Problem: pipeline self-registration mechanism is static


sevein:
When a new pipeline is registered, SS guesses the IP address by looking at the REMOTE_ADDR header of the HTTP request. This mechanism has worked fine for a long time because the pipeline was either running in the same host or in a separate host with a static IP address. In more dynamic environments IP addresses may change often and SS eventually becomes unable to reach the pipeline back, for example, for re-ingest.

In JiscRDSS [this issue was temporary solved](PIPELINE_REMOTE_NAME) by adding a new environment variable `PIPELINE_REMOTE_NAME` - this environment is defined in [rdss-archivematica](https://github.com/JiscRDSS/rdss-archivematica/pull/103). This enables us to use DNS names instead which works as long as the DNS server is up to date and the TCP port is always the same across al the replicas.

sevein:
Temporary the SS admin can manually update the remote name in the admin tab using a DNS name, e.g. in our Compose environment the user can use the name `archivematica-dashboard`.

![image](https://user-images.githubusercontent.com/606459/32983679-5772147e-cc4d-11e7-917d-6c355a58ce5e.png)

sevein:
Long-term solutions like the self registration pattern in service discovery should be considered. A simpler solution could be to allow the pipeline to define its own address. Instead of remote name we should we store the base URL, e.g. `https://archivematica-dashboard:8080`?

sromkey:
@sevein does the referenced PR close this issue?

sevein:
Yay.

Issue #236, opened 2017-09-22
Problem: pipelines are not always discoverable


sevein:
SS stores the `remote_name` of a pipeline when it's created for consequent API access, e.g. re-ingest.

```python
remote_name = models.CharField(
    max_length=256,
    default=None,
    null=True,
    blank=True,
    help_text="Host or IP address of the pipeline server for making API calls.")
```

This field is also editable from the web interface.

When a pipeline is created via the SS API, this field is populated after the `REMOTE_ADDR` header unless the client provides a value via the `remote_name` property.

The problem is that the dashboard doesn't allow users to provide a custom value so SS always fallbacks to the value found in the `REMOTE_ADDR` header which is problematic under some circumstances.

jrwdunham:
Assuming a URL is `scheme:[//[user[:password]@]host[:port]][/path][?query][#fragment]`, are you proposing adding a new attribute to `Pipeline` so that `Pipeline().remote_url` must be a valid `scheme:[//host[:port]]`?

Or are you thinking of modifying the logic around `Pipeline().remote_name` so we can accept both scheme-less and schemed URLs?

sevein:
> Or are you thinking of modifying the logic around Pipeline().remote_name so we can accept both scheme-less and schemed URLs?

I didn't think much about it but I like the way ^ sounds. It would mean that we can solve the problem without a database migration - is that ok?

jrwdunham:
Sounds good to me.

sevein:
I think that this issue needs to remain open until SS learns a new parameter that allows the user to indicate the URI of the pipeline in the create pipeline endpoint. This parameter should be provided by Archivematica when it performs the self-registration process.

sevein:
@sromkey, I think that we can close this issue.

SS v0.11 included https://github.com/artefactual/archivematica-storage-service/pull/267 which made the following changes:

- Parse `Pipeline.remote_name` as URLs (797d6a91c657a6a42296c3215b7ec634704a5e10).
- Create pipeline endpoint to honour `remote_name`(134e33b108d1304f68794de51389f94e4c34ddd1)
- Add `SS_INSECURE_SKIP_VERIFY` env so TLS cert verification can be disabled when the scheme found in the URI is `https` (`ecacbec7d066f072966fc15342417b3ac381b929`)

I think that these changes together already enabled the dashboard to provide a URI as the remote name. The remaining changes (provide a URI) need to happen in the AM repo.

PR #267, opened 2017-11-19
Better Pipeline.remote_name handling

sevein:
Read the commits for more details.

This pull request mitigates #266 (related to #236).

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -133,8 +133,8 @@ class PipelineResource(ModelResource):
         bundle = super(PipelineResource, self).obj_create(bundle, **kwargs)
         bundle.obj.enabled = not utils.get_setting('pipelines_disabled', False)
         create_default_locations = bundle.data.get('create_default_locations', False)
-        # Try to guess Pipeline's IP, if doing default setup
-        if create_default_locations:
+        # Try to guess Pipeline's IP if remote_name is undefined
+        if bundle.data.get('remote_name') is None:
             ip = bundle.request.META.get('REMOTE_ADDR') or None
             bundle.obj.remote_name = ip
         shared_path = bundle.data.get('shared_path', None)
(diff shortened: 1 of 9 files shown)


scollazo:
I think this will fail if the ss uses ssl with self-signed certificates.

sevein:
> I think this will fail if the ss uses ssl with self-signed certificates.

Ok, thank you @scollazo. I'll add a `INSECURE_SKIP_VERIFY` toggle.

sevein:
@scollazo see c8c5b5f70bb8c6b9e499ff880c11da2572913eda.

ross-spencer (commented):
I can't see anything too major, suggestions:

* parse_url() should be changed to parse_and_fix_url(): Rationale: the function name is confusing vs. other urlparse()-like functions out there (we're not just parsing). We're adding something new, e.g. making sure http is explicitly added if the scheme is empty.

* url = parse_url(...), url = uri._replace(...) any instances where you're pairing these two operations (I count three?) I would change this into a single helper function to do both operations at the same time, turning the two lines into two. This is a personal preference whenever I see the same variable being manipulated like this within the same number of lines, so it wraps it in a way that looks more deliberate. It may also help unit testing.

* url as a variable name could be improved (maybe from golang experience - url is just a difficult variable name.

* Lines 104 and 344 in helpers.py the function call is constructed using a parameter:

     verify=not 

* not is a keyword in Python and unless this library explicitly requests the use of the word not, over (?) then i would suggest changing that to another value. (I can't remember not being used like this - so could just be me)

sevein:
@ross-spencer, I've addressed your comments.

I've used vcrpy to test the new methods. It's a handy testing tool that Holly introduced a while ago where you can go beyond unit tests by recording the HTTP requests and responses and keep in them in the repo as fixtures.

ross-spencer (approved):
Looks good @sevein thx for the additional changes.

PR #273, opened 2017-11-25
Menu bar highlights current tab

### LINKED ISSUE #219: Problem: menu bar does not highlight current page
To recreate: Go to admin tab of SS and see that "Home" is highlighted.

jrwdunham:
The SS menu bar will highlight the tab of the current page. Fixes #219.


sevein (approved):
Slightly different from [how it was solved in the Dashboard](https://github.com/artefactual/archivematica/blob/stable/1.6.x/src/dashboard/src/templates/layout.html#L35-L54) but will do it.

PR #275, opened 2017-11-25
Change GPG home directory configuration

### LINKED ISSUE #271: GPG encryption does not work with docker
GPG is expecting to be able to write to a /home/archivematica directory which is not created by the SS's Dockerfile.

jrwdunham:
- If set, the environment variable GNUPG_HOME_PATH will determine where GPG
  writes to disk.
- If the GNUPG_HOME_PATH env var is undefined, GPG will use the absolute path
  of the SS's internal location.
- Fixes #271

### DIFF
--- a/storage_service/common/gpgutils.py
+++ b/storage_service/common/gpgutils.py
@@ -14,17 +14,12 @@ import logging
 
 # Third party dependencies, alphabetical
 import gnupg
+from django.apps import apps
+from django.conf import settings
 from django.utils.translation import ugettext as _
 
 
 LOGGER = logging.getLogger(__name__)
-
-
-# WARNING: by not providing a gnupghome kw param to ``GPG`` here we are
-# deferring to gnupg to determine where the .gnupg/ dir will be. On a default
-# vagrant/ansible deploy, the .gnupg/ dir will be at
-# /var/lib/archivematica/.gnupg/
-gpg = gnupg.GPG()
 
 
 # Defaults for the default AM SS GPG key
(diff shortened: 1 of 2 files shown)


storage_service/storage_service/settings/base.py
```diff
@@ -405,3 +405,5 @@ def is_true(env_str):
         environ.get('SS_BAG_VALIDATION_NO_PROCESSES', 1))
 except ValueError:
     BAG_VALIDATION_NO_PROCESSES = 1
+
+GNUPG_HOME_PATH = environ.get('GNUPG_HOME_PATH', None)
```
sevein: Shouldn't be prefixed with `SS_` like the others?

storage_service/common/gpgutils.py
```diff
@@ -40,23 +35,42 @@
 ENCR_FAILS = 'no'
 
 
+class GPG(object):
```
sevein, line 38: The linter didn't like your style.

storage_service/common/gpgutils.py
```diff
@@ -40,23 +35,42 @@
...
+                Location = apps.get_model(app_label='locations',
+                                          model_name='Location')
+                ss_internal = Location.active.get(
+                    purpose=Location.STORAGE_SERVICE_INTERNAL)
+                gnupg_home_path = ss_internal.full_path
+            self._gpg = gnupg.GPG(gnupghome=gnupg_home_path)
+        return self._gpg
+
+
+gpg = GPG()
```
sevein: Singletons are good for you.

sevein (changes requested):
I've added a couple of observations.

sevein (approved):

Issue #263, opened 2017-11-16
Enhancement: automatic check for django migrations


jhsimpson:
Sometimes new Pull Reqests add code that requires a django migration to be run. It would be very helpful to add a travis check for this to all PR's.

This tool might do the job:
https://github.com/rev112/django-migration-checker

See: artefactual/archivematica#797 - same issue for Archivematica dashboard

sevein:
I did something similar in fpr-admin, here: https://github.com/artefactual/archivematica-fpr-admin/blob/677607a38ed6d18bd8fa1601e1704a4096e30afe/.travis.yml#L7-L15.

jrwdunham:
Fixed by https://github.com/artefactual/archivematica-storage-service/pull/277

PR #277, opened 2017-11-27
Add check for migrations to travis/tox

jrwdunham:
This PR modifies tox.ini and the Travis config file so that Travis CI will fail if the SS needs migrations. This is expected to fail when merged into qa/0.x because a migration is needed.


jrwdunham:
Fixes https://github.com/artefactual/archivematica-storage-service/issues/263

sevein (approved):
Nicely done!

Issue #278, opened 2017-11-28
Problem: a GPG-related migration is needed


jrwdunham:

jrwdunham:
Fixed by https://github.com/artefactual/archivematica-storage-service/pull/279

PR #279, opened 2017-11-28
Add GPG-related minor migration

jrwdunham:
Fixes https://github.com/artefactual/archivematica-storage-service/issues/278


sevein (approved):
I really don't get it but let's trust Django.

sevein:
I think that b19aaa39f64c3087e6298ac9283f54dfbe784db7 should have been part of this PR because now we have qa/0.x with broken state.

PR #280, opened 2017-11-29
Fix `Package.should_have_pointer_file()`

jrwdunham:
Before this, this method was incorrectly returning `True` if the package
was an AIP (or AIC) even if it was not a file (i.e., was not
compressed).

This should fix (or should contribute to the fixing of) https://github.com/artefactual/archivematica/issues/809


jhsimpson (approved):
good catch -this looks good to me

Issue #268, opened 2017-11-22
Problem: Replicators available in interface for all types of locations


sromkey:
Replication only works for AIP Storage location, so there shouldn't be the option to choose replicators for any other type of location (unless/until that functionality is added). 

The risk is a user mistakenly thinking the content from another type of location is bring replicated when it's not.

jrwdunham:
@sromkey should be fixed by https://github.com/artefactual/archivematica-storage-service/pull/283.

My one hesitation about the above PR is that a user could get themselves into this kind of trouble:

1. Create an AS location and assign an existing replicator location to it.
2. Edit the above AS location by changing its purpose to some other non-AS purpose and attempt to save it. The SS will prevent you and will inform you that you cannot do that. However, the "replicators" field will now be hidden because you have changed the purpose to something that is not AS. In order to fix it, you must change the purpose back to AS, de-select the replicators, and then change the purpose back to the originally desired purpose.

Is ^ annoying enough to merit some modification? I could just stop hiding the replicators field based on the selected purpose...

jrwdunham:
Fixed by https://github.com/artefactual/archivematica-storage-service/pull/283

PR #283, opened 2017-11-30
Prevent assigning replicators to non-AS locations

jrwdunham:
If a location is not a an AIP Storage-purposed (AS) location, then it should not be possible to assign a replicator location to it. Fixes issue #268. When merged, this should also be cherry-picked to stable/0.11.x


sevein (approved):

jrwdunham:
Cherry-picked to stable/0.11.x as d230939105584e40baccd1667c6332a52345f8e6

Issue #284, opened 2017-12-01
Problem: Can't configure how browse space method works


jhsimpson:
the storage service Space model has a browse method, that allows a client to get information back about the contents of a particular location.

https://github.com/artefactual/archivematica-storage-service/blob/qa/0.x/storage_service/locations/models/space.py#L205

This method will return information such as the number of objects in a given location (the number of files and folders).  Sometimes calculating this count takes a long time (due to the size or slowness of the underlying filesystem) and the client would like to tell the browse method not to calculate that information.

An example of a workaround that has been implemented by a specific Archivematica user.
https://github.com/artefactual/archivematica-storage-service/commits/stable/0.10.x-ubc

A better solution would allow a way for a client to decide what behaviour they want.

PR #287, opened 2017-12-07
Add a setting to disable object counting when browsing spaces.

jambun:
A shot at fixing issue #284


sevein:
I've changed the target to `qa/0.x` because we're not including this in SS 0.11 - it's too late for that sorry! Thank you @jambun.

sevein (approved):

PR #288, opened 2017-12-08
Ignore errors in shutil.rmtree

### LINKED ISSUE #269: Problem: extract_file endpont may fail to delete temporary folder
The `/api/v1/file/<uuid>/extract_file/` endpoint uses the following code to extract a compressed package:

    (extracted_file_path, temp_dir) = package.extract_file(relative_path_to_file)

Then it runs the following to stream the contents of the file to the client:

```
response = http.FileResponse(open(filepath, 'rb'))
if temp_dir and os.path.exists(temp_dir):
    shutil.rmtree(temp_dir)
return response
```

I guess this approach works because the interpreter holds the file descriptor even when the files are deleted by `shutil.rmtree` meaning that the interpreter can still read the contents of the file. Eventually when the transfer completes Django would release the file descriptor for us and the operative system would be able to claim the space.

However it's been reported by a client a problem creating a new AIC and the logs showed the following:

```
  File "/usr/share/python/archivematica-storage-service/local/lib/python2.7/site-packages/tastypie/resources.py", line 220, in wrapper
    response = callback(request, *args, **kwargs)
  File "/usr/lib/archivematica/storage-service/locations/api/resources.py", line 91, in wrapper
    result = func(resource, request, bundle, **kwargs)
  File "/usr/lib/archivematica/storage-service/locations/api/resources.py", line 655, in extract_file_request
    response = utils.download_file_stream(extracted_file_path, temp_dir)
  File "/usr/lib/archivematica/storage-service/common/utils.py", line 124, in download_file_stream
    shutil.rmtree(temp_dir)
  File "/usr/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/usr/lib/python2.7/shutil.py", line 247, in rmtree
    rmtree(fullname, ignore_errors, onerror)
  File "/usr/lib/python2.7/shutil.py", line 256, in rmtree
    onerror(os.rmdir, path, sys.exc_info())
  File "/usr/lib/python2.7/shutil.py", line 254, in rmtree
    os.rmdir(path)
OSError: [Errno 39] Directory not empty: '/var/archivematica/storage_service/tmpGJtGYm/AIC_test_004-a2b1890f-c089-4d37-b96e-
39486f604445/data'
```

In this case, `shutil.rmtree` failed to delete the `data` sub-directory but I don't exactly how that can happen because the code shouldn't try to do so until it's empty (see https://github.com/python/cpython/blob/2.7/Lib/shutil.py).

I haven't been able to reproduce. One workaround could be to pass `ignore_errors=True` to `shutil.rmtree` (best effort). Or maybe pass an `onerror` so at least we can log it.

sevein:
Using `shutil.rmtree` was problematic in NFS drives (see "silly rename")
because we're trying to delete files that are still being used, i.e. Django
doesn't close the file until the response is fully returned to the browser.

The long-term solution should be to defer the cleanup.

This fixes #269. Read the issue for more details.

It should also be included in qa/0.x.

### DIFF
--- a/storage_service/common/utils.py
+++ b/storage_service/common/utils.py
@@ -132,7 +132,7 @@ def download_file_stream(filepath, temp_dir=None):
 
     # Delete temp dir if created
     if temp_dir and os.path.exists(temp_dir):
-        shutil.rmtree(temp_dir)
+        shutil.rmtree(temp_dir, ignore_errors=True)
 
     return response
 


jhsimpson (approved):
double plus good

sevein:
3596fd5 in qa/0.x.

Issue #299, opened 2017-12-31
Feature: Add Swagger documentation to API


jraddaoui:
Add Swagger documentation and UI using https://github.com/concentricsky/django-tastypie-swagger

jraddaoui:
Requires the fix for #298

jraddaoui:
Doing the same in https://github.com/artefactual/archivematica will require more work because almost everything in there are custom methods.

PR #300, opened 2017-12-31
Add Swagger documentation to API

jraddaoui:
Use django-tastypie-swagger to create Swagger documentation and UI,
available at '.../api/v1/doc/' and '.../api/v2/doc/'.

TODO:
- [x] - Fix file/package resource
- [ ] - Add auth. to doc pages?
- [ ] - Add custom endpoints/methods

connects to #299

### DIFF
--- a/storage_service/locations/api/urls.py
+++ b/storage_service/locations/api/urls.py
@@ -19,6 +19,18 @@ v2_api.register(v2.PipelineResource())
 urlpatterns = [
     url(r'', include(v1_api.urls)),
     url(r'v1/sword/$', views.service_document, name='sword_service_document'),
+    url(r'v1/doc/',
+        include('tastypie_swagger.urls', namespace='v1_api_swagger'),
+        kwargs={
+            'tastypie_api_module': v1_api,
+            'namespace': 'v1_api_swagger',
+            'version': '2.0'}),
     url(r'', include(v2_api.urls)),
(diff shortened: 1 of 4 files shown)


storage_service/locations/api/urls.py
```diff
@@ -19,6 +19,18 @@
...
+            'namespace': 'v1_api_swagger',
+            'version': '2.0'}),
     url(r'', include(v2_api.urls)),
     url(r'v2/sword/$', views.service_document, name='sword_service_document'),
+    url(r'v2/doc/',
+        include('tastypie_swagger.urls', namespace='v2_api_swagger'),
+        kwargs={
+            'tastypie_api_module': v2_api,
+            'namespace': 'v2_api_swagger',
+            'version': '1.0'}),
```
jrwdunham, line 35: @jraddaoui are the versions mixed up here? It looks like the v2/ endpoints are associated to version 1.0 and the v1/ endpoints are associated to version 2.0.

jrwdunham:
This PR makes the Swagger UI visible at /api/v1/doc/ and /api/v2/doc/. It looks like this:

am-ss-swagger-ui-screenshot

storage_service/locations/api/urls.py
jraddaoui: Hi @jrwdunham,

Yes, correct namespace and module but swapped version string, sorry ;)

PR #302, opened 2018-01-03
Allow SIP to be created from encrypted transfer

### LINKED ISSUE #301: Problem: SIPs cannot be created from encrypted transfers in backlog
To re-create:

1. Create a transfer backlog location in an encrypted GPG space in the SS.
2. Disable the default transfer backlog location.
3. Enable the encrypted transfer backlog location.
4. Create a transfer and send it to backlog
5. In the ingest tab, attempt to create a SIP from the encrypted transfer in backlog by dragging the encrypted transfer to the arrange/ folder. The GUI will hang showing the dragged transfer over the arrange/ folder but nothing will happen. It will not be possible to create a SIP from the encrypted transfer.

jrwdunham:
The `_encr_path2key_fingerprint` function was hanging when attempting to query the db for packages whose current path ended with the provided encrypted path. Since the encrypted path can be absolute, the correct behaviour is to find the package whose current path is contained within the provided encrypted path. Fixes #301.

This PR also adds extensive unit tests for the GPG space.

### DIFF
--- a/storage_service/locations/models/gpg.py
+++ b/storage_service/locations/models/gpg.py
@@ -11,6 +11,7 @@ from uuid import uuid4
 from metsrw.plugins import premisrw
 
 # Core Django, alphabetical
+from django.conf import settings
 from django.db import models
 from django.utils.translation import ugettext as _, ugettext_lazy as _l
 
@@ -271,13 +272,20 @@ def _encr_path2key_fingerprint(encr_path):
     used to encrypt the package. Since it was already encrypted, its
     model must have a GPG fingerprint.
     """
-    for package in Package.objects.all():
-        if package.current_path in encr_path:
-            return package.encryption_key_fingerprint
-    fail_msg = 'Unable to find package matching encrypted path {}'.format(
-        encr_path)
-    LOGGER.error(fail_msg)
-    raise GPGException(fail_msg)
+    if 'sqlite' in settings.DATABASES['default']['ENGINE']:
+        sql = ('SELECT * FROM locations_package WHERE %s LIKE "%" ||'
+               ' current_path || "%"')
+    else:
+        sql = ('SELECT * FROM locations_package WHERE %s LIKE CONCAT(\'%%\','
+               ' current_path, \'%%\')')
+    matches = list(Package.objects.raw(sql, [encr_path]))
+    try:
+        return matches[0].encryption_key_fingerprint
+    except IndexError:
+        fail_msg = 'Unable to find package matching encrypted path {}'.format(
+            encr_path)
+        LOGGER.error(fail_msg)
+        raise GPGException(fail_msg)
 
 
 # This replaces non-unicode characters with a replacement character,
--- a/storage_service/locations/models/gpg.py
+++ b/storage_service/locations/models/gpg.py
@@ -24,6 +24,7 @@ from common import gpgutils
 # This module, alphabetical
 from .location import Location
 from .package import Package
+from . import space
 
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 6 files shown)


storage_service/locations/models/gpg.py
```diff
@@ -251,28 +229,51 @@ def _gpg_encrypt(path, key_fingerprint):
...
     used to encrypt the package. Since it was already encrypted, its
     model must have a GPG fingerprint.
     """
-    return Package.objects.get(
-        current_path__endswith=encr_path).encryption_key_fingerprint
+    sql = ('SELECT * FROM locations_package WHERE %s LIKE CONCAT(\'%%\','
+           ' current_path, \'%%\')')
+    if _db_engine() == 'sqlite':
+        sql = ('SELECT * FROM locations_package WHERE %s LIKE "%" ||'
+               ' current_path || "%"')
```
jraddaoui, line 268: Couldn't we use [`contains`](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#contains) instead of a raw query? That way we don't need to know the DB engine.

storage_service/locations/tests/test_gpg.py
```diff
@@ -0,0 +1,498 @@
...
+
+        encr_path = '/abs/path/to/{}/data/objects/somefile.jpg'.format(
+            exp_curr_path)
+        assert gpg._encr_path2key_fingerprint(encr_path) == EXP_FINGERPRINT
+
+        with pytest.raises(gpg.GPGException) as excinfo:
+            encr_path = '/some/non/matching/path.jpg'
+            gpg._encr_path2key_fingerprint(encr_path)
+        assert 'Unable to find package matching encrypted path {}'.format(
+            encr_path) in str(excinfo.value)
```
jraddaoui, line 498: Impressive tests!

One minor detail, some of them are declared with two underscores and other with only one, is there a reason for that? If not, I think we usually use only one.

jraddaoui (changes requested):
Great job @jrwdunham! Really nice tests and good clean-ups along the way.

I have suggested a change and added a minor nitpick.

storage_service/locations/models/gpg.py
jrwdunham: I don't think this is possible @jraddaoui because we are not checking whether `current_path` contains a given string but whether a given string contains `current_path`. This is why I'm doing the db engine gymnastics. I'll think about this a bit more, but I don't think there is a better solution. (Please let me know if you see an obvious one.)

storage_service/locations/tests/test_gpg.py
jrwdunham: The two underscores is actually a consistency because I'm naming each test func with `test_` followed by the name of the function/method being tested. The two underscores cases arise because the function/method being tested has a name that starts with an underscore.

storage_service/locations/models/gpg.py
jraddaoui: Oh, right, it's the other way around. Would it be possible to strip the part we need from `encr_path`? If not, I can't find a better solution.

storage_service/locations/tests/test_gpg.py
jraddaoui: Great, thanks!

jraddaoui (approved):

storage_service/locations/models/gpg.py
jrwdunham: Yeah, I thought of that, but I don't think there is a reliable way of stripping away the location's full path because we don't have access to the location instance. Also, we can't just use the filename of `encr_path` because it may not be a file within an AIP, in which case the `encr_path` file name would not be a substring of the package's current path.

jrwdunham:
Cherry-picked to stable/0.11.x as
0c2e541907748a89ebbd2c90477380c701231c68, 22fd77aedce952e5c0171625aabc21127bdf8982, and cf9a3b3305d38a1d307e56d23bc7ce78c7cb6850.

Issue #314, opened 2018-02-01
Problem: Creating Transfers and Storing AIPs requires files to be copied twice


jhsimpson:
Each type of Space defined in the storage service requires two methods to be defined - `move_to_storage_service()` and `move_from_storage_service()` (see https://github.com/artefactual/archivematica-storage-service/blob/stable/0.11.x/storage_service/locations/models/space.py#L79-L85).

These methods are used in (at least) two places:
 
- when creating a transfer (see https://github.com/artefactual/archivematica-storage-service/blob/stable/0.11.x/storage_service/locations/api/resources.py#L328)
- when storing an aip (see https://github.com/artefactual/archivematica-storage-service/blob/stable/0.11.x/storage_service/locations/models/package.py#L579)

This requires files to be moved from their current location (the source) to the storage service internal storage location, and then copied from there to the final location (the destination).  It is not possible to copy the files from source to destination.  This results in a significant increase in processing time, and doubles the (temporary) storage requirments.  

It should be possible to only peform one move or copy operation when starting transfers and when storing aips.

jhsimpson:
@jambun @marktriggs ping - this is the issue we were discussing by email.

jambun:
Thanks @jhsimpson

jambun:
Also pinging @payten

jambun:
Hi @jhsimpson,

@payten and I have been looking at this. @payten has a commit on our fork at:
https://github.com/hudmol/archivematica-storage-service/commit/2dd36fada1c1149cb828e0273282c5364e8899e7

We took the approach of replacing the move_to and move_from method calls with a call to a new move_direct method which does a move directly from the source to the destination without stopping off via the staging area.

We've only implemented a proof of concept for local_filesystem so far.

Let us know if we're on the right track and we'll push on.


Cheers,
James

sevein:
This was fixed by @payten in #320.
Related issue: https://github.com/artefactual-labs/archivematica-acceptance-tests/issues/76.

jrwdunham:
Reopening this for waffle/kanban purposes. After successful QA, this will be closed.

nickwilkinson:
![](https://github.trello.services/images/mini-trello-icon.png) [Columbia: Phase 2](https://trello.com/c/KCIoHnmo/7-columbia-phase-2)

joel-simpson:
[...]

PR #320, opened 2018-02-08
Reduce number of copies by introducing space.posix_move

jrwdunham:
Connected to #314

@payten I've created a PR here just so we can discuss this work. I realize it's a WIP, but I thought this discussion might be useful as it progresses.

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -35,7 +35,7 @@ from administration.models import Settings
 from common import utils
 from locations.api.sword import views as sword_views
 
-from ..models import (Callback, CallbackError, Event, File, Package, Location, Space, Pipeline, StorageException, Async)
+from ..models import (Callback, CallbackError, Event, File, Package, Location, Space, Pipeline, StorageException, Async, PosixMoveUnsupportedError)
 from ..forms import SpaceForm
 from ..constants import PROTOCOL
 from locations import signals
@@ -347,14 +347,14 @@ class LocationResource(ModelResource):
             destination_path = os.path.join(
                 destination_location.relative_path, destination_path)
 
-            if Space.posix_move_supported(origin_space, destination_space):
+            try:
                 origin_space.posix_move(
                     source_path=source_path,
                     destination_path=destination_path,
                     destination_space=destination_space,
                     package=None
                 )
-            else:
+            except PosixMoveUnsupportedError:
                 origin_space.move_to_storage_service(
                     source_path=source_path,
                     destination_path=destination_path,
@@ -363,6 +363,7 @@ class LocationResource(ModelResource):
                 origin_space.post_move_to_storage_service()
                 destination_space.move_from_storage_service(
                     source_path=destination_path,
+                    destination_path=destination_path,
                     package=None,
                 )
                 destination_space.post_move_from_storage_service(
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -346,19 +346,27 @@ class LocationResource(ModelResource):
                 origin_location.relative_path, source_path)
             destination_path = os.path.join(
                 destination_location.relative_path, destination_path)
-            origin_space.move_to_storage_service(
-                source_path=source_path,
-                destination_path=destination_path,
-                destination_space=destination_space,
-            )
-            origin_space.post_move_to_storage_service()
-            destination_space.move_from_storage_service(
-                source_path=destination_path,
-                destination_path=destination_path,
-                package=None,
-            )
-            destination_space.post_move_from_storage_service(
-                destination_path, destination_path)
+
+            if Space.posix_move_supported(origin_space, destination_space):
+                origin_space.posix_move(
+                    source_path=source_path,
+                    destination_path=destination_path,
+                    destination_space=destination_space,
+                    package=None
+                )
+            else:
+                origin_space.move_to_storage_service(
+                    source_path=source_path,
+                    destination_path=destination_path,
+                    destination_space=destination_space,
+                )
+                origin_space.post_move_to_storage_service()
+                destination_space.move_from_storage_service(
+                    source_path=destination_path,
+                    package=None,
+                )
+                destination_space.post_move_from_storage_service(
+                    destination_path, destination_path)
 
     def _handle_location_file_move(self, move_files_fn, request, *args, **kwargs):
         """
--- a/storage_service/locations/models/local_filesystem.py
+++ b/storage_service/locations/models/local_filesystem.py
@@ -34,10 +34,6 @@ class LocalFilesystem(models.Model):
         Location.REPLICATOR,
     ]
 
-    # Flag this space as a POSIX filesystem to allow direct moves between
-    # POSIX compatible spaces where possible.
-    POSIX_FILESYSTEM = True
-
     def move_to_storage_service(self, src_path, dest_path, dest_space):
         """ Moves src_path to dest_space.staging_path/dest_path. """
         # Archivematica expects the file to still be on disk even after stored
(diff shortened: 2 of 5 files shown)


storage_service/locations/api/resources.py
```diff
@@ -381,19 +381,19 @@ def post_detail(self, request, *args, **kwargs):
                     origin_location.relative_path, source_path)
                 destination_path = os.path.join(
                     destination_location.relative_path, destination_path)
-                origin_space.move_to_storage_service(
+                origin_space.move_direct(
```
jrwdunham: Should we be trying `move_direct` and falling back to the old way if a `NotImplementedError` is raised?

storage_service/locations/models/space.py
```diff
@@ -261,6 +261,28 @@ def delete_path(self, delete_path, *args, **kwargs):
...
+        source_path = os.path.join(self.path, source_path)
+
+        if os.path.isabs(destination_path):
+            destination_path = destination_path.lstrip(os.sep)
+
+        destination_path = os.path.join(destination_space.path, destination_path)
+        LOGGER.debug('move_direct: absolute destination_path: %s', destination_path)
+
+        try:
+            self.get_child_space().move_direct(
```
jrwdunham: `Space.move_direct` needs to return the return value of `self.get_child_space().move_direct()` and not just call it; this is required by `storage_effects = v.src_space.move_direct` in models/package.py where, as things stand, `storage_effects` will be `None`.

payten: Ah thanks! We'll take a closer look at the other usages/implementers of the move_to_storage_service/move_from_storage_service to make we haven't missed anything else like that.

storage_service/locations/api/resources.py
payten: Thanks @jrwdunham, appreciate the review.

Great idea, thanks.

payten:
Hi @jrwdunham, @jhsimpson,

We've cleaned up and refocused the code so it's clear we're only supporting the new "direct move" behaviour on POSIX filesystems.  As discussed offline, making this code work for different protocols could be done but would require further refactoring.  One possible approach would be to extend the space API so a source space produces a generic bytestream which the destination space could then consume. However, we believe the current approach should solve the immediate needs.

~I'll note that there's a good chance this code will conflict with the refactored code in the `async` branch.  Let us know if you'd like this code rebased at all and we'll make sure it's all still worky.~ `async` has been merged and we've rebased this branch with the latest code.

Thanks!

storage_service/locations/models/space.py
```diff
@@ -648,6 +669,19 @@ def _delete_path_local(self, delete_path):
...
+        """
+        Are the two spaces POSIX filesytems and do they support `posix_move`
+        to allow a direct move bypassing staging.
+        """
+        return (hasattr(source_space.get_child_space(), 'POSIX_FILESYSTEM') and
+            source_space.get_child_space().POSIX_FILESYSTEM and
+            hasattr(source_space.get_child_space(), 'posix_move') and
+            hasattr(destination_space.get_child_space(), 'POSIX_FILESYSTEM') and
+            destination_space.get_child_space().POSIX_FILESYSTEM and
+            hasattr(destination_space.get_child_space(), 'posix_move'))
```
jrwdunham: Is it not true that a space-type model will have `POSIX_FILESYSTEM = True` if and only if it also has a `posix_move` method defined? If so, then I think you can remove the `POSIX_FILESYSTEM` class attribute altogether and simplify this test to just `return (hasattr(source_space.get_child_space(), 'posix_move') and hasattr(destination_space.get_child_space(), 'posix_move')`.

If what I said is not true, then two of these conditions could at least be removed by using `getattr`: `getattr(source_space.get_child_space(), 'POSIX_FILESYSTEM', False)`.

storage_service/locations/models/local_filesystem.py
```diff
@@ -51,3 +55,8 @@ def verify(self):
         verified = os.path.isdir(self.space.path)
         self.space.verified = verified
         self.space.last_verified = datetime.datetime.now()
+
+    def posix_move(self, source_path, destination_path, destination_space, package=None, *args, **kwargs):
```
jrwdunham: You don't need `*args, **kwargs` here, right? I mean all of the `posix_move` methods have the same signature. I think `*args, **kwargs` should be removed from the definitions of `posix_move` in local_filesystem.py, nfs.py, and space.py.

storage_service/locations/api/resources.py
```diff
@@ -346,19 +346,27 @@ def _move_files_between_locations(self, files, origin_location, destination_loca
...
+                )
+            else:
+                origin_space.move_to_storage_service(
+                    source_path=source_path,
+                    destination_path=destination_path,
+                    destination_space=destination_space,
+                )
+                origin_space.post_move_to_storage_service()
+                destination_space.move_from_storage_service(
+                    source_path=destination_path,
```
jrwdunham, line 365: The required `destination_path` arg is missing here.

storage_service/locations/models/space.py
```diff
@@ -261,6 +261,27 @@ def delete_path(self, delete_path, *args, **kwargs):
...
+    def posix_move(self, source_path, destination_path, destination_space, package=None, *args, **kwargs):
+        """
+        Move self.path/source_path direct to destination_space.path/destination_path bypassing staging.
+        """
+        LOGGER.debug('posix_move: source_path: %s', source_path)
+        LOGGER.debug('posix_move: destination_path: %s', destination_path)
+        LOGGER.debug('posix_move: destination_space.path: %s', destination_space.path)
+
+        if not Space.posix_move_supported(self, destination_space):
+            raise ValidationError(_('both spaces must be POSIX filesystems'))
```
jrwdunham: You can just call the static method on the instance, i.e., `self.posix_move_supported`. It seems odd to call it from the `Space` class here.

Also, the Django `ValidationError` seems odd here to me given that this is not really form validation. Probably better to just use Python's builtin `TypeError`.

jrwdunham: Also, it seems odd, to duplicate the call to `posix_move_supported` once before you call `posix_move` and once inside it. Why not just attempt to call `.posix_move` and accept the `TypeError` and fallback to `move_to_storage_service`/`move_from_storage_service`, e.g., in resources.py? I can see that the hairiness in package.py may make such an "ask for forgiveness" approach hard (impossible?). If it could be done, I think it would be preferable.

storage_service/locations/models/package.py
```diff
@@ -674,49 +674,81 @@ def _store_aip_to_pending(self, origin_location, origin_path):
...
-                                     self.origin_path),
-            destination_path=self.current_path,  # This should include Location.path
-            destination_space=v.dest_space)
-        self.status = Package.STAGING
-        self.save()
-        v.src_space.post_move_to_storage_service()
-        storage_effects = v.dest_space.move_from_storage_service(
-            source_path=self.current_path,  # This should include Location.path
-            destination_path=os.path.join(
+        if Space.posix_move_supported(v.src_space, v.dest_space):
```
jrwdunham: As detailed more below, I wonder if you could not instead refactor this to try the `posix_move` call and fallback to the `move_to/from` call if a `TypeError` is raised, which would remove the double call to `posix_move_supported`.

storage_service/locations/models/package.py
```diff
@@ -674,49 +674,81 @@ def _store_aip_to_pending(self, origin_location, origin_path):
...
+            # Both spaces are POSIX filesystems and support `posix_move`
+            #
+            # 1. move direct to the SS destination space/location,
+            # 2. setting the status to "uploaded" (if applicable),
```
jrwdunham: The "if applicable" parenthetical doesn't really apply here if the LOM/ARKIVUM condition has been removed.

jrwdunham (commented):
@payten I haven't actually tested this code by running transfers through AM yet, but I have some comments. (I can't approve or request changes, I think because the PR was technically opened by me.) I'll do some actual testing and may add more comments after.

payten:
Thanks for the help @jrwdunham! We've cleaned up those bits and fixed the dropped argument - thanks for catching that.

To avoid ambiguity, we thought it might be nice to have a custom exception, `PosixMoveUnsupportedError`, rather than using `TypeError`. Does that seem ok?  And changing it to use an exception was ok too.

Also, happy to close this out and create a new PR if that would be easier for the review process? And will squash all commits if preferred.

jrwdunham:
@payten thanks. This is looking good to me. I tested with it and confirmed that LocalFilesystem-type space (LFS) to LFS works: `posix_move` gets called and doesn't raise an exception. I also changed the name of that method in `LocalFilesystem` in my test deploy and confirmed that that causes it to fallback to the old spoke->hub->spoke-style move.

Adding the new commit was helpful for me during code review, but yeah, I think squashing these would be good in this case. I don't think a new PR is necessary.

sevein (approved):

sevein:
I've realized that the squashed merge resulted in showing the author as @jrwdunham instead of @payten. Sorry about that and props to @payten!

PR #327, opened 2018-02-16
Add asynchronous versions of endpoints that perform long-running IO.

marktriggs:
@jhsimpson Here's the first of the pull requests we discussed yesterday.  Let me know if there's an issue number I should add to the commit or anything like that.

Connected to #374


marktriggs:
@sevein I've fixed up the linter issues and appeased `checkformigrations.sh` now.  I'm still seeing one error coming from `test_create_pointer_file` that I don't *think* is due to my changes.  Happy to dig in if it's helpful, though.

sevein (approved):

PR #328, opened 2018-02-16
Add "Request Deletion" link to package rows

### LINKED ISSUE #326: Problem: users cannot make AIP deletion requests via the GUI
The solution would be to add a "Delete" link/button to the rows on the /packages/ page.

jrwdunham:
Fixes #326

### DIFF
--- a/storage_service/static/js/project.js
+++ b/storage_service/static/js/project.js
@@ -27,4 +27,42 @@ $(document).ready(function() {
           },
       }
     });
-} );
+
+    $("a.request-delete").click(function() {
+        var self = $(this);
+        var uuid = self.data('package-uuid');
+        var pipeline = self.data('package-pipeline');
+        var $userDataEl = $("#user-data-packages");
+        var userId = $userDataEl.data('user-id');
+        var userEmail = $userDataEl.data('user-email');
+        var userUsername = $userDataEl.data('user-username');
+        var userAPIKey = $userDataEl.data('user-api-key');
+        var uri = $userDataEl.data('uri');
+        var formData = {
+            "event_reason": "Storage Service user wants to delete AIP " + uuid + ".",
+            "pipeline": pipeline,
+            "user_id": userId,
+            "user_email": userEmail};
+        $.ajax({
+            type: "POST",
+            url: uri + "api/v2/file/" + uuid + '/delete_aip/',
+            data: JSON.stringify(formData),
+            dataType: "json",
+            contentType: "application/json; charset=utf-8",
+            headers: {"Authorization": "ApiKey " + userUsername + ":" + userAPIKey},
+            success: function(data) {
+                $("div#package-delete-alert").remove();
+                $("h1").first().after(
+                    "<div id='package-delete-alert' class='alert alert-success'>" +
+                    data.message + "</div>");
+            },
+            failure: function(errMsg) {
+                $("div#package-delete-alert").remove();
+                $("h1").first().after(
+                    "<div id='package-delete-alert' class='alert alert-warning'>" +
+                    data.message + "</div>");
+            }
+        });
+        return false;
+    });
+});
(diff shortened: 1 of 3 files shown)


storage_service/static/js/project.js
```diff
@@ -27,4 +27,34 @@ $(document).ready(function() {
...
+            "pipeline": pipeline,
+            "user_id": userId,
+            "user_email": userEmail};
+        $.ajax({
+            type: "POST",
+            url: "/api/v2/file/" + uuid + '/delete_aip/',
+            data: JSON.stringify(formData),
+            dataType: "json",
+            contentType: "application/json; charset=utf-8",
+            headers: {"Authorization": "ApiKey " + userUsername + ":" + userAPIKey},
```
sevein: Is it the cookie not being sent? AFAIK the SS API supports SessionAuthentication.

storage_service/static/js/project.js
```diff
@@ -27,4 +27,34 @@ $(document).ready(function() {
...
+        var userUsername = $userDataEl.data('user-username');
+        var userAPIKey = $userDataEl.data('user-api-key');
+        var formData = {
+            "event_reason": "Storage Service user wants to delete AIP " + uuid + ".",
+            "pipeline": pipeline,
+            "user_id": userId,
+            "user_email": userEmail};
+        $.ajax({
+            type: "POST",
+            url: "/api/v2/file/" + uuid + '/delete_aip/',
```
sevein: If you don't add the base url this wouldn't work when SS is deployed from a subpath right?

storage_service/static/js/project.js
```diff
@@ -27,4 +27,34 @@ $(document).ready(function() {
...
+            data: JSON.stringify(formData),
+            dataType: "json",
+            contentType: "application/json; charset=utf-8",
+            headers: {"Authorization": "ApiKey " + userUsername + ":" + userAPIKey},
+            success: function(data) {
+                alert(data.message);
```
sevein: Alerts are considered harmful, should we avoid them?
Maybe we can file a bug to remember?

storage_service/templates/snippets/packages_table.html
```diff
@@ -76,8 +76,19 @@
           {% if package.package_type in 'AIP AIC' and not package.replicated_package %}
             <a href="{% url 'aip_reingest' package.uuid %}?next={{ request.path }}">{% trans "Re-ingest" %}</a>
           {% endif %}
+          {% if package.package_type in package.PACKAGE_TYPE_CAN_DELETE %}
+          <a href="javascript:void(0)" class="request-delete"
```
sevein: I vaguely remember that `javascript:void` shouldn't be used.
I've found this: https://stackoverflow.com/a/1293130.

I remember in AngularJS the preferred method was to use `href="#"` + `return false` or buttons.

sevein (commented):
I'm not very fond of this solution because it introduces a new pattern has it hasn't been discussed properly. I understand your reasons though. I'd be fine with this as a temporary solution.

storage_service/static/js/project.js
jrwdunham: Sure, I'll fix this to provide a full URL.

storage_service/static/js/project.js
jrwdunham: Or we could have the JS create an alert that looks like the Django one we use elsewhere...

storage_service/templates/snippets/packages_table.html
jrwdunham: Ok, changed to use `href="#"` and `return false`.

storage_service/static/js/project.js
jrwdunham: On my dev machine, the browser is not sending the cookie, presumably because there is no domain being used and the cross-domain no-cookies thing is kicking in. Without adding CORS/OPTIONS logic, I think sending the API key in the headers is the best solution.

storage_service/static/js/project.js
sevein: Makes sense. There is already a message framework in Django to do this but I guess you can't use it because this is now happening on the client side.

sevein (approved):

jrwdunham:
Cherry-picked to stable/0.11.x as dcad94fa55bc45691c3fdcb9714325732b516898.

PR #335, opened 2018-03-02
Provides lowercase checking of API reingest type

### LINKED ISSUE #325: Issue: Storage service doesn't adopt the robustness principle for v2/ingest/reingest api call
The following call to reingest will send a reingest to `Ingest` instead of `Transfer` when an upper case string of `FULL` is used. This is misleading to the user. It should probably return something else, or accept an upper- or lower-case string.

`curl -v -H "Authorization: ApiKey test:test" -H "Content-Type: application/json" -X POST -d '{"pipeline":"bdf745b9-139d-4438-abe2-0894d9f2ff31","reingest_type":"FULL"}' 'http://127.0.0.1:62081/api/v2/file/07189108-658c-44c2-a053-1ccda2f15990/reingest/' `

Extra logging shows that we're looking for an ALL-CAPS string:

```
archivematica-storage-service_1  | DEBUG     2018-02-13 12:13:42  locations.models.package:package:start_reingest:1888: TYPE: received: FULL wanted: full
archivematica-storage-service_1  | DEBUG     2018-02-13 12:13:42  locations.models.package:package:start_reingest:1894: TARGET: ingest
```
And this is in this portion of code here:

https://github.com/artefactual/archivematica-storage-service/blob/stable/0.11.x/storage_service/locations/models/package.py#L1886

It is especially confusing as the docs describe an upper-case string: https://wiki.archivematica.org/index.php?title=Storage_Service_API&oldid=12269

ross-spencer:
As part of the payload to the API call vs. the API uri itself we can make
our check against reingest-type lowercase to allow for cases where the user
has submitted an uppercase string. It also allows for cases where our docs
might be wrong and describe an upper case string.

Connected to #325

**NB** This is submitted against `qa/0.x`. I don't anticipate this will be wanted in `stable/0.11` at this point.

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -125,7 +125,9 @@ class Package(models.Model):
     PACKAGE_TYPE_CAN_RECOVER = (AIP, AIC)
     PACKAGE_TYPE_CAN_REINGEST = (AIP, AIC)
 
-    # Reingest type options
+    # Reingest type options described in lower case. The body of the request
+    # will have this value converted to lower case to check it against these
+    # values to make the API responses more robust to human error.
     METADATA_ONLY = 'metadata'  # Re-ingest metadata only
     OBJECTS = 'objects'         # Re-ingest metadata and objects for DIP generation
     FULL = 'full'               # Full re-ingest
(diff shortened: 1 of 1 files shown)


jrwdunham (approved):
This makes sense to me. (A more robust solution would be to have all API calls make their params go through validation, but that's for future work I would guess.)

ross-spencer:
@jrwdunham would be good to see as our API work continues to evolve. Will get this merged, etc. etc.

PR #341, opened 2018-03-09
Delete temporary tar file for uncompressed AIP

### LINKED ISSUE #340: Problem: temporary tar file for uncompressed AIP download is not deleted
When uncompressed AIPs are downloaded, temporary .tar files for them are created. Those temporary files need to be deleted after used, or they can cause disk space problems, especially when using batch scripts like https://github.com/artefactual/automation-tools/tree/master/aips

hakamine:
When an uncompressed AIP is downloaded, a temporary .tar file for it
is created. Delete the temporary file after downloaded.

Fixes #340

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -784,7 +784,7 @@ class PackageResource(ModelResource):
             full_path = package.get_download_path(lockss_au_number)
         except StorageException:
             full_path, temp_dir = package.compress_package(utils.COMPRESSION_TAR)
-        response = utils.download_file_stream(full_path)
+        response = utils.download_file_stream(full_path, temp_dir)
         return response
 
     @_custom_endpoint(expected_methods=['get'])


sevein:
@hakamine should we target qa instead?

hakamine:
@sevein even if we target qa we would need the fix applied to a client system (i.e., cherry picked to stable/0.10.x and stable/0.11.x). I targeted this PR to stable/0.11.x as I tested it on a server based off this branch. Would it be better to resubmit targeting qa?

sevein:
Ok I wasn't sure if this was a critical, considering that we reached code freeze but it looks good to me and pretty safe.

sevein (approved):

sevein:
@hakamine can you cherry pick into qa?

hakamine:
16c33bd in qa/0.x and a02983f in stable/0.10.x

sevein:
thanks thanks

Issue #324, opened 2018-02-09
Problem: uncompressed AIPs need pointer files


jrwdunham:
Pointer files are needed to describe uncompressed AIPs. (They are already created for compressed AIPs.) Motivations:

1. Encrypted uncompressed AIPs (which are tarred) need pointer files so we can know which GPG key is needed to decrypt them.
2. Replica uncompressed AIPs need pointer files so we can know what original AIP is their source.
   - Relatedly, uncompressed AIPs cannot currently be replicated. AIP replication requires a pointer file. 
3. Improving the performance of fixity checking requires pointer files for uncompressed AIPs so that bag-based fixity checking of the entire AIP can be documented in the pointer files of all kinds of AIPs. See https://github.com/artefactual/archivematica/issues/918.

The scope of fixing this issue should include introduction of METS/PREMIS validation using the validation already implemented in mets-reader-writer.

TODO: provide an estimate of what resolving this issue would entail.

joel-simpson:
Adding some context and additional detail following a brief discussion with @sromkey.... one of the reasons uncompressed AIPS did not have pointer files was due to a limitation in earlier versions of PREMIS, where you could only describe a bitstream, but not an intellectual entity. PREMIS 3 now allows for description of intellectual entity. 

I would add a further motivation for this fix... simple consistency. It would simply make more sense to have a pointer file for every AIP (one less 'exception' or special case for users to have to understand).  Alternatively, the UI should explain why there isn't a pointer file, or at least not provide a button to view the pointer file when one doesn't exist (currently if you press that button for an uncompressed AIP you get a 404 message from the storage service which is not very user-friendly).

sallain:
@ross-spencer Assigning this to you for the future - it's for 1.12, but probably needs some analysis!

jens-st:
I agree with @joel-simpson, more consistency for AIPs and less confusion for dashboard users.
Clicking on Pointer file --> "View" on a previous selected AIP still results in a 404 (not found) error for the user in Archivematica 1.13.0.

`{"success": false, "message": "Storage service returned 404; check logs?"}`

PR #350, opened 2018-05-01
Fix store_aip wrt uncompressed AIPs (no pointer)

### LINKED ISSUE #349: Problem: encrypted uncompressed AIPs cannot be stored
Running the [@aip-encrypt @uncompressed](https://github.com/artefactual-labs/archivematica-acceptance-tests/blob/master/features/core/aip-encryption.feature#L44) acceptance test, reveals that encrypted uncompressed AIP are not being stored.

At the code level, in package.py the following call is assuming that `pointer_file` is not `None`, despite the fact that it can be `None` when the AIP is uncompressed:

    self.create_new_pointer_file_given_storage_effects(pointer_file...)

This bug was being masked by the fact that Archivematica was previously incorrectly failing to fail when a "store AIP" request to the Storage Service was failing. The fixing of that issue (in [PR 1049](https://github.com/artefactual/archivematica/pull/1049)) revealed issue described here. In short, this bug was always happening but was not visible until PR 1049 was merged.

jrwdunham:
Fixes the `store_aip` method of package.py so that it does not assume that there is a pointer file. Before this change, `AttributeError` was being raised on a `None` pointer file reference and uncompressed AIPs were being stored but their replication was never possible.

Fixes #349

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -530,11 +530,12 @@ class Package(models.Model):
             # a ``lxml.etree._Element`` instance and not the expected
             # ``premisrw.PREMISObject``. As a result, the following is required:
             replica_pointer_file = replica_package.get_pointer_instance()
-            revised_replica_pointer_file = (
-                replica_package.create_new_pointer_file_given_storage_effects(
-                    replica_pointer_file, replica_storage_effects))
-            write_pointer_file(revised_replica_pointer_file,
-                               replica_package.full_pointer_file_path)
+            if replica_pointer_file:
+                revised_replica_pointer_file = (
+                    replica_package.create_new_pointer_file_given_storage_effects(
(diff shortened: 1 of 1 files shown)


jhsimpson (commented):
this looks good to me.  I haven't tested it yet though.

One question, I think this change could be in a patch release (i.e. 0.11.1) rather than in a minor release (0.12.0).  Does that make sense?

sevein (approved):

jrwdunham:
Note: replication of uncompressed AIPs was never possible in the first place. This is an outstanding issue. See https://github.com/artefactual/archivematica-storage-service/issues/324#issue-296031539

This PR makes it so that uncompressed AIPs can at least be stored without displaying a failure in the GUI to the user who stores them.

jhsimpson:
This is not yet cherry-picked to qa/0.x, I think it should be, @jrwdunham

PR #262, opened 2017-11-10
Add API endpoints for search

jrwdunham:
**26 May 2020: Review of current status against stable/1.11.x**

* Developed against Archivematica 1.7 qa branch (1.7 was released some months later).
* The second PREMIS reader/writer module ([yaprw](https://github.com/artefactual-labs/mets-reader-writer/pull/34)) was never merged. It would need to be merged or existing `metsrw` PREMIS capabilities swapped in.
* The PR associated with [yaprw](https://github.com/artefactual-labs/mets-reader-writer/pull/34) looks like it may also provide enhanced Dublin Core (DC) handling.
* Only 9 `.py` files updated in this PR, the conflicting files listed below by Github are: 
   ```
    storage_service/locations/api/resources.py, 
    storage_service/locations/api/urls.py, 
    storage_service/locations/models/event.py, 
    storage_service/locations/models/package.py, 
    storage_service/locations/tests/test_package.py
   ```
* Additionally, the requirements management process has changed across versions.
* Documentation with examples: [rendered text](https://github.com/artefactual/archivematica-storage-service/blob/50d635878f2786ed9cfecfccc5085c1e631eea1d/docs/search.rst).
* File endpoint [example](https://github.com/artefactual/archivematica-storage-service/blob/50d635878f2786ed9cfecfccc5085c1e631eea1d/docs/search.rst#file-search).
* Additional examples [inline in this PR](https://github.com/artefactual/archivematica-storage-service/pull/262#issuecomment-348039975).

**9 November 2017: Original PR**

- Locations, packages and files can all now be searched via GET requests to:
  - http://&lt;storage service URL&gt;/api/v2/search/location/
  - http://&lt;storage service URL&gt;/api/v2/search/package/
  - http://&lt;storage service URL&gt;/api/v2/search/file/    
- See documentation at [docs/search.rst](https://github.com/artefactual/archivematica-storage-service/blob/50d635878f2786ed9cfecfccc5085c1e631eea1d/docs/search.rst).
- Includes migration to File model.
- Uses Django REST Framework (new requirement).

For the requirements that motivated this work, see:

- https://wiki.archivematica.org/Research_data_management#METS_parsing
- https://projects.artefactual.com/issues/11581

Note: much of this work is due to PR https://github.com/artefactual/archivematica-storage-service/pull/89 of @mcantelon. Since that PR is quite old and required extensive rebasing, its commits were squashed and made into this PR, and its CR comments, where applicable, were transferred to this PR.

[...]

PR #357, opened 2018-05-11
WIP: create REST API

### LINKED ISSUE #369: Problem: HTTP REST API is inconsistent, under-documented, divorced from client code, and lacking needed endpoints
1. Inconsistent. The endpoints are defined via code in locations/api/urls.py and locations/api/resources.py and it is not very clear from looking at that code what is the pattern of HTTP method and path pattern (e.g., GET /locations/, GET /locations/\<UUID\>/, etc.) that each resource exposes and why. The pattern of endpoints for making CRUD requests against SS resources should be consistent so that consumers of the API can learn it and use it more easily.

2. Under-documented. The SS's HTTP (REST) API is documented on the Archivematica wiki at https://wiki.archivematica.org/Storage_Service_API. However, because the documentation is completely separate from the code that defines the API, it is an added burden to keep both in the same updated state.

3. Divorced from client code. There are several clients of the SS HTTP API. [Archivematica has its own Python module](https://github.com/artefactual/archivematica/blob/stable/1.7.x/src/archivematicaCommon/lib/storageService.py) for making calls against the SS API. The same is true for the [Archivematica Automation Tools](https://github.com/artefactual/automation-tools/blob/master/transfers/amclient.py), the [Archivematica Automated User Acceptance Tests (AMAUAT)](https://github.com/artefactual-labs/archivematica-acceptance-tests/blob/master/amuser/am_api_ability.py), and probably also for other tools not under Artefactual's stewardship. Whenever the SS API changes, the clients in all of these separate locations must all be individually changed to match, which is another burden that costs time and money. It would be great if client code could be generated from the server-defined API.

4. Needed endpoints are lacking. If a user wants to, say, [search over File models in the database](https://github.com/artefactual/archivematica-storage-service/pull/262) or [create new Location models via the API](https://github.com/artefactual/archivematica-storage-service/pull/368), then these new endpoints must be created on an ad hoc basis. We cannot anticipate all uses that users may want to put our APIs to. Therefore, we should expose standard CRUD endpoints on the resources that should be under user control, instead of waiting to respond to user requests or (worse) encouraging users to stop using the tool because the API is lacking.

Any work on the Storage Service that resolves this issue should be transferable to Archivematica and should help to address related issues, such as https://github.com/artefactual/archivematica/issues/1113.

jrwdunham:
🚧 🚧 🚧 

Fixes #369 

TODO:

- bring in the File indexing changes from [PR 262](https://github.com/artefactual/archivematica-storage-service/pull/262) but use the more advanced search API defined in this PR.
- Make consistent with the Swagger functionality introduced by [PR 300](https://github.com/artefactual/archivematica-storage-service/pull/300).
- Ultimately, use the pattern developed here to resolve [AM Issue 1113](https://github.com/artefactual/archivematica/issues/1113), i.e., adding API endpoints to Archivematica to facilitate the gathering of performance metrics.


jrwdunham:
Closing in favour of https://github.com/artefactual/archivematica-storage-service/pull/370

Issue #358, opened 2018-05-16
Problem: TypeError exception in async module


sevein:
This code is breaking:

https://github.com/artefactual/archivematica-storage-service/blob/2cc49f7debaeeb77bccb77e99c5844dc41b8b845/storage_service/locations/models/async.py#L34-L36

Traceback:

```
INFO      2018-05-15 17:06:01  locations.models.package:package:store_aip:603:  store_aip called in Package class of SS
INFO      2018-05-15 17:06:01  locations.models.space:space:move_rsync:463:  Moving from /var/archivematica/sharedDirectory/currentlyProcessing/DSpace_transfer_2-7e0bf0ad-d71c-4cc0-a443-71fdae492b65/DSpace_transfer_2-7e0bf0ad-d71c-4cc0-a443-71fdae492b65.7z to /var/archivematica/sharedDirectory/www/AIPsStore/7e0b/f0ad/d71c/4cc0/a443/71fd/ae49/2b65/DSpace_transfer_2-7e0bf0ad-d71c-4cc0-a443-71fdae492b65.7z
INFO      2018-05-15 17:06:02  locations.models.package:package:create_pointer_file:1235:  Pointer file constructed for 7e0bf0ad-d71c-4cc0-a443-71fdae492b65 is valid.
ERROR     2018-05-15 17:06:06  django.request.tastypie:resources:_handle_500:301:  Internal Server Error: /api/v2/async/5/
Traceback (most recent call last):
  File "/usr/share/archivematica/virtualenvs/archivematica-storage-service/local/lib/python2.7/site-packages/tastypie/resources.py", line 220, in wrapper
    response = callback(request, *args, **kwargs)
  File "/usr/share/archivematica/virtualenvs/archivematica-storage-service/local/lib/python2.7/site-packages/tastypie/resources.py", line 460, in dispatch_detail
    return self.dispatch('detail', request, **kwargs)
  File "/usr/share/archivematica/virtualenvs/archivematica-storage-service/local/lib/python2.7/site-packages/tastypie/resources.py", line 483, in dispatch
    response = method(request, **kwargs)
  File "/usr/share/archivematica/virtualenvs/archivematica-storage-service/local/lib/python2.7/site-packages/tastypie/resources.py", line 1362, in get_detail
    bundle = self.full_dehydrate(bundle)
[...]

PR #359, opened 2018-05-16
Async: convert buffer to str before pickling

sevein:
Connected to #358.

### DIFF
--- a/storage_service/locations/models/async.py
+++ b/storage_service/locations/models/async.py
@@ -1,11 +1,11 @@
 from __future__ import absolute_import
 # stdlib, alphabetical
-import cPickle
 import logging
 
 # Core Django, alphabetical
 from django.db import models
 from django.utils import six
+from django.utils.six.moves import cPickle as pickle
 from django.utils.translation import ugettext_lazy as _l
 
 __all__ = ('Async',)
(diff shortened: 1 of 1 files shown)


storage_service/locations/models/async.py
```diff
@@ -2,6 +2,7 @@
 # stdlib, alphabetical
 import cPickle
```
jrwdunham: In Py3 `cPickle` is just `pickle` right, so if we're going to anticipate supporting Python 3 here, should we import `pickle` as `cPickle`?

storage_service/locations/models/async.py
```diff
@@ -33,7 +34,14 @@ class Async(models.Model):
...
     @property
     def result(self):
-        return cPickle.loads(self._result)
+        result = self._result
+        # Convert memoryview (PY3) to str.
+        if sys.version_info[0] == 3 and isinstance(result, memoryview):
+            result = str(result)
+        # Convert buffer (PY2) to str.
+        if sys.version_info[0] == 2 and isinstance(result, buffer):
+            result = str(result)
```
jrwdunham: It looks like you could simplify using this by using `from django.utils import six`:

    if isinstance(result, six.memoryview):
        result = str(result)

jrwdunham (approved):
LGTM. One question and a suggestion for a possible improvement.

storage_service/locations/models/async.py
jrwdunham: Cf. https://docs.djangoproject.com/en/2.0/_modules/django/utils/six/

sevein:
@jrwdunham, ok done.

Issue #286, opened 2017-12-06
Problem: test user is always created


sevein:
Storage Service creates a `test` user automatically using the same password: `test`. It is created after the `postdb_sync` (aka `migrate` in recent versions of Django) signal is dispatched. Here's our callback:

https://github.com/artefactual/archivematica-storage-service/blob/2cc49f7debaeeb77bccb77e99c5844dc41b8b845/storage_service/common/models.py#L14-L21

As a result, the `test` user is made available when Storage Service is installed for the first time but also during release upgrades when the `manage.py migrate` command is executed. I think this is highly undesirable from a security point of view. If you find an instance of Storage Service in the wild there is a big chance that you can log in successfully using the default test user.

The user can manually delete the user after installing the application but the upgrade process will re-create it which is something that the user will not expect. The only effective way to void this behaviour today seems to be to change the password of the user. In the docs, we recommend the following:

> It is recommended that you also modify the test user and change the default password. This will also result in a new API key for the test user but that does not need to be copied to the clipboard in this instance.

I don't think that we're making clear very clear what's the actual behaviour.

My preference would be to stop creating the test user.

sevein:
If we fix this our test suite will break because it relies on the fact that the test user is automatically created. I fixed that in our RDSS fork, just by adding the following two objects to the `base.json` fixture:

```json
[
{
    "fields": {
        "username": "test",
        "first_name": "",
        "last_name": "",
        "is_active": true,
        "is_superuser": true,
        "is_staff": true,
        "last_login": null,
        "groups": [],
        "user_permissions": [],
        "password": "pbkdf2_sha256$20000$KtBsDJ1Efvp0$Hyvx9WvqBq2rNZX8DW1dpo8WdsEats9qc0U/NlK/Gq8=",
        "email": "test@test.com",
        "date_joined": "2017-12-06T21:16:42.772Z"
    },
    "model": "auth.user",
    "pk": 1
},
{
    "fields": {
        "user": 1,
        "key": "test",
        "created": "2017-12-06T21:16:42.838Z"
    },
    "model": "tastypie.apikey",
    "pk": 1
},
]
```

sevein:
Not happening in SS 0.11.

sromkey:
I believe @currmie is testing and documenting this.

currmie:
@sevein @sromkey - yep, I'm on it;  see 185 in archivematica-docs for corresponding doc issue

currmie:
[...]

PR #364, opened 2018-05-31
Delete create_testuser signal handler

sevein:
This commit removes the signal handler that was responsible for creating the
`test` superuser. It is still possible to create a superuser from the
command-line interface.

The `base` fixture file is updated so it includes a new superuser. Previously,
the tests were relying on user created by the `create_testuser` signal handler.

Connected to #286


jrwdunham (commented):
@sevein Can you modify #286 to describe *why* this is a problem? Is it an issue of not wanting the default behaviour being creation of a superuser with a very guessable password?

Also, I'm wondering if your fixtures solution will still work if we update Django in the future. That is, does the hashed value of the "test" password change if Django uses a different hashing algorithm?

Finally, are you saying that merging this will require making mods in all of our deployment tools (ansible, packages, docker) so that they become responsible for creating this "test" user?

sevein:
> @sevein Can you modify #286 to describe why this is a problem? Is it an issue of not wanting the default behaviour being creation of a superuser with a very guessable password?

Done. I hope it's clearer now.

> Also, I'm wondering if your fixtures solution will still work if we update Django in the future. That is, does the hashed value of the "test" password change if Django uses a different hashing algorithm?

I'm using `pbkdf2_sha256`which is still listed as a [supported hasher in Django v2.0](https://docs.djangoproject.com/en/2.0/topics/auth/passwords/#included-hashers). If this becomes a problem in the future I think it's safe to assume that the hasher identifier will raise some kind of exception and the test will break when we perform authentication, i.e. we wouldn't be seeing false positives I believe.

> Finally, are you saying that merging this will require making mods in all of our deployment tools (ansible, packages, docker) so that they become responsible for creating this "test" user?

I think that we're going to be solving this differently depending on the distribution type:

- Packages: we're updating the installation instructions (docs.archivematica.org) so the user can create the superuser manually.

- Ansible: I think that we can try to extend @mamedin's solution in https://github.com/artefactual-labs/ansible-archivematica-src/pull/162.

sevein:
@jrwdunham one clarification: I think that https://github.com/artefactual/archivematica-docs/pull/178 could be enough for now when it comes to explaining users how the installation process changes in debs/rpms/ansible without making actual changes in their sources.

Perhaps @mamedin's effort to automate the installation via Ansible (see https://github.com/artefactual-labs/ansible-archivematica-src/pull/162) may need some extra care once this is merged.

IMO this PR is ready to merge.

jrwdunham (approved):
LGTM

sevein:
2976d2526ddb3e680eeaca80b04c8dc5f80f8b0f

PR #366, opened 2018-05-31
Avoid unnecessary move_to_SS for checksum calculation

### LINKED ISSUE #365: Problem: pointer file creation causes unnecessary move operation
Now that the Storage Service (SS) creates pointer files it must calculate the checksum of the AIP in order to document the checksum in the pointer file. However, this checksum creation currently happens *after* the AIP has been moved out of the SS and into its destination space. See the `_create_pointer_file_write_to_disk` method and its call to `utils.generate_checksum`. This checksum calculation calls the `fetch_local_path` method which moves the AIP back *to* the Storage Service.

Not only is the above unnecessary and inefficient, it also breaks the functionality of spaces that do not implement the `move_to_storage_service` method, such as `DSpace`.

jrwdunham:
Instead of moving an AIP FROM the SS to an external space and then moving it back TO the SS again in order to calculate its checksum, this change allows us to calculate the checksum while we have the AIP locally.

This is connected to #365.

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -602,10 +602,10 @@ class Package(models.Model):
         """
         LOGGER.info('store_aip called in Package class of SS')
         v = self._store_aip_to_pending(origin_location, origin_path)
-        storage_effects = self._store_aip_to_uploaded(v, related_package_uuid)
+        storage_effects, checksum = self._store_aip_to_uploaded(v, related_package_uuid)
         self._store_aip_ensure_pointer_file(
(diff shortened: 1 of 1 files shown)


sevein (approved):

jrwdunham:
Notes relevant to QA:

- tested with:
  - local filesystem, compressed AIP
  - local filesystem, uncompressed AIP
- **not** tested with
  - local filesystem, reingested AIP
  - non-local filesystem, compressed AIP
  - non-local filesystem, umcompressed AIP
  - non-local filesystem, reingested AIP

PR #370, opened 2018-06-05
Improve HTTP API

### LINKED ISSUE #369: Problem: HTTP REST API is inconsistent, under-documented, divorced from client code, and lacking needed endpoints

jrwdunham:
Connected to #369 

# Summary

- Adds consistent REST (HTTP JSON) endpoints for almost every model in the Storage Service.
- Builds documentation into the API using a generated OpenAPI 3.0 YAML file and Swagger UI interface to same.
- Auto-generates Python client code which is self-documenting and has client-side validation.
- Introduces [abstractions](https://github.com/artefactual/archivematica-storage-service/pull/370/files#diff-1a6cae71b7e14bd63e08e093fc48f88bR62) and an [example](https://github.com/artefactual/archivematica-storage-service/pull/370/files#diff-ca2ffacd21802c84fba3d14b01ec11eeR87) of how to implement an OpenAPI-compliant custom endpoint `GET locations/{pk}/browse/` using Python data structures.

This PR adds a new set of versioned API endpoints (under the `beta` namespace) which address #369 by being:

- [consistent](https://github.com/artefactual/archivematica-storage-service/blob/dev/issue-369-improve-http-api/storage_service/locations/api/beta/__init__.py#L18-L31): same HTTP method + path regex for all endpoints or operations of a given type; all resources expose the same *read/write* or *read-only* operations.
- thoroughly documented using OpenAPI 3.0 spec—see `/api/beta/`—and a Swagger UI interface—see `/api/beta/doc/`—to same;
- able to generate API client code—see `/api/beta/client/`—which is generated by [clientbuilder.py](https://github.com/artefactual/archivematica-storage-service/blob/dev/issue-369-improve-http-api/storage_service/locations/api/beta/remple/clientbuilder.py); and
- introduces needed endpoints, e.g., advanced search over all exposed resources, DELETE, PUT (update), etc.

# How to Understand OpenAPI 3.0

OpenAPI suffers from too much documentation. I have found this one to be good:

- https://swagger.io/docs/specification/about/

# How to Understand this PR

- The API endpoints introduced here are configured in [storage_service/locations/api/beta/](https://github.com/artefactual/archivematica-storage-service/tree/9105797acdd10c11cf33b6a7e1c9950f2375662b/storage_service/locations/api/beta)
  - [\_\_init\_\_.py](https://github.com/artefactual/archivematica-storage-service/blob/9105797acdd10c11cf33b6a7e1c9950f2375662b/storage_service/locations/api/beta/__init__.py) defines the `resources` dict that tells Remple which resources with which endpoints to expose.
  - resources/ defines the resource classes (which point to Django models to infer read schemata and Formencode schemata for generating mutate schemata).
    - resources.py — simple resources with little configuration (ie no custom endpoints)
    - locations.py — a module of its own because it has a custom endpoint: `browse`
  - [schemata.py](https://github.com/artefactual/archivematica-storage-service/blob/9105797acdd10c11cf33b6a7e1c9950f2375662b/storage_service/locations/api/beta/schemata.py) defines the ([Formencode](http://www.formencode.org/en/latest/)) schemata for viewing, creating and updating resources.
- [Remple](https://github.com/artefactual/archivematica-storage-service/blob/dev/issue-369-improve-http-api/storage_service/locations/api/beta/remple/__init__.py) is the mini REST framework introduced here. If this PR is merged, we will probably want to move Remple into its own repo so that [Archivematica](https://github.com/artefactual/archivematica) can use it also.
  - [resources.py](https://github.com/artefactual/archivematica-storage-service/blob/dev/issue-369-improve-http-api/storage_service/locations/api/beta/remple/resources.py) provides the base classes for defining resources.
  - [querybuilder.py](https://github.com/artefactual/archivematica-storage-service/blob/dev/issue-369-improve-http-api/storage_service/locations/api/beta/remple/querybuilder.py) holds the logic for transforming JSON/Python data structures into Django ORM filters.
  - [routebuilder.py](https://github.com/artefactual/archivematica-storage-service/blob/dev/issue-369-improve-http-api/storage_service/locations/api/beta/remple/routebuilder.py) takes a `resources` config dict and can return a list of Django `url()` instances that can be `include`d in `urlpatterns`.
  - [openapi.py](https://github.com/artefactual/archivematica-storage-service/blob/dev/issue-369-improve-http-api/storage_service/locations/api/beta/remple/openapi.py) can generate an OpenAPI data structure given a configured Remple API and spit out an OpenAPI 3.0-conformant YAML file that can be used to generate the Swagger UI JavaScript/HTML app as well as the Python client code.
  - [clientbuilder.py](https://github.com/artefactual/archivematica-storage-service/blob/dev/issue-369-improve-http-api/storage_service/locations/api/beta/remple/clientbuilder.py) is a Python module that must be modified to contain an OpenAPI data structure so that it can define (at runtime) a set of Python classes that constitute a client for the SS API.

# Questions

- Would it be better to use [TastyPie](http://django-tastypie.readthedocs.io/en/latest/), [Django Rest Framework](http://www.django-rest-framework.org/) or [APIStar](https://github.com/encode/apistar/blob/master/docs/api-guide/api-schemas.md) (Py3 only, but has OpenAPI spec generation built-in) instead of the custom-built Remple mini REST framework [Remple](https://github.com/artefactual/archivematica-storage-service/blob/dev/issue-369-improve-http-api/storage_service/locations/api/beta/remple/__init__.py) introduced here? At a high level, the benefits of rolling our own API mini framework allows us more control as well as the ability to be on the bleeding edge when it comes to the OpenAPI spec (cf. codegen's unavailability in OpenAPI v. 3.0). The benefits of using an existing framework are a lower maintenance burden for us and the pros of using a battle-tested code base with buy-in from disparate stakeholders.
- Can existing code generation tools be used to build the Python client/SDK instead of rolling our own in clientbuilder.py? From my research, the answer was "No, because these tools have not caught up to OpenAPI v. 3.0 yet." See [Swagger Codegen](https://github.com/swagger-api/swagger-codegen). My preference would be to use clientbuilder.py and switch to a standard OpenAPI code generation tool when such tools catch up to the spec, an option that is opened up for us by virtue of using OpenAPI.


sevein:
I think that you've accidentally checked in `storage_service/.pytest_cache/v/cache/*`.

jrwdunham:
Thanks @sevein: removed those

jambun:
Hi @jrwdunham. We have spent some time going through this and our overall impression is positive. We like the approach and were able to understand how you had done things and why.

Using clientbuilder as a stopgap makes sense. Hopefully swagger support isn’t too far away. We noticed this: https://github.com/openapitools/openapi-generator.

Our understanding is a bit abstract at this stage - we’ve only eyeballed the code. We will be able to give more detailed feedback after using it.

storage_service/locations/api/beta/remple/openapi.py
```diff
@@ -0,0 +1,1573 @@
...
+                    ('description', custom_endpoint.description),
+                ])
+            }
+            for attr in ('parameters', 'request_body', 'responses', 'tags'):
+                val = getattr(custom_endpoint, attr)
+                if val:
+                    path_dict[http_method][attr] = val
+            paths[path] = path_dict
+
+    def _get_search_request_body_examples(self, resource_name):
```
jrwdunham, line 228: This is specific to the Archivematica Storage Service so it should be removed or moved outside of Remple. Futhermore, it seems that it is not even being used.

storage_service/locations/api/beta/remple/openapi.py
```diff
@@ -0,0 +1,1573 @@
...
+                                                        'relation': '>',
+                                                        'value': 512}]}]}}}
+            return {'ArraySearchOverPackagesExample': array_search_example,
+                    'ObjectSearchOverPackagesExample': object_search_example}
+
+    def _set_search_paths(self, paths, resource_name, resource_cls, pk_patt,
+                          rsrc_collection_name):
+        for action in ('search_post', 'new_search'):
+            http_method = {'search_post': 'post', 'new_search': 'get'}.get(
+                action, 'search')
```
jrwdunham, line 268: Document or remove the implicit usage here of the non-standard HTTP method `SEARCH`.

storage_service/locations/api/beta/remple/openapi.py
```diff
@@ -0,0 +1,1573 @@
...
+                ('description', description),
+                ('operationId', operation_id),
+            ])
+            if parameters:
+                path_dict[http_method]['parameters'] = parameters
+            if request_body:
+                path_dict[http_method]['requestBody'] = request_body
+            path_dict[http_method]['responses'] = getattr(
+                self, responses_meth)(resource_name)
+            path_dict[http_method]['tags'] = [rsrc_collection_name]
```
jrwdunham, line 303: It should be possible to remove some of the duplication here with `_set_crud_paths`.

storage_service/locations/api/beta/remple/openapi.py
```diff
@@ -0,0 +1,1573 @@
...
+        return '{}{}'.format(
+            self.path_prefix, self.get_api_version_slug())
+
+    def _get_api_servers(self):
+        """Return a list of OrderedDicts for the top-level ``servers``
+        attribute.
+        """
+        return [
+            OrderedDict([
+                ('url', self.get_dflt_server_path()),
```
jrwdunham, line 686: TODO: should we be able to use info in `django_settings` to specify a true URL here instead of just a path?

storage_service/locations/api/beta/remple/openapi.py
```diff
@@ -0,0 +1,1573 @@
...
+        """Return an OpenAPI ``schemas`` OrderedDict.
+
+        It contains a "View", a "Create", and an "Update" schema for each
+        resource in ``self.resources``, e.g., ``LocationView``,
+        ``LocationCreate``, and ``LocationUpdate``. The create and update
+        schemata are only included if the resources is not read-only. The view
+        schema is constructed by introspecting the Django model, while the
+        create and update schemata are constructed by introspecting the relevant
+        Formencode schemata attached as class attributes on the resource class.
+        """
```
jrwdunham, line 810: TODO: it seems that OpenAPI's readOnly and writeOnly boolean property attributes could be used here. See https://swagger.io/docs/specification/data-models/data-types/, in particular the `Read-Only and Write-Only Properties` section.

storage_service/locations/api/beta/remple/openapi.py
```diff
@@ -0,0 +1,1573 @@
...
+        return create_schema_name, create_schema
+
+    def _get_update_schema(self, resource_name, resource_cfg, read_schema):
+        """Return an update schema for the resource named ``resource_name``.
+
+        The update schema describes what is needed to update an instance of the
+        input-named resource.
+
+        TODO: should every parameter in an update request be optional, given
+        that the resource being updated is presumably valid?
```
jrwdunham, line 1200: No. This comment is misleading. Update via Remple has the semantics of `PUT` not `PATCH`. A full representation of the post-update state must be sent in the update request.

storage_service/locations/api/beta/remple/openapi.py
```diff
@@ -0,0 +1,1573 @@
...
+        schema['properties'] = schema_properties
+        if required_fields:
+            schema['required'] = required_fields
+        return schema
+
+    def _get_read_schema(self, resource_name, resource_cfg):
+        """Return a read schema for the resource named ``resource_name``.
+
+        The read schema describes what is returned by the server as a
+        representation of an instance of the input-named resource.
```
jrwdunham, line 1249: Docstring should indicate that this is done by inspecting the Django model.

storage_service/locations/api/beta/remple/openapi.py
```diff
@@ -0,0 +1,1573 @@
...
+        for index, part in enumerate(parts):
+            part_int = int(part)
+            if part_int:
+                new_parts.append(part)
+            else:
+                parts_to_right = parts[index + 1:]
+                non_empty_ptr = [p for p in parts_to_right if int(p)]
+                if non_empty_ptr:
+                    new_parts.append(part)
+        return 'v{}'.format('_'.join(new_parts))
```
jrwdunham, line 1326: The above is only needed if we want to support semver-compliant API versions, which may not be a good idea.

storage_service/locations/api/beta/remple/openapi.py
```diff
@@ -0,0 +1,1573 @@
...
+    if choices:
+        return [c[0] for c in choices]
+    return None
+
+
+def get_default(field):
+    return getattr(field, 'default', NOT_PROVIDED)
+
+
+def get_required(field, default):
```
jrwdunham, line 1378: Fix function name since it returns a boolean indicating whether a field is required and whether it is null(able).

helrond:
This is a huge step forward in making the Storage Service API consistent, predictable and documented. Thank you! I haven't had a chance to spin this up locally yet, but I'm planning to do that in the next week or so and may have more specific feedback after that.

I am not a real developer, and have a very rudimentary knowledge of the Archivematica codebase, however from my perspective I'm not sure it makes sense to implement a custom REST framework. It seems like maintenance for Archivematica is already challenging enough that alleviating some of that burden would outweigh the benefits of OpenAPI 3.0 compliance, especially since as you note the tooling isn't really there yet.  But perhaps I don't fully understand the benefits of OpenAPI 2.x versus 3.0.

PR #379, opened 2018-06-20
Correct Dataverse issues discovered post proof-of-concept (PoC)

### LINKED ISSUE #352: Problem: Dataverse create/edit space page is not rendering correctly
The form component after the staging path when asked to re-render, e.g. submitting the form once correctly or otherwise results in the following output:

image

The HTML when we view source looks as follows:
```
<form action="/spaces/create/" method="post">
    <input type='hidden' name='csrfmiddlewaretoken' value='GDfsIWLVjGdtH6EFnDSxiJl3tbuYdQbq' />
    <p>
        <label for="id_space-access_protocol">Access protocol:</label>
        <select id="id_space-access_protocol" name="space-access_protocol">
            <option value="">---------</option>
            <option value="ARKIVUM">Arkivum</option>
            <option value="DV" selected="selected">Dataverse</option>
            <option value="DC">DuraCloud</option>
            <option value="DSPACE">DSpace via SWORD2 API</option>
            <option value="FEDORA">FEDORA via SWORD2</option>
            <option value="GPG">GPG encryption on Local Filesystem</option>
            <option value="FS">Local Filesystem</option>
            <option value="LOM">LOCKSS-o-matic</option>
            <option value="NFS">NFS</option>
            <option value="PIPE_FS">Pipeline Local Filesystem</option>
            <option value="SWIFT">Swift</option>
        </select> <span class="helptext">How the space can be accessed.</span></p>
    <p>
        <label for="id_space-size">Size:</label>
        <input id="id_space-size" max="9223372036854775807" min="-9223372036854775808" name="space-size" type="number" /> <span class="helptext">Size in bytes (optional)</span></p>
    <p>
        <label for="id_space-path">Path:</label>
        <textarea cols="40" id="id_space-path" name="space-path" rows="10">
            /var/archivematica/storage_service/dataverse/staging/</textarea> <span class="helptext">Absolute path to the space on the storage service machine.</span></p>
    <p>
        <label for="id_space-staging_path">Staging path:</label>
        <textarea cols="40" id="id_space-staging_path" name="space-staging_path" rows="10">
            /var/archivematica/storage_service/dataverse/staging/</textarea> <span class="helptext">Absolute path to a staging area.  Must be UNIX filesystem compatible, preferably on the same filesystem as the path.</span></p>
    <div id="protocol_form">&lt;ul class=&quot;errorlist&quot;&gt;&lt;li&gt;This field is required.&lt;/li&gt;&lt;/ul&gt; &lt;p&gt;&lt;label for=&quot;id_protocol-host&quot;&gt;Host:&lt;/label&gt; &lt;input id=&quot;id_protocol-host&quot; maxlength=&quot;256&quot; name=&quot;protocol-host&quot; type=&quot;text&quot; /&gt; &lt;span class=&quot;helptext&quot;&gt;xxx&lt;/span&gt;&lt;/p&gt; &lt;ul class=&quot;errorlist&quot;&gt;&lt;li&gt;This field is required.&lt;/li&gt;&lt;/ul&gt; &lt;p&gt;&lt;label for=&quot;id_protocol-api_key&quot;&gt;API key:&lt;/label&gt; &lt;input id=&quot;id_protocol-api_key&quot; maxlength=&quot;50&quot; name=&quot;protocol-api_key&quot; type=&quot;text&quot; /&gt; &lt;span class=&quot;helptext&quot;&gt;xxx&lt;/span&gt;&lt;/p&gt; &lt;ul class=&quot;errorlist&quot;&gt;&lt;li&gt;This field is required.&lt;/li&gt;&lt;/ul&gt; &lt;p&gt;&lt;label for=&quot;id_protocol-agent_name&quot;&gt;Agent name:&lt;/label&gt; &lt;input id=&quot;id_protocol-agent_name&quot; maxlength=&quot;50&quot; name=&quot;protocol-agent_name&quot; type=&quot;text&quot; /&gt; &lt;span class=&quot;helptext&quot;&gt;xxx&lt;/span&gt;&lt;/p&gt; &lt;ul class=&quot;errorlist&quot;&gt;&lt;li&gt;This field is required.&lt;/li&gt;&lt;/ul&gt; &lt;p&gt;&lt;label for=&quot;id_protocol-agent_type&quot;&gt;Agent type:&lt;/label&gt; &lt;input id=&quot;id_protocol-agent_type&quot; maxlength=&quot;50&quot; name=&quot;protocol-agent_type&quot; type=&quot;text&quot; /&gt; &lt;span class=&quot;helptext&quot;&gt;xxx&lt;/span&gt;&lt;/p&gt; &lt;ul class=&quot;errorlist&quot;&gt;&lt;li&gt;This field is required.&lt;/li&gt;&lt;/ul&gt; &lt;p&gt;&lt;label for=&quot;id_protocol-agent_identifier&quot;&gt;Agent identifier:&lt;/label&gt; &lt;input id=&quot;id_protocol-agent_identifier&quot; maxlength=&quot;256&quot; name=&quot;protocol-agent_identifier&quot; type=&quot;text&quot; /&gt; &lt;span class=&quot;helptext&quot;&gt;xxx&lt;/span&gt;&lt;/p&gt; &lt;div class=&quot;alert&quot;&gt;Integration with Dataverse is currently a beta feature&lt;/div&gt;</div>
    <input type="submit" value="Create Space" class='btn btn-primary' />
</form>
```

### LINKED ISSUE #355: Problem: Can't filter Dataverse datasets by query-string
The Scholar's Portal team set up a test dataverse instance for us within their development dataverse. The idea discussed at the May 1 meeting was that we would connect to the "Archivematica Test Dataverse", which sits within the "root" dataverse. 

I create a new space & location (transfer source) using the development dataverse URL and the API Key from the Archivematica test user that I created. 

I can now view datasets from within the Archivematica Dashboard. However, I can see all datasets. I set the relative path in the location to "Archivematica", which should limit the query of datasets to those with Archivematica in the ... title? (not entirely sure how the dataverse search API is meant to work). 

I get the same set of 9 datasets returned regardless of whether "Archivematica" or "archivematica" or nothing is in the relative path. 

This suggests to me the functionality of putting keywords in the relative path is not working. But even if it did, it's not clear that this is the best way to restrict the datasets shown in the Dashboard. So there are perhaps several approaches that could be taken here: 

- confirm (or make sure) the search parameter within _relative path_ works
- find a way to restrict the location to a particular dataverse (e.g. our Archivematica Test Dataverse) rather than the root dataverse
- find a way to restrict the location to a particular user (e.g. Archivematica Test User)

It's also important to note that in the Dashboard, what we see are 9 datasets. But the root dataverse has 100 datasets. There doesn't appear to be any way to view any other than the first 9. This should probably be a separate issue but until we define the expected behaviour I'll leave this as a note here.

ross-spencer:
This PR brings the Dataverse proof-of-concept work back inline with the most recent updates to the storage service. The two main issues experienced were around the incorrect manipulation of a Django form that enabled the selection of Dataverse as a transfer source. Secondly, the proof of concept co-opted the relative path field of the Locations model to create a filter that restricted the data-sets available as a transfer source. The issues were #352 and #355 respectively.

* Resolves #352 
* Resolves #355

### DIFF
--- a/storage_service/locations/forms.py
+++ b/storage_service/locations/forms.py
@@ -103,14 +103,6 @@ class DataverseForm(forms.ModelForm):
         model = models.Dataverse
         fields = ('host', 'api_key', 'agent_name', 'agent_type', 'agent_identifier')
 
-    def as_p(self):
-        # Add a warning to the Dataverse-specific section of the form
-        # FIXME this may not be the best way to add Space-specific warnings
-        content = super(DataverseForm, self).as_p()
-        content += '\n<div class="alert">{}</div>'.format(
-            _('Integration with Dataverse is currently a beta feature'))
-        return content
-
 
 class DuracloudForm(forms.ModelForm):
     class Meta:
--- a/storage_service/locations/forms.py
+++ b/storage_service/locations/forms.py
@@ -5,6 +5,7 @@ from django import forms
 import django.utils
 import django.core.exceptions
 from django.db.models import Count
+from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext as _, ugettext_lazy as _l
 
 from common import gpgutils
@@ -39,14 +40,14 @@ class DisableableSelectWidget(forms.Select):
     def render_option(self, selected_choices, option_value, option_label):
         option_value = django.utils.encoding.force_text(option_value)
         if option_value in selected_choices:
-            selected_html = django.utils.safestring.mark_safe(' selected="selected"')
+            selected_html = mark_safe(' selected="selected"')
             if not self.allow_multiple_selected:
                 # Only allow for a single selection.
                 selected_choices.remove(option_value)
         else:
             selected_html = ''
         if option_value in self.disabled_choices:
-            disabled_html = django.utils.safestring.mark_safe(' disabled="disabled"')
+            disabled_html = mark_safe(' disabled="disabled"')
         else:
             disabled_html = ''
         return django.utils.html.format_html(
@@ -102,6 +103,13 @@ class DataverseForm(forms.ModelForm):
     class Meta:
         model = models.Dataverse
         fields = ('host', 'api_key', 'agent_name', 'agent_type', 'agent_identifier')
+
+    def as_p(self):
+        # Add a warning to the Dataverse-specific section of the form
+        content = super(DataverseForm, self).as_p()
+        content += '\n<div class="alert">{}</div>'.format(
+            ('Integration with Dataverse is currently a beta feature'))
+        return mark_safe(content)
 
 
 class DuracloudForm(forms.ModelForm):
--- a/storage_service/locations/models/space.py
+++ b/storage_service/locations/models/space.py
@@ -141,7 +141,7 @@ class Space(models.Model):
     OBJECT_STORAGE = {DATAVERSE, DSPACE, DURACLOUD, SWIFT, S3}
     ACCESS_PROTOCOL_CHOICES = (
         (ARKIVUM, _l('Arkivum')),
-        (DATAVERSE, _l('Dataverse')),
+        (DATAVERSE, _l('Dataverse (Beta)')),
         (DURACLOUD, _l('DuraCloud')),
         (DSPACE, _l('DSpace via SWORD2 API')),
         (FEDORA, _l("FEDORA via SWORD2")),
--- a/storage_service/locations/models/space.py
+++ b/storage_service/locations/models/space.py
@@ -141,7 +141,7 @@ class Space(models.Model):
     OBJECT_STORAGE = {DATAVERSE, DSPACE, DURACLOUD, SWIFT, S3}
     ACCESS_PROTOCOL_CHOICES = (
         (ARKIVUM, _l('Arkivum')),
-        (DATAVERSE, _l('Dataverse (Beta)')),
+        (DATAVERSE, _l('Dataverse')),
         (DURACLOUD, _l('DuraCloud')),
         (DSPACE, _l('DSpace via SWORD2 API')),
         (FEDORA, _l("FEDORA via SWORD2")),
--- a/storage_service/locations/models/dataverse.py
+++ b/storage_service/locations/models/dataverse.py
@@ -22,6 +22,7 @@ from .location import Location  # noqa: E402
 
 class Dataverse(models.Model):
     space = models.OneToOneField("Space", to_field="uuid")
+
     host = models.CharField(
         max_length=256,
         verbose_name=_l("Host"),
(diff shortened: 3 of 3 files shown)


storage_service/locations/models/space.py
```diff
@@ -141,7 +141,7 @@ class Space(models.Model):
     OBJECT_STORAGE = {DATAVERSE, DSPACE, DURACLOUD, SWIFT, S3}
     ACCESS_PROTOCOL_CHOICES = (
         (ARKIVUM, _l('Arkivum')),
-        (DATAVERSE, _l('Dataverse')),
+        (DATAVERSE, _l('Dataverse (Beta)')),
```
sevein: This produces a new database migration which I think is not ideal. You may have noticed that Travis is complaining because your PR does not include the new migration.

storage_service/locations/forms.py
```diff
@@ -103,14 +103,6 @@ class Meta:
         model = models.Dataverse
         fields = ('host', 'api_key', 'agent_name', 'agent_type', 'agent_identifier')
 
-    def as_p(self):
-        # Add a warning to the Dataverse-specific section of the form
-        # FIXME this may not be the best way to add Space-specific warnings
-        content = super(DataverseForm, self).as_p()
-        content += '\n<div class="alert">{}</div>'.format(
-            _('Integration with Dataverse is currently a beta feature'))
-        return content
```
sevein: Maybe this is fixable by using [`mark_safe`](django.utils.safestring.mark_safe) again?

sevein (commented):
@ross-spencer, is the use of black a plan or you're just experimenting?

storage_service/locations/forms.py
ross-spencer: `mark_safe()` didn't come up in the literature I was reading as part of this PR but having tested it - it looks to work. I agree this is the right approach. The `(Beta)` component of this work is temporary so a migration is not ideal.

ross-spencer:
@sevein re: Python 'black' I felt that this was a good class to try it on as it had relatively few lines. It moved the code-base one-step further in the right direction re: formatting and CI checks. I can close this PR and open a new one with just the `mark_safe()` change (and then squish everything again later), but what do you think? If you're happy to take a look at the black changes too, I'd like to keep them.

sevein:
I don't think that you have to close this PR, i.e. `51aa196` fixes up `a4b77c0` and `3bd614c` can be left as it is. You mentioned in the description that you were trying to address #352 and #355 and that's still the case right?

But I would do the black work separately. I like what black could bring to our projects but it's worth starting with a new issue describing the problem that we're trying to solve, e.g. deterministic formatting, ceding control to the machine, etc...  then I think a pull request could bring all that in single commit, so we don't have formatting changes scatted across the history. I'd also like to see some level of integration in tox (`black --check`) to reject code that isn't "blacked" and we still need to decide which line length we would be using.

ross-spencer:
> then I think a pull request could bring all that in single commit, so we don't have formatting changes scatted across the history. 

@sevein for the sake of expediency, then removing the black work from here is okay, but the idea that formatting cannot be completed when a developer is looking at other issues seems like we'd be missing opportunities to be able to ask not just ourselves, but the community to do the same for us while their heads are in the middle of a class or a function, or other unit of code. I think here we have a good example of a discrete class - the Dataverse work is slowly being turned into production so there'll be a lot of testing happening, plus the current controls pass. For me its like the strangler pattern. If we control it by initially having individual commits that state only the formatting has changed to make it easy for code-review then I think we're doing enough to be able to tackle modules one-by-one without new issues and new PRs. 

IDK perhaps this needs to be drawn out somewhere where we discuss that one type of commit is functional. Another is for style, and we need two separate commits if we want to be able to leverage some of what I discuss above. 

RE: Ceding control, black is just a formatting tool, so it is the equivalent of me walking through the class and splitting lines, adding brackets, quotation marks, myself. My doing it, flake8, and a small-CR process, seems to be control enough?

(I agree on the CI part entirely as that then adds the extra control across the code-base in-time. If we add it now though CI will fail across a number of our repositories)

storage_service/locations/forms.py
```diff
@@ -105,11 +106,10 @@ class Meta:
 
     def as_p(self):
         # Add a warning to the Dataverse-specific section of the form
-        # FIXME this may not be the best way to add Space-specific warnings
         content = super(DataverseForm, self).as_p()
         content += '\n<div class="alert">{}</div>'.format(
-            _('Integration with Dataverse is currently a beta feature'))
```
sevein, line 111: `_` should probably not be removed.

sevein:
Replaced by https://github.com/artefactual/archivematica-storage-service/pull/398.

Issue #312, opened 2018-01-31
Problem: gunicorn gevent worker class and sword API multiprocessing are incompatible


jrwdunham:
Using gunicorn with worker class `'gevent'` is incompatible with usage of Python's multiprocessing package, which is used in the sword API logic at https://github.com/artefactual/archivematica-storage-service/blob/stable/0.11.x/storage_service/locations/api/sword/helpers.py#L148. For reference, see

- https://github.com/benoitc/gunicorn/issues/1367#issuecomment-268439104
- https://stackoverflow.com/questions/8678307/gevent-monkeypatching-breaking-multiprocessing

Related to https://github.com/artefactual/archivematica/issues/708.
Refs https://projects.artefactual.com/issues/11642.

The workaround is to set the gunicorn worker class to `'sync'` when SS's sword API is being used:

    $ export SS_GUNICORN_WORKER_CLASS=sync

However, a better fix is probably required here, i.e., some change to storage_service/locations/api/sword/helpers.py that can make concurrent HTTP requests without using `multiprocessing.Process`.

@sevein @jhsimpson any thoughts?

sevein:
It could be implemented with threads too. Or we could wait until the work being done at https://github.com/hudmol/archivematica-storage-service/commit/async is completed.

jrwdunham:
Note: for AM 1.7 we simply documented this issue and described a deployment workaround. See https://github.com/artefactual/archivematica-docs/pull/75.

sevein:
Let's try to use the new `async` module.

sevein:
> Note: for AM 1.7 we simply documented this issue and described a deployment workaround. See artefactual/archivematica-docs#75.

Question: if #383 is merged how do you think we should update the docs? Should we make a distinction between SS v0.11 and SS v0.12 or should we just update archivematica-docs's branch 1.8 and assume there that the user is using SS v0.12?

sevein:
To test this we need to create a space of type "FEDORA via SWORD2". See https://wiki.archivematica.org/Storage_Service#FEDORA_Deposit for more details.

The deposits should work regardless the value used in the configuration attribute `SS_GUNICORN_WORKER_CLASS`. In fact this configuration attribute has been removed.

currmie:
Hi  @sevein , has someone already created a space of type "FEDORA via SWORD2" that I could use for QA/testing?

sevein:
[...]

PR #383, opened 2018-06-21
SWORD: use AsyncManager in background tasks

sevein:
Our implementation of SWORD uses processes to download contents asynchronously.
We do this via the `multiprocessing` package which stopped working when we
moved to gevent. This commit replaces the use of `multiprocessing` with our own
`AsyncManager` which uses threads instead.

This is connected to #312.

### DIFF
--- a/storage_service/locations/api/sword/helpers.py
+++ b/storage_service/locations/api/sword/helpers.py
@@ -3,7 +3,6 @@ import cgi
 import datetime
 import logging
 import os
-from multiprocessing import Process
 import shutil
 import tempfile
 import time
@@ -21,6 +20,7 @@ import requests
 
 # This project, alphabetical
 from locations import models
+from locations.models.async_manager import AsyncManager
 from common.utils import generate_checksum
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


storage_service/locations/api/sword/helpers.py
```diff
@@ -145,8 +145,7 @@ def spawn_download_task(deposit_uuid, objects, subdir=None):
     """
     Spawn an asynchrnous batch download
     """
-    p = Process(target=_fetch_content, args=(deposit_uuid, objects, subdir))
-    p.start()
+    AsyncManager.run_task(_fetch_content, (deposit_uuid, objects, subdir))
```
jrwdunham: @sevein are you sure you should be passing a tuple as the second argument to `run_task` here? Doing it the way it's done here, I think it will end up passing a tuple which contains a tuple which contains `deposit_uuid`, etc. to the `args` kwarg of `threading.Thread`. The signature of `run_task` is `def run_task(task_fn, *args, **kwargs)`.

    >>> def x(a, *args, **kwargs):
    ...   print(a)
    ...   print(args)
    ...   print(kwargs)
    >>> x(1, (2, 3))
    1
    ((2, 3),)
    {}

sevein: You're right, thanks. It's fixed now.

jrwdunham (approved):
Code looks good to me. I haven't tested it in context, but I think we can leave that to QA.

PR #386, opened 2018-06-27
Fix dataverse location browsing

jrwdunham:
Browsing the root of a Dataverse location will return all of the datasets fetchable from the Host. Browsing a dataset will return all of the files in that dataset.

Connected to https://github.com/artefactual/archivematica/issues/1085

**Note:** the branch of this PR is built on top of dev/issue-352-dataverse-template-rendering at commit 90fce1e.

### DIFF
--- a/storage_service/locations/fixtures/dataverse2.json
+++ b/storage_service/locations/fixtures/dataverse2.json
@@ -1,0 +1,42 @@
+[
+{
+    "fields": {
+        "host": "demodv.scholarsportal.info",
+        "api_key": "51c64df4-abd7-4613-af21-6a68715dca92",
+        "space": "d18a2763-60f0-4273-b30c-caa5d14d6d32",
+        "agent_name": "Archivematica Test Dataverse",
+        "agent_type": "TestOrganisation",
+        "agent_identifier": "https://demodv.scholarsportal.info/dataverse/archivematica"
+    },
+    "model": "locations.dataverse",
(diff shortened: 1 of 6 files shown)


ross-spencer:
Hi @jrwdunham the request to CR this is noted. I expect to look at this at the beginning of next week, and again, appreciate the effort to look at this asap for the OCUL work. 

I notice some unit tests are failing, I can't see why this is right now. It seems to be across a number of source types, so I can't see this being introduced by your code, but can investigate more when I come to CR.

jrwdunham:
@ross-spencer I'm hoping those unit test failures are due to the VCR.py and PyYAML fiasco that was affecting all of our tests...

jrwdunham:
@ross-spencer I've fixed the dataverse browse tests to pass with my mods to that method.

ross-spencer:
Thanks @jrwdunham, watching from afar here but looks like good work to me. I have just made a note in Slack for @joel-simpson - all being well, I hope to review this the week beginning 16 June.

ross-spencer:
Thanks for the help @jrwdunham. This work has been wrapped up into https://github.com/artefactual/archivematica-storage-service/pull/398 to enhance use the API calls introduced here in the transfer browser toggle.

ross-spencer:
RE: Deleting the branch. I don't know the best process to follow, but suggest we do this following the conclusion of the Dataverse phase 2 work?

PR #388, opened 2018-06-29
DSpace REST API support

J4bbi:
This version only support DSpace 6.x REST API, there are differences between each major version I think.

This space assumes you want to deposit both AIP and DIP to DSpace and create a digital object in ArchiveSpace that links to the DIP handle record. It will extract DSpace collection ids and ArchivesSpace archival object id from metadata.json if it finds the custom fields: dspace_dip_collection, dspace_aip_collection or archivesspace_dip_collection or else use default values set up when creating the space.

The AIP is assumed to be compressed but the DIP is uncompressed.

Note that the migration file 0018_dspace_rest.py was manually stitched together by me as I has done 2 migrations in my dev environ, so I haven't run it.

Connected to #391


storage_service/locations/forms.py
```diff
@@ -124,6 +124,15 @@ class Meta:
         fields = ('sd_iri', 'user', 'password', 'metadata_policy', 'archive_format')
 
 
+class DSpaceRESTForm(forms.ModelForm):
+    class Meta:
+        model = models.DSpaceREST
+        fields = ('rest_url', 'user', 'password', 'archive_format', 'dspace_dip_collection',
+                  'dspace_aip_collection', 'archivesspace_url',
+                  'archivesspace_user', 'archivesspace_password', 'archivesspace_repository',
+                  'archivesspace_archival_object')
```
jrwdunham: Is it intentional that there is no `password` field in `models.SpaceREST`?

jrwdunham: ::Nitpick alert:: I think there is enough complexity in the number of fields for this model that it would improve readability if all the DSpace-relevant attributes had a short common prefix like `ds_` and the ArchivesSpace all had theirs shortened from the `archivesspace_` prefix to `as_`. If you don't want to do that and want to treat DSpace as the default, then I would suggest removing the `dspace_` prefix from `dspace_aip_collection` and `dspace_dip_collection`.

storage_service/locations/migrations/0018_dspace_rest.py
```diff
@@ -0,0 +1,51 @@
...
+
+    dependencies = [
+        ('locations', '0017_gpg_space_minor_migration'),
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='DSpaceREST',
+            fields=[
+                ('id', models.AutoField(verbose_name='ID', serialize=False, auto_created=True, primary_key=True)),
```
jrwdunham, line 17: ~~All of the `verbose_name` and `help_text` values here need to be wrapped in translation `ugettext as _, ugettext_lazy as _l`. See, for example, https://github.com/artefactual/archivematica-storage-service/blob/stable/0.11.x/storage_service/locations/models/dspace.py#L42.~~

Oh, I see you did that in the model. Right, I think this can be ignored.

jrwdunham: Remove blank text after `'REST URL '` in following line.

storage_service/locations/migrations/0018_dspace_rest.py
```diff
@@ -0,0 +1,51 @@
...
+                ('rest_url', models.URLField(help_text='URL of the REST API. E.g. http://demo.dspace.org/rest', max_length=256, verbose_name='REST URL ')),
+                ('user', models.CharField(help_text='DSpace username to authenticate as', max_length=64, verbose_name='User')),
+                ('password', models.CharField(help_text='DSpace password to authenticate with', max_length=64, verbose_name='Password')),
+                ('archive_format', models.CharField(default=b'ZIP', max_length=3, verbose_name='Archive format', choices=[(b'ZIP', b'ZIP'), (b'7Z', b'7z')])),
```
jrwdunham: Is it probable that we might want to support other archive format types in the future? If so, we might prefer increasing `max_length` a bit now in order to avoid having to migrate/alter the table later.

jrwdunham: Is it more natural to say `Number of the` than `Identifier of the` or `Numeric identifier of the` in the `archivesspace_archival_object` and `archivesspace_repository` fields below?

storage_service/locations/migrations/0018_dspace_rest.py
```diff
@@ -0,0 +1,51 @@
...
+                ('archivesspace_archival_object', models.CharField(help_text='Number of the default ArchivesSpace archival object', max_length=64,
+                                              verbose_name='Default ArchivesSpace archival object')),
+                ('archivesspace_password', models.CharField(help_text='ArchivesSpace password to authenticate with', max_length=64,
+                                              verbose_name='ArchivesSpace password')),
+                ('archivesspace_repository', models.CharField(help_text='Number of the default ArchivesSpace repository', max_length=64,
+                                              verbose_name='Default ArchivesSpace repository')),
+                ('archivesspace_url', models.URLField(help_text='URL of ArchivesSpace server. E.g. http://sandbox.archivesspace.org/ (default port is 8089)', max_length=256,
+                                              verbose_name='ArchivesSpace URL')),
```
jrwdunham: Does this mean I can specify a different port? I think it does. If so, that should be in the help text I think.

storage_service/locations/migrations/0018_dspace_rest.py
jrwdunham: All the comments here should be on the model itself. I just read this PR top to bottom.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
+"""
+Integration with DSpace, using REST API as the protocol.
+
+NOTE THAT IN ALL REQUESTS TO THE REST API THE SSL VERIFICATION HAS BEEN DISABLED
```
jrwdunham: Why not set a toggle for this that can be modified in the GUI and is an attribute of the space? Then have it default to enabling SSL verification for security but still have more flexibility when setting up an integration.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+from common import utils
+from .location import Location
+
+LOGGER = logging.getLogger(__name__)
+
+
+class DSpaceREST(models.Model):
+    """Integration with DSpace using the REST API."""
+    space = models.OneToOneField('Space', to_field='uuid')
+    rest_url = models.URLField(max_length=256,  # ATTENTION could be with or without slash at end
```
jrwdunham: Why not add `' with optional trailing slash'` to the `help_text`?

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+
+    def move_from_storage_service(self, source_path, destination_path, package=None):
+        LOGGER.info('source_path: %s, destination_path: %s, package: %s', source_path, destination_path, package)
+        LOGGER.info('Package UUID: %s', package.uuid)
+        LOGGER.info('Package type: %s', package.package_type)
+        LOGGER.info('Port: %s', self.archivesspace_port)
+        self.rest_url = self.rest_url.replace('http:', 'https:')  # DSPACE REST API really dislikes http
+
+        if self.rest_url[-1:] == '/':  # Get rid of trailing slash
+            self.rest_url = self.rest_url[:-1]
```
jrwdunham: Probably better to just use `self.rest_url = self.rest_url.rstrip('/')` because one line and you'll get rid of multiple trailing slashes.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+
+        if self.archivesspace_url[-1:] == '/':  # Get rid of trailing slash
+            self.archivesspace_url = self.archivesspace_url[:-1]
+
+        if len(self.archivesspace_url.split(':')) == 3:  # If ArchivesSpace url contains port assign it
+            self.archivesspace_port = self.archivesspace_url.split(':')[2]
```
jrwdunham: Probably would be more longterm robust to use `from django.utils.six.moves.urllib.parse import urlparse` for these kind of URL manipulations. Grep this repo for examples or see https://github.com/artefactual/archivematica-storage-service/blob/stable/0.11.x/storage_service/locations/models/pipeline.py.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+
+        if package is None:
+            LOGGER.warning('DSpace requires package param')
+            return
+
+        # This only handles compressed AIPs
+        # if not os.path.isfile(source_path):
+        #    raise NotImplementedError(_('Storing in DSpace does not support uncompressed AIPs'))
```
jrwdunham: Why are you not raising this error under this condition? Does DSpace at the version you are targeting handle uncompressed AIPs? I think that further on in `_get_metadata` you're assuming the AIP is a file that needs to be extracted so when the AIP is a directory you'll trigger a `subprocess.CalledProcessError` and will return an empty dict ... In short, I think you should uncomment the above block to more gracefully deal with attempts to move uncompressed AIPs into this space.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+    ARCHIVE_FORMAT_7Z = '7Z'
+    ARCHIVE_FORMAT_CHOICES = (
+        (ARCHIVE_FORMAT_ZIP, 'ZIP'),
+        (ARCHIVE_FORMAT_7Z, '7z'),
+    )
+
+    archive_format = models.CharField(max_length=3,
+                                      choices=ARCHIVE_FORMAT_CHOICES,
+                                      default=ARCHIVE_FORMAT_ZIP,
+                                      verbose_name=_l('Archive format'))
```
jrwdunham: How come no help text here? Is this a DSpace-targeting configuration or parameter?

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+
+    def _get_metadata(self, input_path, aip_uuid, package_type):
+        """Get metadata for DSpace from METS file."""
+        output_dir = os.path.dirname(input_path) + '/'
+        dirname = os.path.splitext(os.path.basename(input_path))[0]
+        mets_path = ''
+
+        if package_type == 'AIP':
+            # Extract METS file
+            # TODO Should output dir be a temp dir?
```
jrwdunham, line 129: Maybe use `tempfile`, cf. https://github.com/artefactual/archivematica-storage-service/blob/stable/0.11.x/storage_service/locations/models/package.py#L1288.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+            for dip_file in os.listdir(input_path):
+                if dip_file.startswith('METS') and dip_file.endswith(".xml"):
+                    LOGGER.info("Dip file: %s", dip_file)
+                    mets_path = os.path.join(input_path, dip_file)
+
+        # Fetch info
+        root = etree.parse(mets_path)
+
+        dmdids = self._getdmdids(root.find("//mets:structMap[@ID='structMap_1']/", namespaces=utils.NSMAP))
+        inv_dmdids = {v: k for k, v in dmdids.iteritems()}
```
jrwdunham: If you always inverse the output of `_getdmdids`, then why not just have that method return the output in the correct format to begin with?

jrwdunham: Also, and more importantly, `DMDID`s are guaranteed to be globally unique within the document (I think) but `<mets:div>` labels (i.e., directory names) may easily not be unique. Isn't it possible that an `objects/` subdirectory could overwrite the top-level `objects/` directory that you are targeting below?

I'm also wondering if you considered using [mets-reader-writer](https://github.com/artefactual-labs/mets-reader-writer) for this METS parsing work. It might be able to do what you want more elegantly here.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+                self._format_metadata('dc.title', dirname[:dirname.find(aip_uuid) - 1].replace('_', ' ').title())]
+            LOGGER.info('Metadata: %s', metadata)
+
+        os.remove(mets_path)
+        return metadata, repos, package_title
+
+    # Reformats the metadata for the REST API
+
+    def _format_metadata(self, dc, value):
+        return {'key': dc, 'value': value, 'language': ''}
```
jrwdunham: Since `self` isn't used in this method, you should remove it and add the `@staticmethod` decorator to it.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+                    repos['dspace_aip_collection'] = other_metadata.findtext('dspace_aip_collection')
+                    repos['archivesspace_dip_collection'] = other_metadata.findtext('archivesspace_dip_collection')
+                    repos['archivesspace_aip_collection'] = other_metadata.findtext('archivesspace_aip_collection')
+                elif dc_metadata is not None:
+                    for md in dc_metadata:
+                        dc_term = str(md.tag)[str(md.tag).find('}') + 1:]
+
+                        if dc_term == 'title':
+                            dc_term = 'dc.' + dc_term
+                            package_title = dc_metadata.findtext(md.tag, namespaces=utils.NSMAP)
```
jrwdunham, line 172: I think you can just get the text of an XML element by calling `md.text` instead of `dc_metadata.findtext(md.tag, namespaces=utils.NSMAP`.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+                        else:
+                            dc_term = 'dcterms.' + dc_term
+
+                        metadata.append(self._format_metadata(dc_term, dc_metadata.findtext(md.tag,
+                                                                                            namespaces=utils.NSMAP)))
+
+        if len(metadata) == 0:  # We have nothing and therefore filename becomes title
```
jrwdunham: More idiomatic in python to do `if not metadata:`.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+            LOGGER.warning('Could not find SIP level Dublin Core metadata in %s', input_path)
+            metadata = [
+                self._format_metadata('dc.title', dirname[:dirname.find(aip_uuid) - 1].replace('_', ' ').title())]
+            LOGGER.info('Metadata: %s', metadata)
```
jrwdunham: In this case where you're creating a default dc.title out of the `dirname` and putting it into the `metadata` list, do you also want to set `package_title` to that same value?

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+            relative_mets_path = os.path.join(dirname, 'data', 'METS.' + aip_uuid + '.xml')
+            mets_path = os.path.join(output_dir, relative_mets_path)
+            command = ['unar', '-force-overwrite', '-o', output_dir, input_path, relative_mets_path]
+            try:
+                p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+                output, err = p.communicate()
+                LOGGER.info('Output: %s', output)
+            except subprocess.CalledProcessError:
+                LOGGER.error('Could not extract %s from %s', mets_path, input_path, exc_info=True)
+                return {}
```
jrwdunham: Shouldn't you return a 3-tuple here so that the return value is the same as in the happy path? (If not, you're going to trigger a `ValueError` when the destructuring attempt happens:

    a, b, c = {}
      File "<stdin>", line 1, in <module>
    ValueError: need more than 0 values to unpack

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+
+        metadata, repository_collections, package_title = self._get_metadata(source_path,
+                                                                             package.uuid,
+                                                                             package.package_type)
+
+        if package.package_type == 'DIP':
+            if 'dspace_dip_collection' in repository_collections:
+                dspace_collection = repository_collections['dspace_dip_collection']
+            else:
+                dspace_collection = self.dspace_dip_collection
```
jrwdunham: This pattern seems odd to me because you're doing lookup in the dict twice when you only need to do it once. One alternative approach would be:

    dspace_collection = repository_collections.get('dspace_dip_collection')
    if not dspace_collection:
        dspace_collection = self.dspace_dip_collection

An approach like the the above would also allow you to avoid doubly writing long expressions like `'archivesspace_' + package.package_type.lower() + '_collection'` below.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+            if 'dspace_aip_collection' in repository_collections:
+                dspace_collection = repository_collections['dspace_aip_collection']
+            else:
+                dspace_collection = self.dspace_aip_collection
+
+        if 'archivesspace_' + package.package_type.lower() + '_collection' in repository_collections:
+            archivesspace_collection = repository_collections['archivesspace_' + package.package_type.lower() +
+                                                              '_collection']
+        elif package.package_type == 'DIP':
+            archivesspace_collection = self.archivesspace_archival_object
```
jrwdunham: Again, I probably don't understand the jargon here, but is it necessary that the user must specify an AS "collection" in the JSON metadata while specifying an AS "archival object" in the Space attribute? Since the values seem to be used for the same purposes, why not just call them both "collection" or "archival object"?

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+            LOGGER.error('Could not create DSpace record: %s', collection_url)
+
+        dspace_item = response.json()
+
+        LOGGER.info('DSpace item: %s', dspace_item)
+
+        if package.package_type == 'DIP':
+            LOGGER.info('Package is DIP. Splitting package.')
+
+            for root, dirs, files in os.walk(source_path):
```
jrwdunham: Since you're not using the `dirs` var, the convention is to use `_` instead, or use `__` in this case since `_` is already aliasing `ugettext`.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+            LOGGER.info('Logged in to REST API.')
+        except Exception:
+            LOGGER.warning('Error logging in to DSpace REST API, aborting', exc_info=True)
+            return
+
+        set_cookie = response.headers['Set-Cookie'].split(';')[0]
+
+        return set_cookie[set_cookie.find('=') + 1:]
+
+    def move_from_storage_service(self, source_path, destination_path, package=None):
```
jrwdunham: This is a very long function. For readability, it might be nice to break it up into multiple shorter ones.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+
+        # Create item in DSpace
+        try:
+            response = requests.post(collection_url,
+                                     headers=headers,
+                                     cookies={'JSESSIONID': dspace_sessionid},
+                                     data=json.dumps(item),
+                                     verify=False)
+        except Exception:
+            LOGGER.error('Could not create DSpace record: %s', collection_url)
```
jrwdunham: Is this a fatal exception? Looks like it is because the next line is going to fail if this exception is raised. Maybe you should be returning here or raising the exception. Also, it would be nice if you could catch an expected exception like `requests.exceptions.RequestException` instead of the base exception class.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+        if package.package_type == 'DIP':
+            LOGGER.info('Package is DIP. Splitting package.')
+
+            for root, dirs, files in os.walk(source_path):
+                for name in files:
+                    with open(os.path.join(root, name), 'r') as f:
+                        content = f.read()
+
+                    bitstream_url = self.rest_url + '/items/' + dspace_item['uuid']
+                    bitstream_url += '/bitstreams?name=' + urllib.quote(name)
```
jrwdunham: I think it's possible for name to be a Unicode object containing characters out of the ASCII range, so you need to encode it (to UTF-8 I would guess) before calling `quote()`:

    >>> urllib.quote(u'a')
    u'a'
    >>> urllib.quote(u'a\u0301')
      File "<stdin>", line 1, in <module>
      File "/Users/joeldunham/.pyenv/versions/2.7.8/lib/python2.7/urllib.py", line 1288, in quote
        return ''.join(map(quoter, s))
    KeyError: u'\u0301'
    >>> urllib.quote(u'a\u0301'.encode('utf8'))
    'a%CC%81'

Same comment for other use of `urllib.quote` below.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+                # Create digital object in ArchivesSpace linking to DIP
+                client = archivesspace.ArchivesSpaceClient(self.archivesspace_url, self.archivesspace_user,
+                                                           self.archivesspace_password, self.archivesspace_port,
+                                                           self.archivesspace_repository)
+            except Exception:
+                LOGGER.error('Could not login to ArchivesSpace server: %s, port: %s, user: %s, repository: %s',
+                             self.archivesspace_url,
+                             self.archivesspace_port,
+                             self.archivesspace_user,
+                             self.archivesspace_repository)
```
jrwdunham: Again, you should probably raise the caught exception here after logging or raise a custom exception because a `NameError` is just going to be raised in the next statement when the non-existent `client` is used and that may be more confusing I think.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+                        content = f.read()
+
+                    bitstream_url = self.rest_url + '/items/' + dspace_item['uuid']
+                    bitstream_url += '/bitstreams?name=' + urllib.quote(name)
+
+                    try:
+                        requests.post(bitstream_url,
+                                      headers=headers,
+                                      cookies={'JSESSIONID': dspace_sessionid},
+                                      data=content,
```
jrwdunham: Sending the raw bytes of an arbitrary file in the HTTP body seems dangerous. What if the file is very large? Then you're loading the entire thing into memory and issuing a request that will probably time out. Is there no other way to do this via the DSpace REST API, e.g., by chunking the file?

This is an even bigger problem when it comes to sending the entire AIP via HTTP I would think.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+            except Exception as e:
+                LOGGER.error('Could not add digital object to ArchivesSpace')
+                if e.response.status_code == 404:
+                    raise ValueError('Either repository %s or archival object %s does not exist',
+                                     14, archivesspace_collection)
+
+                raise ValueError('ArchivesSpace Server error: %s', e.response.text)
+
+            LOGGER.info('ArchivesSpace Client: %s', client)
+            LOGGER.info('URL: %s', self.rest_url[:-4] + '/handle/' + dspace_item['handle'])
```
jrwdunham: Are all DSpace REST URL endpoints guaranteed to be exposed at the `'/rest'` path? This seems like a brittle assumption. Might be another case where using `urlparse` would be a good idea, e.g.,

    >>> a = 'http://demo.dspace.org/rest'
    >>> a
    'http://demo.dspace.org/rest'
    >>> b = '{}://{}/'.format(urlparse.urlparse(a).scheme, urlparse.urlparse(a).netloc)
    >>> b
    'http://demo.dspace.org/'

Relatedly, you should probably just construct this DSpace handle URL once.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+            except Exception:
+                LOGGER.info('Error sending %s, to %s', source_path, bitstream_url)
+
+            # LOGGER.debug('Dest: %s', bitstream_url)
+
+            # Send AIP to Tivoli Storage Manager using command-line client
+            command = ['dsmc', 'archive', source_path]
+            try:
+                LOGGER.info('Command: %s', command)
+                LOGGER.info('IS file: %s', os.path.isfile(source_path))
```
jrwdunham: I don't think you could have gotten this far if `source_path` were not a file.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+                              data=content,
+                              verify=False)
+                LOGGER.debug('%s successfully sent to %s', source_path, bitstream_url)
+            except Exception:
+                LOGGER.info('Error sending %s, to %s', source_path, bitstream_url)
+
+            # LOGGER.debug('Dest: %s', bitstream_url)
+
+            # Send AIP to Tivoli Storage Manager using command-line client
+            command = ['dsmc', 'archive', source_path]
```
jrwdunham: I think the dependency on this `dsmc` tool will require some additional work, right? I mean, we will need to modify our various deployment strategies (ansible, pacakges, docker) to install it. Maybe this is outside of the scope of what you want to do, but I think it would still need to be done.

Relatedly, would it be possible to make this `Tivoli Storage Manager` upload configurable? This seems like a very customized workflow.

Finally, from a very cursory look at https://www.ibm.com/support/knowledgecenter/TI0003M/p8hcg/p8hcg_dsmc.htm, it looks like the `archive` subcommand just `Displays a list of archived files`. Am I reading that wrong?

jrwdunham (commented):
Hey @J4bbi, I have a bunch of comments, questions and suggestions. I've only read through the code and haven't tested it with real Dspace and ArchivesSpace instances. However, I'll be away until Monday so I thought I'd post an initial review now in case it's helpful.

storage_service/locations/models/dspace_rest.py
jrwdunham: Also, is `archive_format` used at all? If not, is it necessary?

jrwdunham: Also, is `archive_format` used at all? If not, is it necessary?

storage_service/locations/forms.py
J4bbi: You are pointing to forms.py above where there is a ```password``` field. But I see I've accidentally omitted the field in constants.py.

As per the field names, I agree short prefixes would increase legibility.

storage_service/locations/migrations/0018_dspace_rest.py
J4bbi: Yep, you are correct. This was a last minute change by me, to reduce the number of fields by 1 (congested space already).

I'lll add to the help text.

storage_service/locations/migrations/0018_dspace_rest.py
J4bbi: Note that the base I worked on was of course the DSpace space and this was c+p from there.

Change max-length to 4? Or longer?

storage_service/locations/models/dspace_rest.py
J4bbi: I had considered doing that actually. Yeah, why not?

storage_service/locations/models/dspace_rest.py
J4bbi: Agreed.

storage_service/locations/models/dspace_rest.py
J4bbi: This is a "legacy" issue from the DSpace location. I hadn't removed it because I hadn't thought clearly about it.

What the DSpace location does is it takes the zipped AIP and splits it into two zips. So yeah, it's unnecessary.

storage_service/locations/models/dspace_rest.py
J4bbi: I'm just going to use urlparse which will resolve url handling.

storage_service/locations/models/dspace_rest.py
J4bbi: My lack of experience with classes in Python showing. Will do.

storage_service/locations/models/dspace_rest.py
J4bbi: Yeah, that code is a bit ad hoc. I'm not too familiar with METS or the DMDIDs.

It had occured to me that possibility that an `objects` subdirectory might overwrite the value. (That's very impressively spotted btw). I hadn't done more testing.

I will try to take alook at mets-reader-writer.

storage_service/locations/models/dspace_rest.py
J4bbi: Yup. Silly omission on my behalf.

storage_service/locations/models/dspace_rest.py
J4bbi: Condition wise what's happening is that mets files can't be extracted for the AIP.

No need to return anything at that point I believe. I'll just remove the return.

Thanks

storage_service/locations/models/dspace_rest.py
J4bbi: Good call.

storage_service/locations/models/dspace_rest.py
J4bbi: Yeah. I'll admit the logic is a bit confused.

storage_service/locations/models/dspace_rest.py
J4bbi: This is due to the fact that I've (almost willfully) put off learning the different conceptual objects/types of the ArchivesSpace world.

First off I used collection borrowing directly from DSpace but, as I understand it, what I'll want to refer to is an "archival object". I believe a digital object can only be a child of an archival object.

So I suppose I'll refactor to archival object.

storage_service/locations/models/dspace_rest.py
J4bbi: Yes. You are correct.

storage_service/locations/models/dspace_rest.py
J4bbi: Yes. Good point.

storage_service/locations/models/dspace_rest.py
J4bbi: These are good points that you raise.

I'm not familiar with any way to do this by chunking.

I'll give this more thought and ask around.

storage_service/locations/models/dspace_rest.py
J4bbi: Yeah. This was an "old" construct I forgot to replace.

storage_service/locations/models/dspace_rest.py
J4bbi: Yup. This is an old debug helper log. Will remove.

storage_service/locations/models/dspace_rest.py
J4bbi: I see the dsmc functionality as one that's purely customised to our needs, i.e. shouldn't really be in the pull request. To be fair, the DSpace space implemented is pretty customised as well (and only works with certain DSpace 5 I think).

You're reading that correctly but from the wrong source, this is the correct reference:
https://www.ibm.com/support/knowledgecenter/en/SSEQVQ_8.1.0/client/c_arc_cmndlineunx.html

storage_service/locations/models/dspace_rest.py
J4bbi: Done

storage_service/locations/migrations/0018_dspace_rest.py
J4bbi: I'm just going to delete this field.

storage_service/locations/models/dspace_rest.py
J4bbi: Because the getdmdids function is recursive I can't easily have it return the inverse.

I did do a workaround though.

storage_service/locations/models/dspace_rest.py
J4bbi: If I understand correctly, this might be a solution?:

```
with open(os.path.join(root, name), 'rb') as content:
                            requests.post(bitstream_url,
                                      headers=headers,
                                      cookies={'JSESSIONID': dspace_sessionid},
                                      data=content,
                                      verify=False)
```

storage_service/locations/models/dspace_rest.py
jrwdunham: You can't pass a bunch of string interpolation arguments to an exception constructor like you can when calling a method on a logger. Therefore, your statement:


    raise Exception('Could not login to ArchivesSpace server: %s, port: %s, user: %s, repository: %s',
                self.as_url,
                self.as_url.port,
                self.as_user,
                self.as_repository)

has to be changed to something like:

    raise Exception('Could not login to ArchivesSpace server: {}, port: {}, user: {}, repository: {}'.format(
                self.as_url,
                self.as_url.port,
                self.as_user,
                self.as_repository))

storage_service/locations/models/dspace_rest.py
jrwdunham: Even simpler would be the pattern of `dspace_collection = repository_collections.get('dspace_dip_collection', self.ds_dip_collection)`, but no need to do that if you don't want.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,379 @@
...
+                output, err = p.communicate()
+                LOGGER.info('Output: %s', output)
+            except subprocess.CalledProcessError:
+                LOGGER.error('Could not extract %s from %s', mets_path, input_path, exc_info=True)
+                return {}
+        elif package_type == 'DIP':
+            for dip_file in os.listdir(input_path):
+                if dip_file.startswith('METS') and dip_file.endswith(".xml"):
+                    LOGGER.info("Dip file: %s", dip_file)
+                    mets_path = os.path.join(input_path, dip_file)
```
jrwdunham, line 143: Once you've found the METS file you should call `break` so you don't needlessly iterate over all paths in the DIP. Relatedly or alternatively, shouldn't we be able to know exactly what the DIP's METS file name is and just confirm that it exists and return that? For example, we know that the DIP METS filename will be `METS.<AIP_UUID>.xml`.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,398 @@
...
+        if package.package_type == 'DIP':
+            self._deposit_dip_to_dspace(source_path, headers, dspace_item, dspace_sessionid)
+
+            self._link_dip_to_archivesspace(archivesspace_collection, package.uuid, package_title, dspace_item)
+
+        else:
+            self._deposit_aip_to_dspace(source_path, headers, dspace_item, dspace_sessionid)
+
+            # Send AIP to Tivoli Storage Manager using command-line client
+            command = ['dsmc', 'archive', source_path]
```
jrwdunham: While it is ok for the dsmc stuff to remain a part of the DSpaceREST space, it definitely needs at a minimum to be configurable within the space creation interface so that attempting to issue the `dsmc archive ...` command is optional, and probably disabled by default.

Because the Tivoli Storage Manager (TSM) archiving/storing is logically orthogonal to the DSpace storing, it would be preferable to implement this in its own TSM Space and then use the SS's AIP replication functionality to configure a TSM space as a replicator of a DSpaceREST space.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,398 @@
...
+        return dspace_collection, archivesspace_collection
+
+    def _deposit_dip_to_dspace(self, source_path, headers, dspace_item, dspace_sessionid):
+        LOGGER.info('Package is DIP. Sending file by file.')
+
+        for root, __, files in os.walk(source_path):
+            for name in files:
+                bitstream_url = self.ds_rest_url.scheme + '://' + self.ds_rest_url.netloc + self.ds_rest_url.path
+                bitstream_url += '/items/' + dspace_item['uuid'] + '/bitstreams?name=' + urllib.quote(
+                    name.encode('utf-8'))
```
jrwdunham: You should be able to create the base URL (i.e., `rest/items/`) once before the for loop instead of re-computing it for each file.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,398 @@
...
+        # Create item in DSpace
+        try:
+            response = requests.post(collection_url,
+                                     headers=headers,
+                                     cookies={'JSESSIONID': dspace_sessionid},
+                                     data=json.dumps(item),
+                                     verify=self.verify_ssl)
+        except RequestException:
+            LOGGER.error('Could not create DSpace record: %s', collection_url)
+            raise RequestException('Could not create DSpace record: %s', collection_url)
```
jrwdunham: This syntax for instantiating a Python exception is probably not doing what you want: you can't pass string interpolation arguments as args N+1 to an exception constructor. This needs to be fixed here and in the other places where you're doing it. Relatedly, it seems odd to instantiate a new exception using the Requests library's class.) In this case, one improvement might be:

    class DSpaceRESTException(Exception):
        pass
    msg = 'Could not create DSpace record: {}'.format(collection_url)
    LOGGER.error(msg)
    raise DSpaceRESTException(msg)

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,398 @@
...
+        try:
+            with open(source_path, 'rb') as content:
+                requests.post(bitstream_url,
+                              headers=headers,
+                              cookies={'JSESSIONID': dspace_sessionid},
+                              data=content,
+                              verify=self.verify_ssl)
+            LOGGER.debug('%s successfully sent to %s', source_path, bitstream_url)
+        except Exception:
+            LOGGER.info('Error sending %s, to %s', source_path, bitstream_url)
```
jrwdunham: This seems like a significant failure if we get to this point. I think you should be using `LOGGER.error` here. More importantly, I don't think the exception should be swallowed here. We need to make sure it is raised so that the request to store the AIP (or DIP) fails when one of the crucial steps fails. I think you need to go through the code and a) make sure that the log statements are the appropriate level and b) make sure that fatal exceptions are not being suppressed but are being allowed to bubble up so that Archivematica will know when the request to store the AIP has failed. If it is important to make sure that we make the call to logout from DSpace no matter what, then you could put that code in a finally block and any fatal exception will still get raised, for example:

    >>> try:
    ...   print(dog)
    ... except NameError:
    ...   print('no dog')
    ...   raise
    ... finally:
    ...   print('done')
    ... 
    no dog
    done
      File "<stdin>", line 2, in <module>
    NameError: name 'dog' is not defined

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,398 @@
...
+                dc_metadata = root.find('mets:dmdSec[@ID="' + anId + '"]/mets:mdWrap/mets:xmlData/dcterms:dublincore',
+                                        namespaces=utils.NSMAP)
+                other_metadata = root.find('mets:dmdSec[@ID="' + anId + '"]/mets:mdWrap[@MDTYPE="OTHER"]/mets:xmlData',
+                                           namespaces=utils.NSMAP)
+
+                if other_metadata is not None:
+                    repos['dspace_dip_collection'] = other_metadata.findtext('dspace_dip_collection')
+                    repos['dspace_aip_collection'] = other_metadata.findtext('dspace_aip_collection')
+                    repos['archivesspace_dip_collection'] = other_metadata.findtext('archivesspace_dip_collection')
+                    repos['archivesspace_aip_collection'] = other_metadata.findtext('archivesspace_aip_collection')
```
jrwdunham: I don't think that the `archivesspace_aip_collection` metadata value is ever used. Is it? If not, I think it should be deleted.

jrwdunham (changes requested):
This new space needs tests. We use Python's VCR.py to freeze HTTP requests and responses for tests. See https://github.com/artefactual/archivematica-storage-service/blob/stable/0.11.x/storage_service/locations/tests/test_dspace.py as an example.

The requests to ArchivesSpace need to be made configurable. That is, a user needs to be able to create a DSpaceREST space and omit the ArchivesSpace configuration and have the space still work for them. This toggle configuration needs to be implemented and tests need to be written for this.

Given UoE's requirements and workflow, I see why you followed the strategy of configuring ArchivesSpace (AS) within the SS Space as you did. Similar work (as you know) has used the post_store_aip_hook.py client script to achieve a similar result. See https://github.com/artefactual/archivematica/blob/stable/1.7.x/src/MCPClient/lib/clientScripts/post_store_aip_hook.py#L27-L85.  However, because you are interested in an automated workflow that uses transfer metadata to control AS configuration, I can see why the post_store_aip_hook.py strategy would have entailed adding a lot more complexity (likely adding MS Chain Links to Archivematica). Yet another approach that has been used is to use Archivematica's Automation Tools to upload DIP (metadata) post DIP storage. I think that the strategy that you have chosen here is acceptable, assuming you make the AS stuff configurable as mentioned above; I am just pointing out these other alternative strategies to make sure that we considered them and had good reasons for deciding not to go with them.  Another reason is that we may want to modify the logic in post_store_aip_hook to be more in line with the AgentArchives-based approach that is used in this PR.

Finally, this PR should be rebased to target qa/0.x branch I think.

storage_service/locations/models/dspace_rest.py
J4bbi: Yes. Of course.

I had problems accessing the AIP UUID during the DIP function call. I might be misunderstanding something but this taken from you on slack June 9th:

"So via the SS Django ORM you should be able to get from a DIP package to its source AIP package. (I am a bit confused about how this actually happens because looking at the log of the "Store DIP" micro-service, I see that the request body object contains `'related_package_uuid': None,`. The log of my "Store AIP" micro-service, on the other hand, references the DIP in its request body: `'related_package_uuid': u'010f4a96-b25e-4688-b6f5-5d186d957c79',`)."

Can you come up with a way to access it?

storage_service/locations/models/dspace_rest.py
J4bbi: I'm going to suggest that we just remove the TSM stuff from the pull request and we'll have it on a customised branch for our own use.

storage_service/locations/models/dspace_rest.py
J4bbi: Of course. Good point.

storage_service/locations/models/dspace_rest.py
J4bbi: Yeah, cool

storage_service/locations/models/dspace_rest.py
J4bbi: Done

storage_service/locations/models/dspace_rest.py
J4bbi: Yeah. I'll be the first to admit that exception handling is sloppy.

I'll do just what you suggest.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,404 @@
...
+        try:
+            ds_item = self._create_dspace_record(headers, metadata, ds_sessionid, ds_collection)
+
+            LOGGER.info('DSpace item: %s', ds_item)
+
+            if package.package_type == 'DIP':
+                self._deposit_dip_to_dspace(source_path, headers, ds_item, ds_sessionid)
+
+                if self.as_url != '' and self.as_user != '' and self.as_password != '' and self.as_repository != '':
+                    self._link_dip_to_archivesspace(as_archival_obj, package.uuid, package_title, ds_item)
```
jrwdunham: It would be more idiomatic to express this condition as `if all([self.as_url, self.as_user, self.as_password, self.as_repository]):`, but more importantly I think it's possible for `as_archival_obj` to evaluate to `None` here, so I think it needs to be `if all([self.as_url, self.as_user, self.as_password, self.as_repository, as_archival_object]):`

J4bbi: Spot on

jrwdunham:
I'm closing this in favour of https://github.com/artefactual/archivematica-storage-service/pull/394 which. PR 394 has the following changes:

- all commits from this PR squashed down to 1
- rebased on qa/0.x
- unit tests added.

Issue #391, opened 2018-07-11
Problem: packages cannot be stored in DSpace via its REST API


jrwdunham:
The [current DSpace space](https://github.com/artefactual/archivematica-storage-service/blob/stable/0.11.x/storage_service/locations/models/dspace.py) makes requests to DSpace using the SWORD2 protocol. Some users of Archivematica (the Storage Service) need to make use of the DSpace REST API.

currmie:
Verified. Also, see https://github.com/artefactual/archivematica-storage-service-docs/pull/12 for corresponding SS documentation updates.

PR #394, opened 2018-07-18
Add DSpace REST space

jrwdunham:
Connected to #391 

Replaces https://github.com/artefactual/archivematica-storage-service/pull/388


.travis.yml
```diff
@@ -1,7 +1,7 @@
 ---
 
 dist: trusty
-sudo: false
+sudo: required
```
sevein: I'm hoping we don't have to go into full virtual machines because it's going to be much slower. Do you mind checking again without this change to see what's the error? Maybe it was just a temporary issue.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,494 @@
...
+from django.utils.six.moves.urllib.parse import urlparse
+from django.utils.translation import ugettext as _, ugettext_lazy as _l
+
+# Third party dependencies, alphabetical
+from lxml import etree
+import requests
+from requests import RequestException
+
+# This project, alphabetical
+from agentarchives import archivesspace
```
sevein, line 25: I'm confused, this is a space that uses both ArchivesSpace and DSpace? Can you point me to a document where this integration is described?

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,494 @@
...
+        """Logout from DSpace API."""
+        rest_url = '{0.scheme}://{0.netloc}{0.path}'.format(self.ds_rest_url)
+        url = rest_url + '/logout'
+        try:
+            requests.post(
+                url,
+                headers=headers,
+                cookies={'JSESSIONID': ds_sessionid},
+                verify=self.verify_ssl)
+            LOGGER.info('Logged out of REST API.')
```
sevein: One recurrent thought I have about logging is that we should just log human-actionable events otherwise it's just noise for the application administrator. If you're a developer and you want this level of detail then use `debug`. Not something that we should start changing now but it'd be good to keep in mind, maybe someday we can apply a new strategy generally.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,494 @@
...
+    as_archival_object = models.CharField(
+        blank=True,
+        max_length=64,
+        verbose_name=_l("Default ArchivesSpace archival object"),
+        help_text=_l('Identifier of the default ArchivesSpace archival object'))
+
+    verify_ssl = models.BooleanField(
+        blank=True, default=True, verbose_name=_l("Verify SSL certificates?"),
+        help_text=_l('If checked, HTTPS requests will verify the SSL'
+                     ' certificates'))
```
sevein, line 122: We also have a global `INSECURE_SKIP_VERIFY`. This is probably better or more granular but I figured I'd mention it just in case you didn't remember and it'd change your mind.

https://github.com/artefactual/archivematica-storage-service/blob/31bad497be0ce1c94f3bf3413a93b75bae9815e3/storage_service/storage_service/settings/base.py#L411-L417

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,494 @@
...
+
+    @staticmethod
+    def _format_metadata(dc, value):
+        """Reformats the metadata for the REST API."""
+        return {'key': dc, 'value': value, 'language': ''}
+
+    @staticmethod
+    def _debug_ds_login(url, email, password):
+        return ('using url "{url}", email "{email}", and password'
+                ' "{password}"'.format(url=url, email=email, password=password))
```
sevein: I wonder if it would be cleaner if `DSpaceRESTException` had a constructor capable of taking `(url, email, password)` and include it in the final error message. Also I'd suggest not to include the password unless the debug mode is enabled.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,494 @@
...
+            '{rest_url.scheme}://{rest_url.netloc}{rest_url.path}/'
+            'collections/{ds_collection}/items'.format(
+                rest_url=self.ds_rest_url,
+                ds_collection=ds_collection))
+        # Create item in DSpace
+        try:
+            response = requests.post(collection_url,
+                                     headers=headers,
+                                     cookies={'JSESSIONID': ds_sessionid},
+                                     # data=json.dumps(item),
```
sevein: This line can be removed.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,494 @@
...
+            msg = 'Could not create DSpace record: {}'.format(collection_url)
+            LOGGER.error(msg)
+            raise DSpaceRESTException(msg)
+        try:
+            return response.json()
+        except ValueError:
+            msg = 'Not a JSON response'
+            LOGGER.error(msg)
+            LOGGER.error(response.text)
+            raise DSpaceRESTException(msg)
```
sevein: Do we have to log the error **and** raise and exception? It seems that the former is not needed unless you want it for debugging purposes so then I'd recommend to use the `debug` level instead, particularly when logging the body of the response.

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,494 @@
...
+        """Log in to get DSpace REST API token."""
+        body = {'email': self.ds_user, 'password': self.ds_password}
+        rest_url = '{0.scheme}://{0.netloc}{0.path}'.format(self.ds_rest_url)
+        try:
+            response = requests.post(
+                rest_url + '/login', data=body, verify=self.verify_ssl)
+            response.raise_for_status()
+        except requests.HTTPError:
+            LOGGER.warning('Error logging in to DSpace REST API, aborting',
+                           exc_info=True)
```
sevein: This is going to include the traceback which is rather noisy.
Should we just include the message in `HTTPError`?

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,494 @@
...
+        except Exception as err:
+            msg = 'ArchivesSpace Server error: {}'.format(err)
+            LOGGER.error(AS_DO_ADD_ERR)
+            LOGGER.error(msg)
+            raise DSpaceRESTException(msg)
+
+    def _deposit_aip_to_dspace(self, source_path, headers, ds_item, ds_sessionid):
+        LOGGER.info('Package is AIP. Sending file.')
+        bitstream_url = (
+            '{url.scheme}://{url.netloc}{url.path}'
```
sevein: There's more than one place where we build this base URL, should we have a dedicated function for it?

storage_service/locations/models/dspace_rest.py
```diff
@@ -0,0 +1,494 @@
...
+            source_path, package.uuid, package.package_type)
+        ds_collection, as_archival_obj = self._assign_destination(
+            package.package_type, destinations)
+
+        # Logging in to REST api gives us a session id
+        ds_sessionid = self._login_to_dspace_rest()
+
+        # Headers for connecting to DSpace REST API
+        headers = {'Accept': 'application/json',
+                   'Content-Type': 'application/json'}
```
sevein: Shouldn't this be a default down in the stack?

sevein (changes requested):
Nice tests @jrwdunham. Not easy because of the number of dependencies. It's the dependencies which confused me a bit. DSpace, ArchivesSpace, Tivoli... is the name of the space representative to this workflow?

I was wondering if we need a `MultiSpace` that the user can configure to combine multiple backends - perhaps a crazy idea out of scope, feel free to ignore.

storage_service/locations/models/dspace_rest.py
jrwdunham: Thanks @sevein, I had forgotten about `INSECURE_SKIP_VERIFY`. I think it's good to keep `DSpaceREST.verify_ssl` because, as you say, it is more fine-grained. This allows users to decrease their security wrt trusting responses from external services in the specific DSpace REST case but not elsewhere.

storage_service/locations/models/dspace_rest.py
jrwdunham: I'll cut out some of the logging cruft in any case.

storage_service/locations/models/dspace_rest.py
jrwdunham: Yeah, I've gone through and tried to use `raise DSpaceRESTException` consistently. Cleaner now I think.

storage_service/locations/models/dspace_rest.py
jrwdunham: mais oui

storage_service/locations/models/dspace_rest.py
jrwdunham: Good idea about the constructor. I've done that. Also removed the password from the exception message.

.travis.yml
sevein: I've confirmed this was just a temporary issue in Travis infra.

storage_service/locations/models/dspace_rest.py
jrwdunham: good idea

storage_service/locations/models/dspace_rest.py
jrwdunham: K. removed tracebacks in general.

storage_service/locations/models/dspace_rest.py
jrwdunham: I was unable to find a document for you @sevein. This development was done primarily by an Edinburgh dev (Hrafn) and, in general, they defined their own requirements internally. I have modified the DSpaceREST space creation form to make it clear that by omitting a single AS field you will be effectively "disabling" the AS functionality. @kellyannewithane or @jhsimpson may be able to provide some insight here.

jrwdunham:
@sevein I think I have addressed your change requests. Can you look again?

> is the name of the space representative to this workflow?

Good question. I think it is, despite the fact that DSpace, ArchivesSpace and Tivoli are involved. I have made the Tivoli upload logic configurable and disabled by default. The ArchivesSpace DIP linking stuff is closely linked to the DSpace DIP upload: the DIP is uploaded to DSpace and given a DSpace handle and that handle is provided when creating the ArchivesSpace digital object. (It may be that AIPs should also be linked/registered in AS, but that is another separate question.)

For the Tivoli stuff, the better way to do it would probably have been to create a dedicated Tivoli-specific space and set up AIP replication from the DSpace space to the Tivoli one. But at this point I think we'll have to consider that an improvement for the future.

sevein (approved):

PR #395, opened 2018-07-25
API: space creation and default location

sevein:
This pull request brings two commits:

- API: allow space creation (connects to https://github.com/archivematica/Issues/issues/36)
- API: create default location (connects to https://github.com/archivematica/Issues/issues/37)

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -319,6 +319,14 @@ class LocationResource(ModelResource):
                     pipeline=item.obj,
                     location=bundle.obj)
 
+    def obj_create(self, bundle, **kwargs):
+        """Create a new location and make it the default when requested."""
+        if 'default' in bundle.data:
+            # This is going to result in calling the `default` attribute setter
+            # in the underlying model (Location).
+            kwargs['default'] = bundle.data['default']
+        return super(LocationResource, self).obj_create(bundle, **kwargs)
+
     @_custom_endpoint(expected_methods=['get'])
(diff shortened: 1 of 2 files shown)


storage_service/locations/api/resources.py
```diff
@@ -319,6 +319,11 @@ def save_m2m(self, bundle):
                     pipeline=item.obj,
                     location=bundle.obj)
 
+    def obj_create(self, bundle, **kwargs):
+        if 'default' in bundle.data:
+            kwargs['default'] = bundle.data['default']
+        return super(LocationResource, self).obj_create(bundle, **kwargs)
+
```
jrwdunham: It was a bit subtle to me that `obj_create` calls `for key, value in kwargs.items(): setattr(bundle.obj, key, value)` in tastypie/resources.py, which calls the `default` attr's setter on the location, which ultimately results in `set_default_location_post_save` creating the `Setting` instance to store the default nature of the location.

jrwdunham (approved):
Small changes, with tests. Most excellent.

storage_service/locations/api/resources.py
sevein: Yeah, not great. The "default" attribute is fairly simple at the model layer but I could probably do better in `LocationResource` (tastypie). I remember trying to add `default` as a new filed in `LocationResource` hoping that'd make things much cleaner but I had some issues.

sevein: I've added a comment to the method hoping that will make things easier to understand for the reader.

PR #398, opened 2018-08-05
Update Dataverse Storage Service features

### LINKED ISSUE #352: Problem: Dataverse create/edit space page is not rendering correctly

### LINKED ISSUE #355: Problem: Can't filter Dataverse datasets by query-string

### LINKED ISSUE #380: Problem: Search flexibility needs expanding for Dataverse
Issue #355 discusses an issue with filtering Dataverse searches ([here](https://github.com/artefactual/archivematica-storage-service/issues/355#issuecomment-397835304)). There is a short-term workaround proposed which would take a short number of hours to correct, but it might not be the most appropriate solution for a more thorough solution to ingest of Dataverse material. The most appropriate approach needs to be discussed. 355 is solved by providing the capability equivalent to that of the PoC but little more.

ross-spencer:
This commit updates various features of the Storage Service related
to browsing and transferring Dataverse datasets inside Archivematica.
The most substantial changes are around transfer browsing where the
toggle will now show the contents of a Dataset when a user looks
into a dataset folder.

Creating a location will no longer cause a Django rendering error
when displaying the 'beta feature' string.

Query flexibility has been improved to include subtree. This
provides users with a greater flexibility to restrict datasets
shown in the transfer browser.

There are various code formatting changes too in-line with PEP8.

Resolves #352
Resolves #355
Resolves #380
Resolves archivematica/issues#113
Supersedes #386 
Connected to artefactual/archivematica#1085
Connected to archivematica/issues#108

Cherry-picked from https://github.com/artefactual/archivematica-storage-service/pull/402

This PR is required to work with the Dataverse features planned to be introduced with Archivematica `stable/1.8`.

### DIFF
--- a/storage_service/locations/forms.py
+++ b/storage_service/locations/forms.py
@@ -5,6 +5,7 @@ from django import forms
 import django.utils
 import django.core.exceptions
 from django.db.models import Count
+from django.utils.safestring import mark_safe
 from django.utils.translation import ugettext as _, ugettext_lazy as _l
 
 from common import gpgutils
@@ -39,14 +40,14 @@ class DisableableSelectWidget(forms.Select):
     def render_option(self, selected_choices, option_value, option_label):
         option_value = django.utils.encoding.force_text(option_value)
         if option_value in selected_choices:
-            selected_html = django.utils.safestring.mark_safe(' selected="selected"')
+            selected_html = mark_safe(' selected="selected"')
             if not self.allow_multiple_selected:
                 # Only allow for a single selection.
                 selected_choices.remove(option_value)
         else:
             selected_html = ''
         if option_value in self.disabled_choices:
-            disabled_html = django.utils.safestring.mark_safe(' disabled="disabled"')
+            disabled_html = mark_safe(' disabled="disabled"')
         else:
             disabled_html = ''
         return django.utils.html.format_html(
@@ -116,11 +117,10 @@ class DataverseForm(forms.ModelForm):
 
     def as_p(self):
         # Add a warning to the Dataverse-specific section of the form
-        # FIXME this may not be the best way to add Space-specific warnings
         content = super(DataverseForm, self).as_p()
         content += '\n<div class="alert">{}</div>'.format(
             _('Integration with Dataverse is currently a beta feature'))
-        return content
+        return mark_safe(content)
 
 
 class DuracloudForm(forms.ModelForm):
@@ -273,6 +273,20 @@ class LocationForm(forms.ModelForm):
         # Associated with all enabled pipelines by default
         self.fields['pipeline'].initial = models.Pipeline.active.values_list('pk', flat=True)
         self.fields['default'].initial = self.instance.default
+        # If we are accessing a Dataverse we want to be able to do nice things
+        # with the Transfer Browser. Create a namespace to work with the path
+        # here.
+        if space_protocol == models.Space.DATAVERSE:
+            dv_config = ("Query: \nSubtree: ")
+            self.fields['relative_path'].initial = dv_config
+            # Help text for the relative path field on the Dataverse form. The
+            # form itself is not very responsive, so the smallest amount of
+            # text has been used to try and convey something useful.
+            dv_help_text = _(
+                "Path to location, relative to the storage space's path. "
+                "Optional. Filter Dataverse by matches and/or Dataverse "
+                "subtree.")
+            self.fields['relative_path'].help_text = dv_help_text
 
     def clean(self):
         cleaned_data = super(LocationForm, self).clean()
--- a/storage_service/locations/models/dataverse.py
+++ b/storage_service/locations/models/dataverse.py
@@ -1,8 +1,11 @@
 from __future__ import absolute_import
+
 # stdlib, alphabetical
 import json
 import logging
 import os
+import re
+import zipfile
 
 # Core Django, alphabetical
 from django.db import models
@@ -17,91 +20,189 @@ LOGGER = logging.getLogger(__name__)
 # This module, alphabetical
 from . import StorageException  # noqa: E402
 from .location import Location  # noqa: E402
+from .urlmixin import URLMixin  # noqa: E40
 
 
-class Dataverse(models.Model):
-    space = models.OneToOneField('Space', to_field='uuid')
-    host = models.CharField(max_length=256,
-        verbose_name=_l('Host'),
-        help_text=_l('Hostname of the Dataverse instance. Eg. apitest.dataverse.org'))
-    api_key = models.CharField(max_length=50,
-        verbose_name=_l('API key'),
-        help_text=_l('API key for Dataverse instance. Eg. b84d6b87-7b1e-4a30-a374-87191dbbbe2d'))
-    agent_name = models.CharField(max_length=50,
-        verbose_name=_l('Agent name'),
-        help_text=_l('Agent name for premis:agentName in Archivematica'))
-    agent_type = models.CharField(max_length=50,
-        verbose_name=_l('Agent type'),
-        help_text=_l('Agent type for premis:agentType in Archivematica'))
-    agent_identifier = models.CharField(max_length=256,
-        verbose_name=_l('Agent identifier'),
-        help_text=_l('URI agent identifier for premis:agentIdentifierValue in Archivematica'))
+class Dataverse(URLMixin, models.Model):
+    space = models.OneToOneField("Space", to_field="uuid")
+
+    host = models.CharField(
+        max_length=256,
+        verbose_name=_l("Host"),
+        help_text=_l(
+            "Hostname of the Dataverse instance. Eg. apitest.dataverse.org"
+        ),
+    )
+    api_key = models.CharField(
+        max_length=50,
+        verbose_name=_l("API key"),
+        help_text=_l(
+            "API key for Dataverse instance. Eg. "
+            "b84d6b87-7b1e-4a30-a374-87191dbbbe2d"
+        ),
+    )
+    agent_name = models.CharField(
+        max_length=50,
+        verbose_name=_l("Agent name"),
+        help_text=_l("Agent name for premis:agentName in Archivematica"),
+    )
+    agent_type = models.CharField(
+        max_length=50,
+        verbose_name=_l("Agent type"),
+        help_text=_l("Agent type for premis:agentType in Archivematica"),
+    )
+    agent_identifier = models.CharField(
+        max_length=256,
+        verbose_name=_l("Agent identifier"),
+        help_text=_l(
+            "URI agent identifier for premis:agentIdentifierValue "
+            "in Archivematica"
+        ),
+    )
     # FIXME disallow string in space.path
 
     class Meta:
         verbose_name = _l("Dataverse")
-        app_label = 'locations'
+        app_label = "locations"
 
-    ALLOWED_LOCATION_PURPOSE = [
-        Location.TRANSFER_SOURCE,
-    ]
+    ALLOWED_LOCATION_PURPOSE = [Location.TRANSFER_SOURCE]
+
+    @staticmethod
+    def get_query_value(key, path, default=None):
+        """Retrieve the value corresponding to ``key`` from the string
+        ``path``, then remove both the key and the value from path so that we
+        can recurse through it for all the key-value pairs used to construct
+        a Dataverse query. Finally return the value and path as a 2-tuple.
+        """
+        path = path.lower()
+        value = path.split(key.lower(), 1)
+        try:
+            value = value[1].strip()
+        except IndexError:
+            value = default
+        if value == "":
+            value = default
+        if value is None:
+            path = path.replace(key, "")
+        else:
+            path = re.sub(re.escape(key) + r"\s*" + re.escape(value), "", path)
+        return value, path.strip()
+
+    def get_query_and_subtree(self, path):
+        """Split the query string we have received and try and break it into
+        components that we can use to populate the transfer tree.
+
+        All the components are optional. We pull them apart right-to-left,
+        where a dataset identifier might be provided to us from the transfer
+        browser. We can then retrieve the subtree and the query string which
+        can be used to return a listing of datasets available in the Dataverse.
+        """
+        dataset, path = self.get_query_value("/", path)
+        subtree, path = self.get_query_value("subtree:", path)
+        query_string, path = self.get_query_value("query:", path, default="*")
+        return dataset, subtree, query_string
 
     def browse(self, path):
+        """Fetch the datasets in this dataverse or the files in the dataset
+        referenced in ``path``.
         """
-        Fetch a list of datasets from Dataverse based on the query in the location path.
+        LOGGER.info("Path received: %s", path)
+        dataset_id, subtree, query_string = self.get_query_and_subtree(path)
+        LOGGER.info("Dataset ID: %s Subtree: %s Query: %s",
+            dataset_id, subtree, query_string)
+        if dataset_id:
+            return self._browse_dataset(dataset_id)
+        return self._browse_dataverse(query_string, subtree)
 
-        Datasets are considered directories when browsing.
+    def _browse_dataverse(self, query_string, subtree):
+        """Return all datasets in all Dataverses (conforming to ``browse``
+        protocol).
         """
-        # Remove things earlier layers added
-        path = path.rstrip('/')
-        # Use http://guides.dataverse.org/en/latest/api/search.html to search and return datasets
-        # Location path is query string
-        # FIXME only browse one layer deep
-        url = 'https://' + self.host + '/api/search/'
+        LOGGER.info("Subtree: %s", subtree)
+        LOGGER.info("Query: %s", query_string)
+        url = self._generate_dataverse_url(slug="/api/search/")
         params = {
             'key': self.api_key,
-            'q': path,
+            'q': query_string,
+            'subtree': subtree,
             'type': 'dataset',
             'sort': 'name',
             'order': 'asc',
             'start': 0,
-            'per_page': 10,
+            'per_page': 50,
             'show_entity_ids': True,
         }
-        entries = []
         properties = {}
         while True:
             LOGGER.debug('URL: %s, params: %s', url, params)
             response = requests.get(url, params=params)
             LOGGER.debug('Response: %s', response)
+            # If the request isn't successful, i.e. doesn't return 200, then
+            # raise an exception. Other use cases from Dataverse might need to
+            # be considered, but we haven't examples of those as yet to go on
+            # and test with. We're only looking for 200 OK at present.
             if response.status_code != 200:
                 LOGGER.warning('%s: Response: %s', response, response.text)
                 raise StorageException(
-                    _('Unable to fetch datasets from %(url)s with query %(path)s') %
-                    {'url': url, 'path': path})
+                    _('Unable to fetch datasets from %(url)s')
+                    % {'url': url})
             try:
                 data = response.json()['data']
+                items = data['items']
             except json.JSONDecodeError:
                 LOGGER.error('Could not parse JSON from response to %s', url)
                 raise StorageException(
-                    _('Unable parse JSON from response to %(url)s with query %(path)s') %
-                    {'url': url, 'path': path})
-
-            entries += [str(x['entity_id']) for x in data['items']]
-
-            properties.update({
-                str(x['entity_id']): {'verbose name': x['name']}
-                for x in data['items']
-            })
-
+                    _('Unable to parse JSON from response to %(url)s')
+                    % {'url': url})
+            properties.update(
+                {str(ds['entity_id']): {'verbose name': ds['name']}
+                 for ds in items})
             if params['start'] + data['count_in_response'] < data['total_count']:
                 params['start'] += data['count_in_response']
             else:
                 break
-
-        directories = entries
+        entries = list(properties.keys())
         return {
-            'directories': directories,
+            'directories': entries,
+            'entries': entries,
+            'properties': properties,
+        }
+
+    def _browse_dataset(self, dataset_identifier):
+        """Return all files in the dataset with ``entity_id``
+        ``dataset_identifier`` (conforming to ``browse`` protocol).
+        """
+        files_in_dataset_path = (
+            '/api/v1/datasets/{dataset_identifier}/versions/:latest'.format(
+                dataset_identifier=dataset_identifier))
+        url = self._generate_dataverse_url(slug=files_in_dataset_path)
+        params = {'key': self.api_key}
+        LOGGER.debug('URL: %s, params: %s', url, params)
+        response = requests.get(url, params=params)
+        LOGGER.debug('Response: %s', response)
+        # If the request isn't successful, i.e. doesn't return 200, then raise
+        # an exception. Other use cases from Dataverse might need to be
+        # considered, but we haven't examples of those as yet to go on and test
+        # with. We're only looking for 200 OK at present.
+        if response.status_code != 200:
+            LOGGER.warning('%s: Response: %s', response, response.text)
+            raise StorageException(
+                _('Unable to fetch datasets from %(url)s')
+                % {'url': url})
+        try:
+            data = response.json()['data']
+            files = data['files']
+        except json.JSONDecodeError:
+            LOGGER.error('Could not parse JSON from response to %s', url)
+            raise StorageException(
+                _('Unable to parse JSON from response to %(url)s')
+                % {'url': url})
+        properties = {
+            f['dataFile']['filename']: {'size': f['dataFile']['filesize']}
+            for f in files}
+        entries = list(properties.keys())
+        return {
+            'directories': [],
             'entries': entries,
             'properties': properties,
         }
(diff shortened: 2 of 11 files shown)


storage_service/locations/forms.py
```diff
@@ -116,11 +117,10 @@ class Meta:
 
     def as_p(self):
         # Add a warning to the Dataverse-specific section of the form
-        # FIXME this may not be the best way to add Space-specific warnings
         content = super(DataverseForm, self).as_p()
         content += '\n<div class="alert">{}</div>'.format(
-            _('Integration with Dataverse is currently a beta feature'))
-        return content
+            ('Integration with Dataverse is currently a beta feature'))
+        return mark_safe(content)
```
jrwdunham, line 123: Is there a reason that you removed the `ugettext` (= `_`) translation function from the alert here? I'm guess it was just an omission because the parens were left behind.

storage_service/locations/forms.py
```diff
@@ -273,6 +273,13 @@ def __init__(self, *args, **kwargs):
         # Associated with all enabled pipelines by default
         self.fields['pipeline'].initial = models.Pipeline.active.values_list('pk', flat=True)
         self.fields['default'].initial = self.instance.default
+        # If we are accessing a Dataverse we want to be able to do nice things
+        # with the Transfer Browser. Create a namespace to work with the path
+        # here.
+        if space_protocol == "DV":
+            dv_config = ("Query: \n"
+                 "Subtree: ")
```
jrwdunham: Nitpick but why the linebreak here at all? (Not questioning the literal "\n" but the extra line 281.)

jrwdunham: Also, I think `"DV"` should be `models.Space.DATAVERSE`

storage_service/locations/forms.py
```diff
@@ -273,6 +273,13 @@ def __init__(self, *args, **kwargs):
...
+            self.fields['relative_path'].initial = dv_config
```
jrwdunham, line 281: Maybe it would be a good idea to change the help text in the Location Edit form when displaying the Relative Path field for a Dataverse Transfer Source location. Something like this:

    self.fields['relative_path'].help_text += '. Dataverse-specific relative path help...'

storage_service/locations/models/dataverse.py
```diff
@@ -20,52 +22,120 @@
...
         verbose_name = _l("Dataverse")
-        app_label = 'locations'
+        app_label = "locations"
+
+    ALLOWED_LOCATION_PURPOSE = [Location.TRANSFER_SOURCE]
+
+    def get_query_value(self, key, path, default=None):
+        """Retrieve a value from a string and then remove it where it exists so
+        that we can recurse through it for all the values we need.
+        """
```
jrwdunham, line 76: I think this docstring could be made more clear. How about `Retrieve the value corresponding to ``key`` from the string ``path``, then remove both the key and the value from path so that we can recurse through it for all the values we need, and finally return the value and path as a 2-tuple.`

More importantly than the docstring, I think this method could be problematic. Consider what happens if your dataset value is equal to, or is a substring of, your query string or subtree values. For example, the path `query: drinks_1 subtree: archivematica_1/1` will give you dataset `1`, which you want, but subtree `archivematica_` and query_string: `drinks_`, which you don't want.

jrwdunham: Couldn't you just require that the relative path of a Dataverse location have the form `Query={q}Subtree={s}/` where `q` and `s` may be empty strings? I think that would simplify the logic, though it would make the user do more work.

storage_service/locations/models/dataverse.py
```diff
@@ -20,52 +22,120 @@
...
+            "query:", path, "*")
+        return dataset, subtree, query_string
 
     def browse(self, path):
+        """Fetch the datasets in this dataverse or the files in the dataset
+        referenced in ``path``.
         """
-        Fetch a list of datasets from Dataverse based on the query in the location path.
+        path = path.rstrip('/').replace(self.space.path, "")\
+            .replace("\r\n", " ")
```
jrwdunham: What is the purpose of replacing `self.space.path` with the empty string here?

This actually messed up my testing because I had set the path value of the DV space to `/`. This was then causing DV TS relative paths like `query: drinks_1 subtree: archivematica_1/111` to be converted to `query: drinks_1 subtree: archivematica_1111`.

Since `/` is a valid path value for a Space, isn't the above a real trap for a user? Is removing the space path necessary? When I logged the input `path` value to `browse()` there was no `self.space.path` in it.

storage_service/locations/models/dataverse.py
```diff
@@ -20,52 +22,120 @@
...
+        # Log the path we want to deconstruct now the noise has been removed.
+        LOGGER.info("Path received: %s", path)
+        dataset_id, subtree, query_string = self.get_query_and_subtree(path)
+        LOGGER.info("Dataset ID: %s Subtree: %s Query: %s",
+            dataset_id, subtree, query_string)
+        if dataset_id:
+            return self._browse_dataset(dataset_id)
+        else:
+            self.query_string = query_string
+            self.subtree = subtree
```
jrwdunham: Nitpick but why not just pass these to the method instead of setting them on the instance?

jrwdunham: Also, the above `else:` is unnecessary because a `return` precedes it.

storage_service/locations/models/dataverse.py
```diff
@@ -77,31 +147,61 @@ def browse(self, path):
...
+        """
+        files_in_dataset_path = (
+            '/api/v1/datasets/{dataset_identifier}/versions/:latest'.format(
+                dataset_identifier=dataset_identifier))
+        url = 'https://{}{}'.format(self.host, files_in_dataset_path)
+        params = {'key': self.api_key}
+        LOGGER.debug('URL: %s, params: %s', url, params)
+        response = requests.get(url, params=params)
+        LOGGER.debug('Response: %s', response)
+        if response.status_code != 200:
```
jrwdunham, line 187: Is it possible that this request could return valid 2xx or 3xx codes that are not 200? If so, maybe `response.raise_for_status()` should be used here. Same comment applies to the similar check in `_browse_dataverse`.

jrwdunham: Also, I don't think the entire response text should be logged in the error case. It's a huge string of HTML in the logs and imo it actually makes debugging more difficult.

storage_service/locations/models/dataverse.py
```diff
@@ -77,31 +147,61 @@ def browse(self, path):
...
+            files = data['files']
+        except json.JSONDecodeError:
+            LOGGER.error('Could not parse JSON from response to %s', url)
+            raise StorageException(
+                _('Unable to parse JSON from response to %(url)s')
+                % {'url': url})
+        entries = [f['dataFile']['filename'] for f in files]
+        properties = {
+            f['dataFile']['filename']: {'size': f['dataFile']['filesize']}
+            for f in files}
```
jrwdunham, line 202: Small thing, but if the `properties` dict comp were performed first, then the entries could be derived from it via `entries = list(properties.keys())`. Same comment applies to the similar construction in _browse_dataverse.

storage_service/locations/models/dataverse.py
```diff
@@ -112,63 +212,113 @@ def move_to_storage_service(self, src_path, dest_path, dest_space):
...
         # HACK strip everything that isn't a number
-        src_path = ''.join(c for c in src_path if c.isdigit())
+        src_path = "".join(c for c in src_path if c.isdigit())
         # Verify src_path has to be a number
         if not src_path.isdigit():
-            raise StorageException(_('Invalid value for src_path: %(value)s. Must be a numberic entity_id') % {'value': src_path})
+            raise StorageException(
+                _(
+                    "Invalid value for src_path: %(value)s. Must be a "
+                    "numberic entity_id"
```
jrwdunham: Typo: "numberic"

storage_service/locations/models/dataverse.py
```diff
@@ -112,63 +212,113 @@ def move_to_storage_service(self, src_path, dest_path, dest_space):
...
         try:
-            dataset = response.json()['data']
+            dataset = response.json()["data"]
         except json.JSONDecodeError:
-            LOGGER.error('Could not parse JSON from response to %s', url)
-            raise StorageException('Unable parse JSON from response to %s' % url,)
+            LOGGER.error("Could not parse JSON from response to %s", url)
+            raise StorageException(
+                "Unable parse JSON from response to %s" % url
+            )
```
jrwdunham, line 240: The translation functions are not being consistently used on exception messages. Do we need to translate exceptions? I'm thinking maybe we don't.

storage_service/locations/models/dataverse.py
```diff
@@ -112,63 +212,113 @@ def move_to_storage_service(self, src_path, dest_path, dest_space):
...
         # Fetch all files in dataset.json
-        for file_entry in dataset['latestVersion']['files']:
-            entry_id = str(file_entry['dataFile']['id'])
-            if not file_entry['label'].endswith('.tab'):
-                download_path = os.path.join(dest_path, file_entry['dataFile']['filename'])
-                url = 'https://' + self.host + '/api/access/datafile/' + entry_id
+        for file_entry in dataset["latestVersion"]["files"]:
+            zipped_bundle = False
+            entry_id = str(file_entry["dataFile"]["id"])
+            if not file_entry["label"].endswith(".tab"):
```
jrwdunham: You didn't introduce this, but I always find the `if not COND PATH1 else PATH2` construction confusing.
Can we rearrange this to `if COND PATH2 else PATH1`?

storage_service/locations/models/dataverse.py
```diff
@@ -112,63 +212,113 @@ def move_to_storage_service(self, src_path, dest_path, dest_space):
...
+                )
+                url = "https://{}/api/access/datafile/bundle/{}".format(
+                    self.host, entry_id
+                )
+            LOGGER.debug("URL: %s, params: %s", url, params)
+            response = requests.get(url, params=params, stream=True)
+            with open(download_path, "wb") as f:
+                for chunk in response.iter_content(8192):
+                    f.write(chunk)
+            if zipped_bundle and ".zip" in download_path:
```
jrwdunham: Isn't the second clause in the condition redundant here? I mean, you just added the ".zip" extension in the only path where `zipped_bundle` gets set to `True`, so it's not possible for one condition to be true but not the other.

storage_service/locations/models/dataverse.py
```diff
@@ -112,63 +212,113 @@ def move_to_storage_service(self, src_path, dest_path, dest_space):
...
+    def extract_and_remove_bundle(dest_path, bundle_path):
+        try:
+            unzipper = zipfile.ZipFile(bundle_path, 'r')
+        except zipfile.BadZipfile as e:
+            # Log the error and return without extracting the bundle.
+            # Archivematica may still be able to work with the data returned.
+            LOGGER.info("Bundle '%s' error: %s", bundle_path, e)
+            return
+        unzipper.extractall(dest_path)
+        unzipper.close()
```
jrwdunham: Can you use a context manager in the above call to `ZipFile` so that you don't have to manually close `unzipper`? I.e., `with zipfile.ZipFile(bundle_path, 'r') as unzipper`.

storage_service/locations/tests/test_dataverse.py
```diff
@@ -83,3 +106,43 @@ def test_move_to(self):
...
+            # Retrieve our query our result sets.
+            query_ = test.get("query")
+            res_ = test.get("result")
+
+            # Assign values to: Dataset, Subtree, Query
+            d1, s1, q1 = self.dataverse.get_query_and_subtree(query_)
+            d2, s2, q2 = res_
+            assert d2 == d2
+            assert s1 == s2
+            assert q1 == q2
```
jrwdunham: Nitpick: no need for defensive `.get(` in lines 140,141. Also, lines 144-148 could just be `assert res_ == self.dataverse.get_query_and_subtree(query_)`. Bug: `assert d2 == d2` should be `assert d1 == d2`, but that would be avoided if my first suggestion were followed.

jrwdunham (changes requested):

storage_service/locations/models/dataverse.py
```diff
@@ -20,52 +22,120 @@
...
+            return self._browse_dataverse()
+
+    def _browse_dataverse(self):
+        """Return all datasets in all dataverses (conforming to ``browse``
+        protocol).
 
-        Datasets are considered directories when browsing.
+        Note: we could add a subtree key to the ``params`` dict below to
+        restrict this to returning the datasets within a specified dataverse.
+        For now, we are not doing that.
```
jrwdunham: Is this `Note:` relevant any more? It looks like a subtree value is now being provided in the params dict.

storage_service/locations/forms.py
ross-spencer: Unfortunately I removed it accidentally but a good catch. That being said, I hadn't appreciated what this was doing. Is the way we have used the underscore for the translation functions considered idiomatic do you know?

`from django.utils.translation import ugettext as _, ugettext_lazy as _l`

storage_service/locations/forms.py
ross-spencer: I think this just made visual sense to me comparing it to what we'd see in the form when we look at it, but I appreciate its redundancy.

storage_service/locations/forms.py
ross-spencer: :+1:

storage_service/locations/models/dataverse.py
ross-spencer: I can't remember my original logic for doing this, but testing both ways, it does seem that this isn't needed. I'll work with `path` as is.

storage_service/locations/models/dataverse.py
ross-spencer: Good idea!

storage_service/locations/forms.py
sevein: You should use `ugettext_lazy` in forms and models, but don't aliase it as `_l`, use `_` instead. Otherwise `makemessage` won't find the strings. That's a known but that I fixed in AM here (https://github.com/artefactual/archivematica/pull/1251 -> https://github.com/artefactual/archivematica/pull/1251/commits/351c80f6e553fe94aefd779ec8dbbaf328a49337) but now I realize we have the same issue in SS so I'll be filing an issue soon.

storage_service/locations/forms.py
sevein: ^ that wouldn't be i18n friendly AFAIK.

storage_service/locations/models/dataverse.py
ross-spencer: I'm not sure @jrwdunham. The reason I want to test for `200` is that is the only result we care about that determines whether or not we see any results in the transfer browser. I could do something like follows to provide information more specific about the response code, does that work for you?
```
            if response.status_code != 200:
                try:
                    requests.raise_for_status()
                except requests.HTTPError as err:
                    raise StorageException(
                        _("Unable to fetch datasets from %(url)s, "
                          "error: %(err)s") % {'url': url, "err": err})
```

storage_service/locations/models/dataverse.py
ross-spencer: This is a nice shortcut. Good thinking!

storage_service/locations/models/dataverse.py
ross-spencer: I think this is the only one we've missed so will add it here.

storage_service/locations/models/dataverse.py
ross-spencer: Another good spot!

storage_service/locations/models/dataverse.py
ross-spencer: :+1:

storage_service/locations/models/dataverse.py
ross-spencer: :+1:

storage_service/locations/models/dataverse.py
ross-spencer: :+1:

storage_service/locations/forms.py
ross-spencer: @sevein is there a way we can make it i18n friendly, or should I avoid adding the help text do you think?

sevein: I think it's ok to add the help text here as long as you don't make up messages from pieces (concatenation), e.g. you could redefine it completely here, wrapping the new string with `_()`. You could also try to use interpolation but I'm not sure if it's a good fit in this case.

storage_service/locations/models/dataverse.py
ross-spencer: Hmm, the function just doesn't seem to work at all as anticipated. I will have a think about what a good fix would look like. I'd like to take as much work away from the user here as possible.

ross-spencer: This was a good catch. I've added it to the unit tests and then made a code change to fix it. I'd like to give the user as little as possible to do with configuring this and we've found some good flexibility with the current approach. Let me know what you think of the change, hopefully it is a little better. 

Ultimately, I'd like to see a custom form written for Dataverse where the values can be grabbed from the form elements instead. This was a little bit out of scope for the work done for OCUL this time around. Hopefully there is scope to make this a more robust solution in the near future.

ross-spencer:
@jrwdunham I've made a first pass at these changes following CR. Thanks for all the suggestions. I think the one thing I need to draw your attention to is the removal of a few of the asserts from the Dataverse unit tests. They were testing positional aspects of lists which I think made the tests brittle. If we were to sort the results before displaying them to the user, or add a sort to the Dataverse query string they may easily break. I feel that the other checks in the tests cover enough of what we're looking for from the API calls to demonstrate that the code still works as expected. 

**NB.** The build is still failing due to the Arkivum changes. I'm not sure what the status is on these being fixed.

sevein:
> NB. The build is still failing due to the Arkivum changes. I'm not sure what the status is on these being fixed.

I think that's fixed, but you need to rebase.

storage_service/locations/models/dataverse.py
```diff
@@ -313,12 +307,10 @@ def move_to_storage_service(self, src_path, dest_path, dest_space):
     @staticmethod
     def extract_and_remove_bundle(dest_path, bundle_path):
         try:
-            unzipper = zipfile.ZipFile(bundle_path, 'r')
+            with zipfile.ZipFile(bundle_path, 'r') as unzipper:
+                unzipper.extractall(dest_path)
+                os.unlink(bundle_path)
         except zipfile.BadZipfile as e:
```
sevein: Should we also capture `IOError`? In case `unlink` raises it.

storage_service/locations/models/dataverse.py
```diff
@@ -112,63 +207,110 @@ def move_to_storage_service(self, src_path, dest_path, dest_space):
...
         # Verify src_path has to be a number
         if not src_path.isdigit():
-            raise StorageException(_('Invalid value for src_path: %(value)s. Must be a numberic entity_id') % {'value': src_path})
+            raise StorageException(
+                _(
+                    "Invalid value for src_path: %(value)s. Must be a "
+                    "numeric entity_id"
+                )
+                % {"value": src_path}
+            )
```
sevein: github.com/ambv/black can be pretty verbose at times!

storage_service/locations/models/dataverse.py
```diff
@@ -112,63 +207,110 @@ def move_to_storage_service(self, src_path, dest_path, dest_space):
...
         # Fetch dataset info
-        url = 'https://' + self.host + '/api/datasets/' + src_path
-        params = {
-            'key': self.api_key,
-        }
-        LOGGER.debug('URL: %s, params: %s', url, params)
+        url = "https://{}/api/datasets/{}".format(self.host, src_path)
```
sevein: I've found three places where we're building the URL, e.g. `url = "https://..."`. I recommend to do that from a new private method so that concern is addressed in just one place.

Do you know if the server is always going to be using TLS? I'm asking because the scheme is hard-coded.

storage_service/locations/models/dataverse.py
ross-spencer: :+1:

storage_service/locations/models/dataverse.py
ross-spencer: :+1:

storage_service/locations/models/dataverse.py
ross-spencer: Do you have an opinion @sevein on what the best strategy for managing the scheme should look like? 

In the DV form, we'd specify something like this:

image

So I can parse the URL and pull the scheme out of there, and if there isn't one, add `https` as a default. Does that sound okay?

sevein: That makes sense, @ross-spencer. I recently made a similar change in the pipeline settings, there is a method that could be converted into a mixin and be used here too?

https://github.com/artefactual/archivematica-storage-service/blob/07ae87501b93f4c917f1c0a49dc76ba5c369039d/storage_service/locations/models/pipeline.py#L76-L89

In this case `scheme` defaults to `http` but I guess that you can make it a parameter.

Thanks!

ross-spencer: I've given it a whirl in https://github.com/artefactual/archivematica-storage-service/pull/398/commits/c9b630ba6abcc671033e0e573df65a10afaae930, does that look okay?

sevein: Looks good to me. I hope JD is okay with this OOP-ish solution haha!

The double underscore may not be great because in Python I think they convention is to use them for magic methods, like `__repr__`, etc... I think you could just call it `_generate_dataverse_url`.

storage_service/locations/forms.py
jrwdunham: @sevein I don't understand, does that mean that we can't have both `ugettext` or `ugettext_lazy` in the same module, or does it mean that `makemessage` will only find funcs named `ugettext`, `ugettext_lazy` or `_`?

sevein: `makemessages` is only looking for `_` AFAIK. I think it'd be possible to use both `ugettext` and `ugettext_lazy` in the same module if you use `import` in different scopes but generally what I've found is that people just use either `ugettext` or `ugettext_lazy` but never combined in the same module.

storage_service/locations/models/dataverse.py
jrwdunham: Yeah, this is probably fine. It just makes me uneasy because it's still brittle. I guess I don't have the implicit knowledge of what a Dataverse `dataset`, `subtree`, or `query_string` value can look like, but again it's not hard to imagine an input that breaks expectations here. Compare the happy:

    query: drinks_1 subtree: archivematica_1/1
    ('1', 'archivematica_1', 'drinks_1')

to the sad:

    query: drinks/1 subtree: archivematica_1/1
    ('1 subtree: archivematica_1/1', None, 'drinks')

Assuming the above is too edge-case to worry about at this stage, one thing you might still want to do is avoid the new kwarg and treat all keys the same by doing something like the following:

    import re

    def get_query_value(key, path, default=None):
        path = path.lower()
        value = path.split(key.lower(), 1)
        try:
            value = value[1].strip()
        except IndexError:
            value = default
        if value == "":
            value = default
        if value is None:
            path = path.replace(key, "")
        else:
            path = re.sub(
                re.escape(key) + r"\s*" + re.escape(value), "", path)
        return value, path.strip()

As implied by the above, `get_query_value` could be a `staticmethod` or a function cuz it doesn't need `self`.

storage_service/locations/models/dataverse.py
jrwdunham: Hmm, I'm not sure that this new change is helpful. As it now stands, if status code is > 200 or < 400, it's going to assume the request was successful and *not* raise a `StorageException`. Assuming that is what you want, the code seems more difficult to me to read now. I would suggest either relying on `raise_for_status` alone or go back to how you had it initially.

storage_service/locations/models/dataverse.py
jrwdunham: The `entries = []` line in `_browse_dataverse` is no longer necessary. It could be removed and `entries += list(properties.keys())` could become `entries = list(properties.keys())`.

storage_service/locations/models/dataverse.py
jrwdunham: Nitpick, but the first exception message in `move_to_storage_service` is still not being translated.

storage_service/locations/models/dataverse.py
jrwdunham: +1 to removing one of those underscores.

Also, I don't see why the `slug` and `scheme` kwargs are needed in this method if they're never used:

    $ ack __generate_dataverse_url ./
    storage_service/locations/models/dataverse.py
    121:        url = self.__generate_dataverse_url("/api/search/")
    175:        url = self.__generate_dataverse_url(files_in_dataset_path)
    219:        url = self.__generate_dataverse_url(datasets_url)
    271:                url = self.__generate_dataverse_url(bundle_url)
    277:                url = self.__generate_dataverse_url(datafile_url)
    321:    def __generate_dataverse_url(self, slug="", scheme="https"):

storage_service/locations/models/dataverse.py
```diff
@@ -17,55 +19,115 @@
...
+            value = value[1].strip()
+        except IndexError:
+            value = default
+        if value == "":
+            value = default
+        if value is None:
+            path = path.replace(key, "")
+        else:
+            if dataset:
+                return value, path.replace(key + value, "").strip()
```
jrwdunham: You could avoid the use of the dataset kwarg and treat all keys the same way if you did something like the following:

    +import re
     import zipfile
    
     # Core Django, alphabetical
    @@ -66,7 +67,7 @@ class Dataverse(URLMixin, models.Model):
    
         ALLOWED_LOCATION_PURPOSE = [Location.TRANSFER_SOURCE]
    
    -    def get_query_value(self, key, path, dataset=False, default=None):
    +    def get_query_value(self, key, path, default=None):
             """Retrieve a value from a string and then remove it where it exists so
             that we can recurse through it for all the values we need.
             """
    @@ -81,9 +82,8 @@ class Dataverse(URLMixin, models.Model):
             if value is None:
                 path = path.replace(key, "")
             else:
    -            if dataset:
    -                return value, path.replace(key + value, "").strip()
    -            path = path.replace(value, "").replace(key, "")
    +            path = re.sub(
    +                re.escape(key) + r"\s*" + re.escape(value), "", path)
             return value, path.strip()
    
         def get_query_and_subtree(self, path):
    @@ -95,7 +95,7 @@ class Dataverse(URLMixin, models.Model):
             browser. We can then retrieve the subtree and the query string which
             can be used to return a listing of datasets available in the Dataverse.
             """
    -        dataset, path = self.get_query_value("/", path, dataset=True)
    +        dataset, path = self.get_query_value("/", path)
             subtree, path = self.get_query_value("subtree:", path)
             query_string, path = self.get_query_value("query:", path, default="*")
             return dataset, subtree, query_string

jrwdunham (commented):
I think this is pretty close now. I've added some comments to the changes you've made to my initial CR comments. I hope you can see them. I have not looked through the test module thoroughly, but I ran the tests with and without my `re.sub` suggested change and they pass both ways.

ross-spencer:
Thanks @jrwdunham I have made those changes. I like the changes to the way we're constructing the query string and acknowledge your unease. I think there'll be an opportunity to solve that more holistically with future Dataverse users. RE: The `slug` keyword in the generate url function, I've left that simply to make it explicit we're working with the slug of the Dataverse query string, not the host part. The scheme has been removed. 

You were right about some of your comments not showing up. I am not sure why this happened, unless I did a force push, so I'll be more careful of this while code is being reviewed. Fortunately I had the email from GitHub with the comments inline,

jrwdunham (approved):

PR #400, opened 2018-08-28
Arkivum: fix replication status (ref.#12343)

hakamine:
To get replication status of an uncompressed package use fileInfo API
call (on file bag-info.txt)

The previous way to get the replication status, that used
ingest-manifest/status API call using arkivum_identifier has not been
reliable.


sevein (approved):

hakamine:
77a36861c03563cbbc4977885a4f1a3a1f97d311 in qa/0.x

sromkey:
Since this was cherry-picked into qa/0.x I've added the 1.8 milestone (we're using the 1.8 milestone to track both 1.8 and SS 0.13).

Issue #319, opened 2018-02-07
Problem: there is no way to delete stored DIPs


sromkey:
While you can request the deletion of an AIP from the dashboard and complete the deletion in the Storage Service, there is no such functionality for stored DIPs. I'm of the opinion that the two-step process for deleting AIPs (request in the dashboard, complete in the Storage Service) is not necessary for deletion of stored DIPs because if a DIP is accidentally deleted, it can always be regenerated by re-ingesting the AIP. We should add a delete option to the stored DIPs in the Packages tab.

sevein:
We've agreed to implement it as follows:
> Delete button to be added to the SS so the deletion is initiated from there rather than through the API.

sevein:
For QA, the delete dip button appears in the package list of the Storage Service. The workflow is a bit different from AIP deletion. In this case, a dialog appears requesting the user to confirm the removal of the package.

PR #401, opened 2018-08-29
Delete DIPs

jrwdunham:
UPDATE(@sevein): see comments below for more details on how this is being implemented!

Connected to #319.


sevein:
Adding reviewers so the approach can be validated. JD originally updated `PACKAGE_TYPE_CAN_DELETE` so DIPs could be deleted following the exact same workflow than other package types, i.e. deletion request is submitted, administrator confirms deletion, etc...

As per Sarah's request, I've changed it so the confirmation is not required. The "Delete" links appear next to DIPs in the package list sending the request directly to q new `package_delete` view which performs the deletion. The template has been updated so the deletion button is hidden when the package status is `DELETED`. I was surprised to see that's not the case for other action buttons.

If this looks okay for you let me know, I wasn't entirely sure if this is the right thing to do.

I've added a bunch of tests.

sevein (commented):
CC @artefactual/amcr

ablwr:
Code looks good and it works fine. I wonder if we should have an additional pop-up "Are you sure you want to delete this DIP?", especially because this is such a leap from the existing flow of request deletion, confirm deletion steps we have for AIPs.

sevein:
@ablwr, pop-up added! I've used the bs2 dialog support, this is how it looks like...

image

ablwr:
:tada: :tada: :tada:

ablwr:
This works for me in Chrome but not in Firefox. In Firefox, it just doesn't pop up and goes straight through to deleting.

Another thing... after I hit delete, "download" is still visible. We should hide it, right? The status does update to Deleted though. Will those linger around forever, how would someone clear those out if that's their intent?

sevein:
> This works for me in Chrome but not in Firefox. In Firefox, it just doesn't pop up and goes straight through to deleting.

Ops, fixed now, thanks for testing that!

> Another thing... after I hit delete, "download" is still visible. We should hide it, right? The status does update to Deleted though. Will those linger around forever, how would someone clear those out if that's their intent?

Ok, see the extra commit. It was also affecting AIPs.

ablwr:
@sevein OK it works! Although it is firing twice for me in both Firefox/Chrome if I hit cancel, it asks again for me to approve or cancel. Lemme look at the js...

ablwr:
Oops, forgot to press "Comment" here. :laughing: 

Yeah, the JS fires for every DIP on the page instead of the one that is being clicked!

sevein:
:man_facepalming: sorry I'll take a look, my js skillz are gone haha

sevein:
Ok @ablwr I've changed the code a bit, the previous attempt was a hot mess because the callback was being triggered on each match. Now using `jQuery.each` instead.

```javascript
    $("form.submit-confirm").each(function(index, target) {
      var $form = $(target);
      var $modal = $form.find(".confirm-modal");
      $form.on("click", "button[type=submit]", function(event) {
          var $button = $(event.target);
	  if ($button.parents(".confirm-modal").length) {
            return true;
          }
          event.preventDefault();
          $modal.modal("show");
      });
```

ablwr (approved):

ablwr:
Just tested and seems all good now!

PR #402, opened 2018-08-29
Stream HTTP response to avoid loading large files into memory

marktriggs:
Addresses https://github.com/archivematica/Issues/issues/113

### DIFF
--- a/storage_service/locations/fixtures/dataverse2.json
+++ b/storage_service/locations/fixtures/dataverse2.json
@@ -1,0 +1,42 @@
+[
+{
+    "fields": {
+        "host": "demodv.scholarsportal.info",
+        "api_key": "51c64df4-abd7-4613-af21-6a68715dca92",
+        "space": "d18a2763-60f0-4273-b30c-caa5d14d6d32",
+        "agent_name": "Archivematica Test Dataverse",
+        "agent_type": "TestOrganisation",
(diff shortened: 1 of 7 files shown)


ross-spencer:
This looks good @marktriggs I will cherry-pick it into the am dev branch in gitolite to test. 

I didn't realize it would be this simple via the requests library so I've just been looking at the docs. Do you think it necessary to add `with` to the request, e.g. via the docs:

> If you set stream to True when making a request, Requests cannot release the connection back to the pool unless you consume all the data or call Response.close. This can lead to inefficiency with connections. If you find yourself partially reading request bodies (or not reading them at all) while using stream=True, you should make the request within a with statement to ensure it’s always closed:
```
with requests.get('http://httpbin.org/get', stream=True) as r:
    # Do things with the response here.
```

marktriggs:
Ah nice catch @ross-spencer; I think adding that `with` block would be the right thing to do.  My loop should have consumed all the data anyway, but an IO error writing to the file could get it into trouble and leak that connection, so best to be safe.

sromkey:
Connected to https://github.com/archivematica/Issues/issues/113

ross-spencer:
Cherry-picked for commit: de6ff565a503edb2fedb81348a078018225ea298 (Dataverse transfer feature). Thanks once again @marktriggs.

PR #403, opened 2018-08-29
Comment out DSpace REST

jrwdunham:
The DSpace REST-type space is not ready to be released. This commit comments out the relevant lines of code so that it will no longer be possible to create DSpace REST spaces either via the GUI or the API.

Connected to archivematica/Issues#126


jrwdunham:
The Travis failures are pre-existing and are being dealt with in https://github.com/archivematica/Issues/issues/128

sevein (approved):

PR #404, opened 2018-08-29
Fix flake8 and pytest errors in qa/0.x

jrwdunham:
- [x] Fix flake8 errors
- [x] Fix pytest failures:
  storage_service/locations/tests/test_arkivum.py::TestArkivum::test_update_package_status_uncompressed
  storage_service/locations/tests/test_package.py::TestPackage::test_fixity_amber_arkivum
  storage_service/locations/tests/test_package.py::TestPackage::test_fixity_failure_arkivum
storage_service/locations/tests/test_package.py::TestPackage::test_fixity_scheduled_arkivum
  storage_service/locations/tests/test_package.py::TestPackage::test_fixity_success_arkivum

Connected to archivematica/Issues#128
Cf. [PR 400](https://github.com/artefactual/archivematica-storage-service/pull/400)

### DIFF
--- a/storage_service/locations/models/arkivum.py
+++ b/storage_service/locations/models/arkivum.py
@@ -177,14 +177,14 @@ class Arkivum(models.Model):
         if package.is_compressed:
             url = 'https://' + self.host + '/api/2/files/release/' + package.misc_attributes['arkivum_identifier']
         else:
-            # To get replication status of an uncompressed package, 
+            # To get replication status of an uncompressed package,
             # use files/fileInfo API call (on file bag-info.txt)
-            # instead of ingest-manifest/status call with arkivum_identifier 
+            # instead of ingest-manifest/status call with arkivum_identifier
             # (which has not been reliable)
             location = Location.objects.get(uuid=package.current_location_id)
             url = 'https://' + self.host + '/api/2/files/fileInfo/' + \
-                   location.relative_path + '/' + \
-                   package.current_path + '/bag-info.txt'
+                location.relative_path + '/' + \
+                package.current_path + '/bag-info.txt'
 
         LOGGER.info('URL: %s', url)
         try:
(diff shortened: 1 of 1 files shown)


hakamine:
@jrwdunham Thank you for opening the issue and the PR. The pytest failures revealed that the fix in PR #400 was inadvertently affecting other parts of the application (it was modifying the return value of the function _get_package_info() for uncompressed packages). The fix has been reimplemented so that the interface of _get_package_info() is not affected.

storage_service/locations/models/arkivum.py
```diff
@@ -207,6 +200,39 @@ def _get_package_info(self, package):
...
+        Return information for the bag-info.txt from a package in Arkivum.
+        To be used as a fallback to obtain replication status of an
+        uncompressed package when it is not returned by the
+        ingest-manifest/status call in _get_package_info()
+        """
+        location = Location.objects.get(uuid=package.current_location_id)
+        url = 'https://' + self.host + '/api/2/files/fileInfo/' + \
+            location.relative_path + '/' + \
+            package.current_path + '/bag-info.txt'
+        LOGGER.info('URL: %s', url)
```
jrwdunham: I have confirmed that `location = Location.objects.get(uuid=package.current_location_id)` can be replaced with the simpler `location = package.current_location`, which may not incur an additional db query.

Also, I think we may want to strip leading and trailing forward slashes from the variables in the URL in order to avoid double forward slashes. Something like the following should work:

    url = 'https://' + self.host + '/api/2/files/fileInfo/' + \
        location.relative_path.strip('/') + '/' + \
        package.current_path.strip('/') + '/bag-info.txt'

jrwdunham (commented):
I have one small comment which I think should be addressed, but I think this is pretty much good to be merged (though I am the author of one of the commits so I can't approve :)

storage_service/locations/models/arkivum.py
hakamine: @jrwdunham thank you! Fixed as suggested.

sevein (approved):
Thanks @hakamine!

hakamine:
also cherry-picked to qa/0.9.x-moma

PR #406, opened 2018-08-30
DSpace REST form: mark HTML as safe

jrwdunham:
Adds `django.utils.safestring.mark_safe` so that `DSpaceRESTForm.as_p` returns a `django.utils.safestring.SafeText` instance instead of a string. Previously (before this was implemented), if the form had to be re-rendered because it was invalid or if the user were trying to edit an existing DSpace REST space, then the HTML would be escaped and would be displayed to the user as unusable text.

Connected to archivematica/Issues#124

### DIFF
--- a/storage_service/locations/forms.py
+++ b/storage_service/locations/forms.py
@@ -155,7 +155,8 @@ class DSpaceRESTForm(forms.ModelForm):
               ' being sent, unless the metadata is included in the packages'
               ' METS file in a dmdSec.'),
             needle)
-        return content.replace(needle, as_fields_alert)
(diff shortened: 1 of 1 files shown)


jrwdunham:
**Note: in order to test this, first re-enable the DSpace REST space by modifying https://github.com/artefactual/archivematica-storage-service/blob/dev/issue-124-fix-dspace-rest-space-form-html/storage_service/locations/models/space.py#L143 so that `BETA_PROTOCOLS` has the value `set()`.**

sevein (approved):

PR #407, opened 2018-08-31
Stop using JSON headers in DSpace login request

jrwdunham:
Stops using 'application/json' headers in the login request to DSpace in the DSpace REST space. Use of these headers was preventing login from working. Also improves DSpace REST exception messages for debugging.

Connected to archivematica/Issues#123


jrwdunham:
**Note 1:** in order to test this, first re-enable the DSpace REST space by modifying https://github.com/artefactual/archivematica-storage-service/blob/dev/issue-124-fix-dspace-rest-space-form-html/storage_service/locations/models/space.py#L143 so that BETA_PROTOCOLS has the value set().

**Note 2:** the Travis failures are pre-existing and should be fixed by https://github.com/artefactual/archivematica-storage-service/pull/404

sevein (approved):

PR #412, opened 2018-09-16
Fix use of gettext decorator

sevein:
`makemessages` won't be able to find strings unless they're passed to
`_()`, other forms are not accepted.

Connects to archivematica/Issues#172.


jraddaoui (approved):
Please, don't squash.

PR #417, opened 2018-10-09
Handle xml parse exception if offline

ross-spencer:
When the Archivematica Storage Service cannot access the Library of
Congress METS XSD file for any reason, the validation of pointer files
will fail, except, it is not a fatal error. Warn in the storage
service logs, and return the pointer file back to the pipeline.

Connected to artefactual/archivematica#1050

@ablwr and @sevein I do not think this resolves the issues seen today with the LoC website down, so we should consider other mechanisms to manage that, but I don't believe this failure in the Storage Service should cause the pipeline to error and fail indexing. Looking forward to discussing further. 

Via Slack - @sromkey noted some ambiguity in the solution here. We think this PR resolves a big part of the issue of a user being offline, notably, the SS not completing the generation of a pointer file and returning nothing to Archivematica. The store AIP process would then fail but we avoid that here. But... two further questions now AM can deal with this more gracefully are: Do we need to validate having studiously constructed a pointer file? What does validation look like if we are offline? 

We recognize that this is a likely use-case for a number of users.

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -1237,15 +1237,26 @@ class Package(models.Model):
         for agent in premis_agents:
             mets_fs_entry.add_premis_agent(agent.serialize())
         pointer_file.append_file(mets_fs_entry)
-        # 3. Validate the pointer file
+        # Validate the pointer file
         if validate:
-            is_valid, report = metsrw.validate(
-                pointer_file.serialize(), schematron=metsrw.AM_PNTR_SCT_PATH)
-            if is_valid:
-                LOGGER.info('Pointer file constructed for %s is valid.', self.uuid)
-            else:
-                LOGGER.warning('Pointer file constructed for %s is not valid.\n%s',
-                            self.uuid, metsrw.report_string(report))
+            try:
+                is_valid, report = metsrw.validate(
+                    pointer_file.serialize(),
+                    schematron=metsrw.AM_PNTR_SCT_PATH)
+            except etree.XMLSchemaParseError as err:
+                # It has been observed that this function can fail validating,
+                # for example against the Library of Congress' (LoC) XSD for
+                # METS if the LoC server is down it will fail. We don't want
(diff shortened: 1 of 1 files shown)


ablwr (approved):

storage_service/locations/models/package.py
```diff
@@ -1237,15 +1237,26 @@ def create_pointer_file(self,
...
+            except etree.XMLSchemaParseError as err:
+                # It has been observed that this function can fail validating,
+                # for example against the Library of Congress' (LoC) XSD for
+                # METS if the LoC server is down it will fail. We don't want
+                # this to happen so if the pointer file xml is not valid at
+                # this point in time we need to rely on a different mechanism
+                # to manage that.
+                LOGGER.warning("Failed to parse validation schema: %s", err)
+                return pointer_file
+            if not is_valid:
```
sevein, line 1255: Not sure if I'm missing something here but isn't this going to raise a `NameError` since `is_valid` is undefined when   `etree.XMLSchemaParseError` is raised?

ross-spencer: I'm expecting `is_valid` to either be declared and then an error to be raised and then we return. `is_valid` remains in scope when we test `if not is_valid`. 

That's not to say it is correct @sevein but if there is potential for something to go wrong here then perhaps a better way to approach this is in the https://github.com/archivematica/issues log with with a complete bug report? This PR is already closed and was merged six months back.

sevein: I wasn't sure if it was a bug!
I think it was an oversight on my part, you're returning inside the except.
I've no idea why I landed into this PR, I was doing some tidying yesterday and realized I missed this change.
Thanks! :heart_eyes:

ross-spencer: Fair dos @sevein!! You know previously this snippet from me would have been far more verbose somehow - but someone taught me about the happy path :wink: (which I think (i hope!) this might be a small demonstration of :smile:)

PR #422, opened 2018-10-25
Location: disallow move based on purpose

sevein:
The [`posix_move` op](https://github.com/artefactual/archivematica-storage-service/pull/320) should not be always allowed for certain locations. This commit ensures that it's not used when the original location is backlog or AIP storage.

Connects to https://github.com/archivematica/Issues/issues/270.

### DIFF
--- a/storage_service/locations/models/location.py
+++ b/storage_service/locations/models/location.py
@@ -41,6 +41,12 @@ class Location(models.Model):
     TRANSFER_SOURCE = 'TS'
     REPLICATOR = 'RP'
 
+    # List of purposes where moving is not allowed.
+    PURPOSES_DISALLOWED_MOVE = (
+        BACKLOG,
+        AIP_STORAGE,
+    )
+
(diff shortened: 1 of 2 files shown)


sevein:
Code reviewers! This is a simple fix that mitigates a side effect of https://github.com/artefactual/archivematica-storage-service/pull/320 for certain locations.

We implemented `posix_move` so we didn't have to rsync from the "source" to "internal" and from "internal" to "currently processing". In the long run we probably want to revisit this.

storage_service/locations/models/location.py
```diff
@@ -41,6 +41,12 @@ class Location(models.Model):
     TRANSFER_SOURCE = 'TS'
     REPLICATOR = 'RP'
 
+    # List of purposes where
```
jraddaoui: where ... what? ;)

jraddaoui (approved):
Looks good! Just a minor comment in a comment.

PR #423, opened 2018-10-30
Nginx: use HTTP 1.1 in proxy

sevein:
This commits sets the HTTP protocol version for proxying to 1.1. Nginx
uses 1.0 by default.

This seems to mitigate the problem described in the original issue. However,
the reasons are still being investigated.

This is connected to https://github.com/archivematica/Issues/issues/280.

Similar updates may be required in am-packbuild and ansible-archivematica-src repos.

### DIFF
--- a/install/storage
+++ b/install/storage
@@ -19,6 +19,7 @@ server {
 
     proxy_read_timeout 172800s;
     proxy_pass http://archivematica_storage_service_backend;
+    proxy_http_version 1.1;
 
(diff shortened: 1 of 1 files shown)


mamedin (approved):
LGTM, this is a setting that has no inconveniences and adds several improvements:

https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html#sec8

PR #424, opened 2018-10-30
Avoid access to non existing pointer files

jraddaoui:
Some packages may not have pointer files, like uncompressed AIPs, and
the `get_pointer_instance` call will return `None` in those cases.

Connects to https://github.com/archivematica/Issues/issues/294.


ross-spencer (approved):
Looks like a sensible set of changes. LGTM! :+1:

sevein (approved):

jraddaoui:
d1f44a2 in stable/0.12.x

PR #425, opened 2018-10-31
 Support multiple GnuPG binary names

sevein:
Also, `gnupg1` is added to bionic's osdeps to ensure that GnuPG v1 is also installed. The binary is named `gpg1` which is now matched first by Storage Service.

This is connected to https://github.com/archivematica/Issues/issues/306.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,6 +15,7 @@ RUN set -ex \
 	&& apt-get update -qq \
 	&& apt-get install -qq -y --no-install-recommends \
 		gettext \
+		gnupg \
 		p7zip-full \
 		rsync \
 		unar \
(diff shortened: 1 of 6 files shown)


storage_service/common/which.py
```diff
@@ -0,0 +1,70 @@
+"""Backport of ``shutil.which``.
+
+Extracted from ``Lib/shutil.py`` (Python 3.7).
```
jraddaoui, line 3: Did you do any changes to make it work in Python 2.7? I'd be nice to point them out in here if that's the case.

Dockerfile
```diff
@@ -15,6 +15,7 @@ RUN set -ex \
 	&& apt-get update -qq \
 	&& apt-get install -qq -y --no-install-recommends \
 		gettext \
+		gnupg \
```
jraddaoui, line 18: I read somewhere in our code that installing `rng-tools` is useful to gain entropy while generating the keys, maybe we want to install that too.

jraddaoui (approved):

Dockerfile
sevein: It's not listed here because it's a daemon that you'd typically run somewhere else. E.g. in our development environment we just advise to run it in the host (the device is shared automatically with the container). In production environments I suspect you'd do similarly unless you want to run rng inside a container that can write to the device.

storage_service/common/which.py
sevein: To my surprise, it worked out of the box. I've just added their tests too, which needed some tweaking but they're working!

PR #408, opened 2018-09-01
Add move AIP endpoint

jambun:
Implements feature described in:
https://github.com/archivematica/Issues/issues/71

A reimplementation of:
https://github.com/artefactual/archivematica-storage-service/pull/143

PR #428, opened 2018-12-07
Add move aip endpoint

ablwr:
This pull request supercedes both https://github.com/artefactual/archivematica-storage-service/pull/143 and https://github.com/artefactual/archivematica-storage-service/pull/408 and I will close both of them after submitting this PR. This PR is connected to https://github.com/archivematica/Issues/issues/71.

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -598,6 +598,9 @@ class PackageResource(ModelResource):
             # Reingest
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/reingest%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('reingest_request'), name="reingest_request"),
 
+            # Move
+            url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/move%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('move_request'), name="move_request"),
+
             # FEDORA/SWORD2 endpoints
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/sword%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('sword_deposit'), name="sword_deposit"),
             url(r"^(?P<resource_name>%s)/(?P<%s>\w[\w/-]*)/sword/media%s$" % (self._meta.resource_name, self._meta.detail_uri_name, trailing_slash()), self.wrap_view('sword_deposit_media'), name="sword_deposit_media"),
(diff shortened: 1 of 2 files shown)


ablwr:
I tested this locally by:

- Making a new AIP Storage location in the Storage Service
- Hitting the API like so,
curl -d 'location_uuid=NEW_LOCATION_UUID' -s -u test:test 'http://localhost:62081/api/v2/file/AIP_TO_MOVE/move/'

I also tested out some improper uses of the command to see the different error messages, like "New location must have the same purpose as the current location."

storage_service/locations/api/resources.py
```diff
@@ -1082,6 +1085,99 @@ def reingest_request(self, request, bundle, **kwargs):
...
+                            to the current location'),
+            }
+            return self.create_response(request,
+                                        response,
+                                        response_class=http.HttpBadRequest)
+
+        if location.purpose != file.current_location.purpose:
+            response = {
+                'error': True,
+                'message': _('New location must have the same purpose \
```
ablwr: @sevein Perhaps I should modify this line in a different way so that it is both pep8-compliant but also doesn't create a space for the end user. I didn't google how to do that :stuck_out_tongue_closed_eyes:

storage_service/locations/api/resources.py
```diff
@@ -1064,24 +1067,127 @@ def reindex_request(self, request, bundle, **kwargs):
...
+    @_custom_endpoint(expected_methods=['post'])
+    def move_request(self, request, bundle, **kwargs):
+        r"""Request to move a stored AIP.
+
+        To test the move request, use the following sample command.
+        curl -d 'location_uuid=688c6909-9b97-43b0-9a32-4c6247bc0e84' \
+         -s -u test:test \
+        '/api/v2/file/3e8387db-2614-4215-8a19-f4b121ba1777/move/'
+        """
+        file = bundle.obj
```
sevein: `file` is a builtin function in Py2, you should use a different var name! You could use `package` which is the name of the model anyways, e.g. that's the name used in other methods of this class.

storage_service/locations/api/resources.py
```diff
@@ -1064,24 +1067,127 @@ def reindex_request(self, request, bundle, **kwargs):
...
 
+    @_custom_endpoint(expected_methods=['post'])
+    def move_request(self, request, bundle, **kwargs):
+        r"""Request to move a stored AIP.
+
+        To test the move request, use the following sample command.
+        curl -d 'location_uuid=688c6909-9b97-43b0-9a32-4c6247bc0e84' \
+         -s -u test:test \
+        '/api/v2/file/3e8387db-2614-4215-8a19-f4b121ba1777/move/'
+        """
```
sevein, line 1101: There is a way to inject code snippets in docstrings using rst, see for example: https://github.com/artefactual/archivematica/blob/stable/1.8.x/src/archivematicaCommon/lib/bindpid.py. It requires that use of double quote and extra indentation.

storage_service/locations/api/resources.py
```diff
@@ -1064,24 +1067,127 @@ def reindex_request(self, request, bundle, **kwargs):
...
+            }
+            return self.create_response(request,
+                                        response,
+                                        response_class=http.HttpBadRequest)
+
+        location_uuid = request.POST.get('location_uuid')
+
+        if not location_uuid:
+            return http.HttpBadRequest(_('All of these fields must be provided:'
+                                         'location_uuid'))
```
sevein, line 1119: There is no white space after the colon.

storage_service/locations/models/package.py
```diff
@@ -340,6 +341,70 @@ def _update_quotas(self, space, location):
...
+            # Add USE="Archival Information Package" to fileGrp.  Required for
+            # LOCKSS, and not provided in Archivematica <=1.1
+            if root.find('.//mets:fileGrp[@USE="Archival Information Package"]',
+               namespaces=utils.NSMAP) is None:
+                root.find('.//mets:fileGrp', namespaces=utils.NSMAP)\
+                    .set('USE', 'Archival Information Package')
+
+            with open(self.full_pointer_file_path, 'w') as f:
+                f.write(etree.tostring(root, pretty_print=True))
+
```
sevein, line 387: Have you considered using metsrw here? It was agreed that further development would use the new library when possible. Do you wan to give it a try? It's already installed and used in a few places.

storage_service/locations/models/package.py
```diff
@@ -340,6 +341,70 @@ def _update_quotas(self, space, location):
         location.used += self.size
         location.save()
 
+    def move(self, to_location):
+        """Move the package to location."""
```
sevein, line 345: A few questions:
- Should this operation be restricted for certain location (purpose)?
- Should this operation be restricted for non-admin users?
- What would happen if I try to operate on a package that is being moved? E.g. say I try to download the package while the move operation is still ongoing? Do we want to protect users from that scenario?

ablwr:
@sevein Moving your questions to the main thread so they don't get lost in the commit updates:

>     Should this operation be restricted for certain location (purpose)?

This does restrict moves to only being to similar types of locations. I think that's wise. But is there a use case for making it more broad? We could also start with this restricted method and loosen it up later if it becomes a need.

>     Should this operation be restricted for non-admin users?

Ah, perhaps! I imagine similar actions are given that higher clearance requirement.

>     What would happen if I try to operate on a package that is being moved? E.g. say I try to download the package while the move operation is still ongoing? Do we want to protect users from that scenario?

I think we definitely don't want to lock processes up, but I can set things to be stored in a "processing" setting while in movement and update it when the move is complete. I think this was recommended by hudmol too.

OK I gotta learn about metsrw and I'll update my PR.

ablwr:
I spent enough time with metsrw to realize there is a function (not written with metsrw) that does everything I need it to do at the end of this move function (in packages).

ablwr:
And while testing, I tested that AIPs set themselves as a MOVING state when they are being moved, so they can't otherwise be acted upon.

Offline @sevein and I discussed that the basic Storage Service credentials should be sufficient to protect from average users hitting the API, as people given access to the Storage Service are usually at the administrative level.

I think this is ready for re-review!

storage_service/locations/api/resources.py
```diff
@@ -1064,24 +1067,125 @@ def reindex_request(self, request, bundle, **kwargs):
...
+                             'the current location - %s'
+                             % package.current_location.purpose),
+            }
+            return self.create_response(request,
+                                        response,
+                                        response_class=http.HttpBadRequest)
+
+        number_matched = (Package.objects
+                          .filter(id=package.id, status=Package.UPLOADED)
+                          .update(status=Package.MOVING))
```
sevein: We'd usually write this like:

```python
        number_matched = Package.objects \
            .filter(id=package.id, status=Package.UPLOADED) \
            .update(status=Package.MOVING))

storage_service/locations/api/resources.py
sevein: This hasn't been fixed yet.

storage_service/locations/api/resources.py
sevein: You can use rst literals and snippets like in rst, Python docstrings are rst!
https://github.com/artefactual/archivematica-storage-service/pull/428#discussion_r242352938

sevein (changes requested):
Almost there! Also, can you squash?

storage_service/locations/api/resources.py
ablwr: Instead of giving an example command, I changed this to look like all the other definitions/descriptions, which were arranged like this!

sevein (approved):

PR #431, opened 2018-12-18
Add black formatting tool

sevein:
This pull request introduces the black formatting tool that @ross-spencer was experimenting with at some point.

I've changed the configuration of tox with new env called `linting` which uses `pre-commit` to run both `black` and `flake8`. The good thing about this approach is that `pre-commit` is also developer friendly (it can be installed so the checks run during git-commit) while not mandatory.
- https://github.com/ambv/black
- https://github.com/pre-commit/pre-commit

Thoughts?

Connects to https://github.com/archivematica/Issues/issues/393.


ablwr:
This would be a big one-time change but overall I think it'll help us move faster, give more accurate code reviews, and spend more time focusing on the code itself instead of fighting with the linter, yeah.

ross-spencer:
Cool, so I did have to do some reading around this today, of course, no objections, a big fan of `go fmt` in `go`! That being said, a few thoughts:

* I think it would be prudent to make a change like this as the final commit before the release candidate so that there is a useful diff for all the features introduced in the release prior to that, and regression-testing captures any potential issues introduced even if we're not expecting anything to break. It's easier to roll-back at that point, than at any point in-between where features are straddling either side.

* There should be a corresponding commit for the `CONTRIBUTING.md` with dev guidelines (how-to run, expectations, and rationale for some of the choices here):

I tested with:

* `tox -e linting`
* `pre-commit run --all-files`

And those offer two good approaches for users running this code. Tests aren't expected to pass standalone any more outside of the Docker- or travis environments, afaik.

* I recognize that you're using the defaults for line-length (88) https://github.com/ambv/black#line-length and the Flake8 `B950` setting. I am conflicted about this, but think ultimately the default approach is the most pragmatic, i.e. contributors don't have to use additional args if running `black` alone to work with the code. 

Overall I think we'll benefit from this change.

sevein:
Thanks for the feedback.

PR #432, opened 2018-12-24
Update metsrw to v0.2.4

sevein:
Connects to https://github.com/archivematica/Issues/issues/260.

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -14,7 +14,7 @@ gunicorn==19.9.0
 jsonfield==2.0.1
 logutils==0.3.4.1
 lxml==3.7.3
-metsrw==0.2.0
+metsrw==0.2.4
 ndg-httpsclient==0.4.2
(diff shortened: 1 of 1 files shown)


ross-spencer (commented):
Might be related to my question in the `metsrw` review, but we've normally been referring to the package version, https://github.com/artefactual/archivematica/blob/stable/1.8.x/src/MCPClient/requirements/base.txt#L9 we we want to do the same in the Storage Service?

sevein:
Yay! It's just that the usual process is to submit the PRs pointing to the commit so we enable testing and people can confirm the behaviour before making the actual release upstream.

sevein:
I should have mentioned that in the description, sorry.
PR is meant to be updated pointing to the final tag, once the fix is confirmed!

ross-spencer (approved):
Ah! That makes sense. Useful to see the procedure play out. Thanks for the info! :+1:

PR #433, opened 2018-12-25
Update to PREMIS 3

ablwr:
This is connected to https://github.com/archivematica/Issues/issues/370. Along with https://github.com/artefactual/archivematica/pull/1323. Just like with those, though, there needs to be a big update in the tests, and some other bugs need to  be ironed out.


sevein:
We've decided to close this for now. Our mail goal right now is to update Archivematica METS to PREMIS 3. Updating the pointer files could be done later.

PR #438, opened 2019-03-06
Support BagIt in transfers

sevein:
Update `_parse_mets` so baggited transfers are indexed properly.
It is backward compatible - should be able to handle non-baggited transfers.
It's been also refactored so it uses `metsrw`.

Connects to https://github.com/archivematica/Issues/issues/444.
Connects to https://github.com/archivematica/Issues/issues/446.


storage_service/locations/fixtures/tiny_transfer/metadata/submissionDocumentation/METS.xml
```diff
@@ -0,0 +1,65 @@
...
+    </mets:div>
+  </mets:structMap>
+  <mets:structMap LABEL="processed" TYPE="physical">
+    <mets:div TYPE="directory" LABEL="TinyTransfer-328f0967-94a0-4376-bf92-9224da033248">
+      <mets:div TYPE="directory" LABEL="objects">
+        <mets:fptr FILEID="file-d8c71be6-4088-49c0-91f0-11a4b02dfd5b"/>
+      </mets:div>
+    </mets:div>
+  </mets:structMap>
+</mets:mets>
```
sevein, line 99: This transfer is incomplete but carries enough information for the existing tests.

storage_service/locations/tests/test_package.py
```diff
@@ -126,9 +129,9 @@ def test_fixity_failure(self):
         package = models.Package.objects.get(uuid='9f260047-a9b7-4a75-bb6a-e8d94c83edd2')
         success, failures, message, timestamp = package.check_fixity()
         assert success is False
-        # Failures are: missing file (dne.txt), bad checksum (dne.txt, test.txt, manifest-md5.txt)
-        assert len(failures) == 4
-        assert message == 'invalid bag'
+        assert len(failures) == 1
+        assert isinstance(failures[0], bagit.FileMissing)
+        assert message == 'Bag validation failed'
```
sevein: The version bump of bagit==1.7.0 changed the way errors are returned.

storage_service/locations/models/package.py
```diff
@@ -1524,16 +1524,13 @@ def compress_package(self, algorithm, extract_path=None):
 
         return (compressed_filename, extract_path)
 
-    def _parse_mets(self, prefix=None, relative_path=['metadata', 'submissionDocumentation', 'METS.xml']):
+    def _parse_mets(self, prefix):
```
cole, line 1728: I think this is a good opportunity to move this function over to using `metsrw` rather than its own logic if possible. If `metsrw` doesn't work for for this case (and it should result in a simpler function) maybe we can make some improvements to it in future.

storage_service/locations/models/package.py
```diff
@@ -2841,3 +2842,12 @@ def write_pointer_file(pointer_file, pointer_file_path):
     if not os.path.isdir(pointer_dir_path):
         os.makedirs(pointer_dir_path)
     pointer_file.write(pointer_file_path, pretty_print=True)
+
+
+def _is_bagit(path):
+    """Determine whether ``path`` is a BagIt package."""
+    try:
+        bagit.Bag(path)
```
cole, line 3215: 👍 looking a the bagit package, this seems like this is a good test — it checks for all the files a bag needs without running full validation. One thing I didn't think about is it may be slow for large bagit manifests, but I think if that becomes a problem we can work to optimize it later.

storage_service/locations/tests/test_package.py
```diff
@@ -238,3 +241,57 @@ def test_extract_file_aip_from_compressed_aip(self):
...
+    It uses ``subprocess`` and ``rsync``.
+    """
+
+    def setUp(self):
+        self.tmp_dir = tempfile.mkdtemp()
+
+    def tearDown(self):
+        shutil.rmtree(self.tmp_dir)
+
+    def _rsync_fixture(self, fixture_dir, dst, make_bagit=False):
```
cole: Does `shutil.copytree` work instead of rsync? It seems better to avoid a dependency on rsync if we can (I know it's in basically all distros, but it's not core, and it requires the use of subprocess).

cole (commented):
Looks good overall!

storage_service/locations/tests/test_package.py
sevein: Cool, it works.

storage_service/locations/models/package.py
sevein: Good point. It fails pretty fast when it's not a BagIt, but it took more than a minute to open one with 500k files.

```
$ time python3 -c "import bagit; bagit.Bag('/home/jesus/archivematica-sampledata')"
60.18user 12.11system 1:12.73elapsed 99%CPU (0avgtext+0avgdata 441052maxresident)k
624inputs+0outputs (0major+140945minor)pagefaults 0swaps
```

sevein:
Ready for CR again. Now `Package._parse_mets` is using metsrw and its latest release v0.3.0.
SS is ready for baggited transfers!

FYI we can't merge until the issues get a milestone but we'll probably get that done on Monday.

PR #441, opened 2019-03-25
WIP: Develop integration testing around handling AM package-types and make package-extract more robust

ross-spencer:
This commit attempts to make reingest more robust where a pointer file isn't available.

The two biggest changes are:

* Teaching the storage service to use the File command when a pointer file isn't available
* Integration tests attempting to prove the consistent behavior of the storage service for all package types that it is possible to create in Archivematica. 

The fixtures for the unit tests are stored as follows:
```
compression_fixtures/
├── 7z.bz-b95b4ad1-c2e0-4fed-b02a-e124f3848cdb.7z
├── 7z.copy-39166cf0-b7c5-414c-b73e-da378142b692.7z
├── 7z.lzma-37ddac23-f41d-46d9-a204-b77491ef2a03.7z
├── pointer_files
│   ├── pointer.37ddac23-f41d-46d9-a204-b77491ef2a03.xml
│   ├── pointer.39166cf0-b7c5-414c-b73e-da378142b692.xml
│   ├── pointer.78bf60f4-018c-4298-b015-7a57bd785ad4.xml
│   └── pointer.b95b4ad1-c2e0-4fed-b02a-e124f3848cdb.xml
├── tar-0b97c973-6be2-4aff-96f1-d7590a294b80.tar
└── tar.bz-78bf60f4-018c-4298-b015-7a57bd785ad4.tar.bz2
1 directory, 9 files
```

**Notes**

Because coverage hasn't always proven perfect in Archivematica I have designed each commit as well as possible to walk the reader through the changes. You should see:

* Moving compression handling code related to re-ingest moved to its own module,
* Pointers updated to where those functions were being called across the storage service.
* Teaching the compression handling module to use `file`.
* Unit tests to validate the output of `package.py` for each type of package available in Archivematica.
* Other commits along the way should represent minor tweaks where things didn't immediately work. 

**Additionally**

This change as noted below might be more appropriate for `1.10.0`. In which case, I can rebase and format with `black`. There may be pieces to scavenge from here but as an exploration, it doesn't look trivial to ask the storage service to deal with packages differently without a pointer file for direction.

Connected to archivematica/issues#616
Connected to archivematica/issues#594
Connected to archivematica/issues#605
Related to archivematica/issues#493
Related to https://github.com/artefactual/archivematica-storage-service/issues/191


sevein (changes requested):
@ross-spencer can we try to keep changes in `stable/0.14.x` to a minimum? Your refactoring is very welcome in `qa/0.x` though.

If the gist is that you want to run `file` when the pointer file doesn't exist, can't you just do that from the existing `Package._get_decompr_cmd` for the time being?

ross-spencer:
@sevein can you explain more about your rationale? Have you concerns from what I've already put in place? 

I am happy to find a compromise, but I also don't want to add to the complexity of the existing work which is already quite a high. I think if you walk through the commits  you'll see that I've adopted a staged approach to walk people through it. The changes should be fairly minimal if you have a look yourself, i.e. this isn't an attempt to refactor all of the compression/decompression work done in `locations.models.package.py`.

sevein:
I think the problem is that coverage is not as high as we wanted. So there are many unknowns and we want to avoid risk because we can't afford more issues down the road.
I'll give you a recent example: we had https://github.com/artefactual/archivematica/pull/1369 merged, a +24 -46 diff that looked pretty trivial but it broke a couple of things in stable that needed further work: indexing of AIP size and the management command. We had two other devs involved, it blocked other work and deployment we were doing. The gist of that work was between 5-10 lines of code only.

ross-spencer:
I feel like that's unfair @sevein. I don't know if it warranted the example given. The only way to improve our process is to have the opportunity to improve it. We learned a lot from that small issue. There's still room to merge a smaller fix here but I don't think I've fully uncovered the impact of that.

> If the gist is that you want to run file when the pointer file doesn't exist, can't you just do that from the existing Package._get_decompr_cmd for the time being?

This is a trivial looking change but with potentially the bigger impact we can't easily judge, refactoring aside. I think it shows that the idea of what is a trivial or non-trivial is quite subjective.

PR #442, opened 2019-03-25
pipeline: fix remote_name help text

hakamine:
Connected to https://github.com/archivematica/Issues/issues/604

### DIFF
--- a/storage_service/locations/migrations/0022_update_pipeline_help_text.py
+++ b/storage_service/locations/migrations/0022_update_pipeline_help_text.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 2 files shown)


cole (changes requested):
@hakamine unfortunately Django requires migrations for help text changes (even though they are a no-op for most DBs) which is why Travis has failed. You can create a migration with the following command: `python manage.py makemigrations locations --name update_pipeline_help_text` or similar, and then commit the resulting file. See [the migration docs](https://docs.djangoproject.com/en/1.8/topics/migrations/) for more details. If you need more info happy to answer any questions.

hakamine:
@cole Thank you very much for you help. I added a commit with the missing migration. If all looks good I will squash and merge.

cole (approved):
👍 looks good!

PR #443, opened 2019-03-26
Ensure that the storage service can see an AIP to test whether a pointer file needs creating

ross-spencer:
Enable ptr file creation for remote filesystem

When a Storage Service is configured to connect to a pipeline on a
remote server it cannot see the location of the package before it
is transferred over to the storage service. In its current
configuration Archivematica tries to create a pointer file based on
an AIP being a file, or not, before this happens. We implement a
change to query the remote location here and test for isfile here.

Connected to archivematica/issues#594
Connected to archivematica/issues#599

### DIFF
--- a/storage_service/locations/models/pipeline_local.py
+++ b/storage_service/locations/models/pipeline_local.py
@@ -136,3 +136,13 @@ class PipelineLocalFS(models.Model):
 
         # Move file
         return self.space.move_rsync(source_path, destination_path, assume_rsync_daemon=self.assume_rsync_daemon, rsync_password=self.rsync_password)
+
+    def isfile(self, path):
+        """Verify that something is a file in this Space's context."""
+        LOGGER.info("Testing if isfile in pipeline file system: %s", path)
+        basename = os.path.basename(path)
+        if not basename:
+            return False
+        location_entries = self.space.browse(os.path.dirname(path))
+        return (basename in location_entries.get("entries", []) and
+            basename not in location_entries.get("directories", []))
--- a/storage_service/locations/models/space.py
+++ b/storage_service/locations/models/space.py
@@ -448,6 +448,17 @@ class Space(models.Model):
                 {'protocol': self.get_access_protocol_display()}
             )
 
+    def isfile(self, path):
+        """Verify that something is a file in the context of a given space."""
+        child = self.get_child_space()
+        if hasattr(child, 'isfile'):
+            return child.isfile(path)
+        else:
+            raise NotImplementedError(
+                _('Space %(protocol)s does not implement isfile') %
+                {'protocol': self.get_access_protocol_display()}
+            )
+
     # HELPER FUNCTIONS
 
     def move_rsync(self, source, destination, try_mv_local=False, assume_rsync_daemon=False, rsync_password=None):
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -611,7 +611,12 @@ class Package(models.Model):
                 self.current_path)
         if not package_type:
             package_type = self.package_type
-        isfile = os.path.isfile(package_full_path)
+        # The package hasn't been moved yet, test for it being a file on the
+        # originating Space.
+        try:
+            isfile = self.origin_location.space.isfile(package_full_path)
+        except NotImplementedError:
+            isfile = os.path.isfile(package_full_path)
         isaip = package_type in (Package.AIP, Package.AIC)
         ret = isfile and isaip
         if not ret:


storage_service/locations/models/pipeline_local.py
```diff
@@ -136,3 +136,17 @@ def move_from_storage_service(self, source_path, destination_path, package=None)
 
         # Move file
         return self.space.move_rsync(source_path, destination_path, assume_rsync_daemon=self.assume_rsync_daemon, rsync_password=self.rsync_password)
+
+    def isfile(self, path_):
```
jraddaoui: Is `path` a reserved keyword? I think you could remove the trailing slash, at least it's being used like that in other places of this module.

storage_service/locations/models/pipeline_local.py
```diff
@@ -136,3 +136,17 @@ def move_from_storage_service(self, source_path, destination_path, package=None)
...
+        """Verify that something is a file in this Space's context."""
+        LOGGER.info("Testing if isfile in pipeline file system: %s", path_)
+        basename = os.path.basename(path_)
+        if not len(basename):
```
jraddaoui: You could evaluate the string directly, without using `len()`.

storage_service/locations/models/pipeline_local.py
```diff
@@ -136,3 +136,17 @@ def move_from_storage_service(self, source_path, destination_path, package=None)
...
+            return False
+        path_ = os.path.join(os.path.dirname(path_), "")
```
jraddaoui: Is this really necessary? Could `os.path.dirname(path_)` be enough?

storage_service/locations/models/pipeline_local.py
```diff
@@ -136,3 +136,17 @@ def move_from_storage_service(self, source_path, destination_path, package=None)
...
+        ssh_path = self._format_host_path(path_)
+        location_entries = self.space.browse_rsync(ssh_path, assume_rsync_daemon=self.assume_rsync_daemon, rsync_password=self.rsync_password)
+        if basename in location_entries.get("entries", []) \
+                and basename not in location_entries.get("directories", []):
```
jraddaoui: Doesn't have a `files` key? Just making sure, it could be a great addition in the future. I also prefer the parenthesis syntax for multi-line ifs, I guess black will help us with that ;)

storage_service/locations/models/space.py
```diff
@@ -448,6 +448,17 @@ def check_package_fixity(self, package):
                 {'protocol': self.get_access_protocol_display()}
             )
 
+    def isfile(self, path_):
```
jraddaoui: Same in here, I think you could use `path`.

storage_service/locations/models/pipeline_local.py
```diff
@@ -136,3 +136,17 @@ def move_from_storage_service(self, source_path, destination_path, package=None)
...
+
+    def isfile(self, path_):
+        """Verify that something is a file in this Space's context."""
+        LOGGER.info("Testing if isfile in pipeline file system: %s", path_)
+        basename = os.path.basename(path_)
+        if not len(basename):
+            return False
+        path_ = os.path.join(os.path.dirname(path_), "")
+        ssh_path = self._format_host_path(path_)
+        location_entries = self.space.browse_rsync(ssh_path, assume_rsync_daemon=self.assume_rsync_daemon, rsync_password=self.rsync_password)
```
jraddaoui: Line length should be easy to fix in here.

jraddaoui: Actually, the last three lines in here should do the same as calling `self.browse(path_)` directly.

storage_service/locations/models/pipeline_local.py
jraddaoui: Also, you could return the condition directly and remove the two returns bellow.

storage_service/locations/models/pipeline_local.py
jraddaoui: It's also used in the `browse` method like that, so I'd use that method and live it be.

jraddaoui (changes requested):
Thanks @ross-spencer! I've added a few comments, mostly cosmetic changes.

storage_service/locations/models/pipeline_local.py
ross-spencer: :+1:

storage_service/locations/models/pipeline_local.py
ross-spencer: Good spot - thank you for that. I've made that change which cleans a lot up quite nicely.

storage_service/locations/models/pipeline_local.py
ross-spencer: Unfortunately no, there is a `properties` so one could check if it has 'size'. That being said, I don't know if you read this the same as me, but we're not actually attaching the `properties dict` to anything so it's not returned for this function??? (My eyes might just be tired)

https://github.com/artefactual/archivematica-storage-service/blob/d83d383f44dd258f6ba4c1b5a879ca22270e2ae6/storage_service/locations/models/space.py#L653-L668

ross-spencer:
Thanks @jraddaoui that was great about `browse`. `path_` you are right about the module convention so I have matched that. I like to use `path_` a lot in my own work, but not `path` because folks can do `from os import path`.

storage_service/locations/models/pipeline_local.py
jraddaoui: What I mean is that you could use directly:

```py
return basename in location_entries.get("entries", []) and basename not in location_entries.get("directories", [])
```

And remove the two `return` bellow, as they actually return the condition value.

ross-spencer: Yup :+1: that's in my changes :nerd_face:

jraddaoui (approved):
Awesome, thanks @ross-spencer!!

storage_service/locations/models/pipeline_local.py
jraddaoui: Right, I misunderstood the comment. Yes, not for now, but checking and returning an extra key for files in `browse_rsync` could be a great addition in the future.

PR #448, opened 2019-04-09
S3: Enable the creation of storage spaces in different AWS/S3 regions

mamedin:
Connects to https://github.com/archivematica/Issues/issues/638 and https://github.com/archivematica/Issues/issues/639

**EDIT:** *(from Ross)*

This had been rebased and ensures that Boto3 is upgraded. There were complications testing using the [`head_bucket`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.head_bucket) client function with Boto3, namely, the 404 we anticipated wasn't being returned, at least in my environment. And my testing of nonsense bucket names proved inconsistent as well. I swapped `head_bucket` out for [`get_bucket_location`](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/s3.html#S3.Client.get_bucket_location) which is equally usable, and can only return information if the bucket exists. If the bucket doesn't exist we can create it per the original logic, and return an exception if there are any other difficulties. 

I have added notes on this in the docstring. 

**Notes on testing**

Those looking at this might be tempted to use the same space delete the bucket and then change the region settings to show S3 switching between locations. I would caution against this without leaving adequate time between bucket deletions as the S3 service may still be processing that.

The easiest way to test that we can create locations in different spaces is to create two different spaces.

Two locations I used for testing are: 

* ca-central-1
* us-east-2

Good work @mamedin and thanks to whomever may CR this!

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -11,8 +11,8 @@ attrs==19.1.0             # via pytest
 babel==2.6.0
 backports.shutil-get-terminal-size==1.0.0  # via ipython
 bagit==1.7.0
-boto3==1.6.5
-botocore==1.9.23
+boto3==1.9.174
+botocore==1.12.175
 brotli==0.5.2
 certifi==2019.3.9
 cffi==1.12.3
@@ -97,7 +97,7 @@ pyyaml==5.1
 requests-oauthlib==1.2.0
 requests==2.21.0
 rfc3986==1.3.2
-s3transfer==0.1.13
+s3transfer==0.2.1
 scandir==1.10.0           # via pathlib2
 simplegeneric==0.8.1      # via ipython
 six==1.12.0
--- a/storage_service/locations/models/s3.py
+++ b/storage_service/locations/models/s3.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from __future__ import absolute_import
 
 # stdlib, alphabetical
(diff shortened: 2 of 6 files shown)


storage_service/locations/models/s3.py
```diff
@@ -76,7 +76,13 @@ def resource(self):
         return self._resource
 
     def _ensure_bucket_exists(self):
-        self.client.create_bucket(Bucket=self._bucket_name())
+        try:
+            self.client.head_bucket(Bucket=self._bucket_name())
+        except:
```
ross-spencer:
```suggestion
        except self.client.exceptions.ClientError:
```
Nice! I've offered a small suggestion for the except. While there's little that can go wrong here, we should try and capture specific exceptions just in case, so we don't mask what those might be. 

I think this looks like the right syntax from these docs here but you might need to try it out: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/migrations3.html#accessing-a-bucket.

ross-spencer (changes requested):
One small change requested! But otherwise nice addition.

storage_service/locations/models/s3.py
mamedin: Thanks @ross-spencer :)

mamedin:
I included the @ross-spencer suggestion.

ross-spencer:
Thanks @mamedin - there's one last thing - I'm not sure what the expected behavior of the Black linter is but it wants to change the line:  `CreateBucketConfiguration={ 'LocationConstraint': self.region }` to `CreateBucketConfiguration={"LocationConstraint": self.region},` which is a valid and good change, but can't do it via CI and so fails. 

Can you run `Black` across this file and then rebase the commits, and force push, and then we can approve these changes to be merged? 

Travis is a bit of a new art to me, I wonder if there's something we need to do to make sure black has permissions to do its changes as required. It seems that it should fail but IDK.

ross-spencer:
This is a slightly related PR from Wellcome where Boto3 is upgraded: https://github.com/wellcometrust/archivematica-storage-service/pull/4

storage_service/locations/models/s3.py
```diff
@@ -76,7 +76,13 @@ def resource(self):
         return self._resource
 
     def _ensure_bucket_exists(self):
-        self.client.create_bucket(Bucket=self._bucket_name())
+        try:
+            self.client.head_bucket(Bucket=self._bucket_name())
+        except self.client.exceptions.ClientError:
+            self.client.create_bucket(
```
cole, line 103: I think this is still missing the check that the error was a 404:

```
            # If a client error is thrown, then check that it was a 404 error.
            # If it was a 404 error, then the bucket does not exist.
            error_code = e.response["Error"]["Code"]
            if error_code == "404":
                 ...
```

ross-spencer: :+1:

requirements/test.txt
```diff
@@ -113,3 +112,6 @@ virtualenv==16.6.0        # via tox
 wcwidth==0.1.7            # via prompt-toolkit
 whitenoise==3.3.0
 wrapt==1.11.1
+
+# The following packages are considered to be unsafe in a requirements file:
```
replaceafill: I think this has been discussed already here https://github.com/artefactual/archivematica-storage-service/pull/473#discussion_r297287410

replaceafill (approved):
Nice work @mamedin and @ross-spencer. I tested this with two spaces as per the **Notes on testing** (thanks a lot for those btw!) and the buckets were created in the right regions, so it's :+1: for me.

ross-spencer (approved):

PR #449, opened 2019-04-09
Extend service callbacks

jraddaoui:
Allows to create request callbacks for AIP, AIC and DIP stored events
alongside the existing "post_store" event used for AIP source files.

- Add new event options for callbacks.
- Add "body" and "headers" fields for better callback customization.
- Manage headers key/value fields dynamically.
- Execute new post store AIP, AIC and DIP callbacks after the package
  is moved to uploaded.
- Replace `<package_uuid>` by the package UUID in the callback URI
  and body.
- Check enabled field on callback execution.
- Display event labels.
- Improve callback form description.

https://github.com/archivematica/Issues/issues/147

### DIFF
--- a/storage_service/locations/forms.py
+++ b/storage_service/locations/forms.py
@@ -1,4 +1,6 @@
 from __future__ import absolute_import
+from collections import OrderedDict
+import json
 import logging
 
 from django import forms
@@ -392,10 +394,120 @@ class ConfirmEventForm(forms.ModelForm):
         self.fields["status_reason"].required = True
 
 
+class HeaderWidget(forms.MultiWidget):
+    """Key/value text inputs widget for Callback header."""
+
+    def __init__(self, *args, **kwargs):
+        widgets = (forms.TextInput(), forms.TextInput())
+        super(HeaderWidget, self).__init__(widgets, *args, **kwargs)
+
+    def decompress(self, value):
+        return value or [None, None]
+
+
+class HeaderField(forms.fields.MultiValueField):
+    """Key/value form field for Callback header."""
+
+    widget = HeaderWidget
+
+    def __init__(self, *args, **kwargs):
+        fields = (forms.fields.CharField(), forms.fields.CharField())
+        super(HeaderField, self).__init__(fields, *args, **kwargs)
+
+    def compress(self, data_list):
+        return data_list or [None, None]
+
+
 class CallbackForm(forms.ModelForm):
     class Meta:
         model = models.Callback
-        fields = ("uri", "event", "method", "expected_status")
+        fields = ("event", "uri", "method", "body", "expected_status", "enabled")
+
+    def __init__(self, *args, **kwargs):
+        """Extends model form initialization.
+
+        Adds extra fields for Callback headers and reorders form fields.
+        """
+        super(CallbackForm, self).__init__(*args, **kwargs)
+        # Save field order to include the headers after the method field
+        field_order = ["event", "uri", "method"]
+        # Add fields for existing headers
+        counter = 0
+        if self.instance:
+            for header in self.instance.get_headers().items():
+                field_name = "header_%s" % counter
+                self.fields[field_name] = HeaderField(
+                    initial=header, label="", required=False
+                )
+                field_order.append(field_name)
+                counter += 1
+        # And add an extra empty header field
+        field_name = "header_%s" % counter
+        self.fields[field_name] = HeaderField(label="", required=False)
+        field_order.append(field_name)
+        # Add label only to first header field
+        self.fields["header_0"].label = _("Headers (key/value)")
+        # Add remaining fields to field order and reorder
+        field_order.extend(["body", "expected_status", "enabled"])
+        self.order_fields(field_order)
+
+    def order_fields(self, field_order):
+        """Rearranges the fields according to field_order.
+
+        Copied from Django 1.9, BaseForm class.
+        """
+        if field_order is None:
+            return
+        fields = OrderedDict()
+        for key in field_order:
+            try:
+                fields[key] = self.fields.pop(key)
+            except KeyError:  # ignore unknown fields
+                pass
+        fields.update(self.fields)  # add remaining fields in original order
+        self.fields = fields
+
+    def clean(self):
+        """Formats headers inputs to save them in the Callback instance.
+
+        Checks the key and value for each header multi-field to create an OrderedDict
+        that is dumped as a string to be added to the intance's `headers` text field.
+        """
+        cleaned_data = super(CallbackForm, self).clean()
+        headers = OrderedDict()
+        counter = 0
+        field_name = "header_%s" % counter
+        while cleaned_data.get(field_name):
+            key, value = cleaned_data[field_name]
+            if key and value:
+                headers[key] = value
+            cleaned_data.pop(field_name)
+            counter += 1
+            field_name = "header_%s" % counter
+        # Fields added dinamically via JS are not included in the cleaned data.
+        # This fields are multi-fields for the headers key/vale pairs and, since
+        # they are not combined in the cleaning process, we need to look for both
+        # fields in the data QueryDict following the counter.
+        key = self.data.get("header_%s_0" % counter)
+        value = self.data.get("header_%s_1" % counter)
+        # Continue if the key and value exist in data
+        while key is not None and value is not None:
+            # But only add them to the headers if populated
+            if key and value:
+                headers[key] = value
+            counter += 1
+            key = self.data.get("header_%s_0" % counter)
+            value = self.data.get("header_%s_1" % counter)
+        cleaned_data["headers"] = json.dumps(headers) if headers else ""
+        return cleaned_data
+
+    def save(self, commit=True):
+        """Adds the formatted headers to the instance before save."""
+        callback = super(CallbackForm, self).save(commit=False)
+        callback.headers = self.cleaned_data["headers"]
+        if commit:
+            callback.save()
+        return callback
 
 
 class ReingestForm(forms.Form):
--- a/storage_service/locations/models/event.py
+++ b/storage_service/locations/models/event.py
@@ -1,4 +1,6 @@
 # stdlib, alphabetical
+from collections import OrderedDict
+import json
 
 # Core Django, alphabetical
 from django.conf import settings
@@ -70,7 +72,12 @@ class Callback(models.Model):
     particular HTTP method.
     """
 
-    EVENTS = (("post_store", "Post-store"),)
+    EVENTS = (
+        ("post_store", _("Post-store AIP (source files)")),
+        ("post_store_aip", _("Post-store AIP")),
+        ("post_store_aic", _("Post-store AIC")),
+        ("post_store_dip", _("Post-store DIP")),
+    )
 
     HTTP_METHODS = (
         ("delete", "DELETE"),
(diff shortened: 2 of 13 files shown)


jraddaoui:
Hi @sevein, @cole,

Apart from adding tests and checking the AIC question in the TODOs (I won't have the time to limit the callbacks to a set of locations for now), this is ready for an initial review. If I understood correctly, you're planing to work on something similar for another project, so I'd appreciate your feedback about this changes before moving forward.

Thasnks!

storage_service/locations/forms.py
```diff
@@ -392,10 +394,120 @@ def __init__(self, *args, **kwargs):
...
+
+        Adds extra fields for Callback headers and reorders form fields.
+        """
+        super(CallbackForm, self).__init__(*args, **kwargs)
+        # Save field order to include the headers after the method field
+        field_order = ["event", "uri", "method"]
+        # Add fields for existing headers
+        counter = 0
+        if self.instance:
+            for header in self.instance.get_headers().items():
```
cole, line 437: It's been a while since I've used one, but [formsets](https://docs.djangoproject.com/en/1.8/topics/forms/formsets/) exist for this use case — although it may not be that easy because headers isn't a model. You should be able to use a separate HeaderForm and formset for that, though.

Formsets require some JS handling (not really more than you already have in place though), it looks like it hasn't been updated in a while but https://github.com/elo80ka/django-dynamic-formset may make that easier.

storage_service/locations/models/event.py
```diff
@@ -117,12 +137,28 @@ class Meta:
         verbose_name = _("Callback")
         app_label = "locations"
 
-    def execute(self, url=None):
+    def get_event_label(self):
```
cole: They builtin `get_event_display` should cover this case. See: https://docs.djangoproject.com/en/1.8/ref/models/instances/#django.db.models.Model.get_FOO_display

storage_service/locations/models/event.py
```diff
@@ -100,6 +106,20 @@ class Callback(models.Model):
         verbose_name=_("Method"),
         help_text=_("HTTP request method to use in connecting to the URL."),
     )
+    body = models.TextField(
```
cole, line 110: I can see a case both ways, but I'm not sure yet if we actually need to be configurable. It may be more useful to have a standard webhook payload that we document.

cole (changes requested):
I think these changes are definitely an improvement on what is in place right now, so that's great. 

I've been looking how we'll use these (if at all) for NHA though, and I think it might be beneficial to use a more standard, less configurable approach to webhooks, which pushes a more work to the consumer (the app receiving the callback).

This article has a good overview of the approaches taken by many services:
https://restful.io/webhooks-dos-and-dont-s-what-we-learned-after-integrating-100-apis-d567405a3671

For example there's a lot of work in this PR to be able to set arbitrary headers, I assume for authentication. That isn't a bad thing to have, but is kind of a limited approach to webhook security, which is a bit different from typical API endpoint security (you can't use CSRF protection on endpoints that handle webhooks, for example). Most webhook systems take the approach of either:

a) signing the payload with a shared secret, a la GitHub: https://developer.github.com/webhooks/securing/
b) providing a "skinny payload" — a notification only that something has changed, which results in an API call to fetch any required data, meaning your endpoint can be open without much risk.

I think in all cases right now we're already doing B, correct? The only data we are sending is a UUID; an impersonator sending a fake ID will just force us to make an API request that returns a 404.

Likewise for a configurable payload body — that might be a case of YAGNI, since the current setup only actually substitutes one variable, so anything new will likely require code changes anyways.

I guess in summary, I would say for now just don't bother with the header and body fields, and open up the endpoint on the CCA side.

cole:
@jraddaoui One more quick thing, I realize it's kind of late in the process here to bring up those kind of concerns — sorry I didn't notice earlier!

storage_service/locations/forms.py
jraddaoui: I looked at formsets and examples using django-dynamic-fomset but I thought it would be easier to treat the header with a dynamic multi-input. In the end I wasn't so sure about that, but it was kind of late to go back :(

storage_service/locations/models/event.py
jraddaoui: Oh, that's nice!

storage_service/locations/models/event.py
jraddaoui: The idea of making it configurable was to facilitate SS upgrades and consumer integrations. For example, in SCOPE I'm checking if a download URL is present in the payload and falling back to build a SS API v2 URL with the DIP uuid and the Origin URL.

jraddaoui:
Thanks @cole!

In the integration we're doing in SCOPE, multiple SS instances could connect to it to send notifications. One of the reasons to make the headers configurable is to be able to send an "Origin" header with the SS URL, which will be used to check the credentials in the SCOPE configuration, build the download URL and fetch the DIP. As you mention, there are many options to solve this issue, but I thought it will be nice to do it this way and have an editable payload and headers in the callbacks, up to the user to use them or not.

cole:
@jraddaoui What about always including an `Origin` header in the request? It's useful information for everyone. Same for download URL in the payload?  I think my main concern here is by putting everything in the db, if we want to change things or add headers in the future it will require data migrations.

On that note though, if we're going to include a download URL in the payload, then the receiving (SCOPE side) endpoint can't be open as I suggested (or if we're going to build the URL from the Origin header and download, as that's easy to spoof as well).

So either we can sign the request, or use the `Authorization` header. I do think maybe there's an opportunity to have this work as you've planned and make this change a lot simpler by only storing the `Authorization` header in the db, and setting `Origin` automatically (ideally from the  [sites framework](https://docs.djangoproject.com/en/1.8/ref/contrib/sites/), but I haven't checked if that is being used in SS).

jraddaoui:
> @jraddaoui What about always including an Origin header in the request? It's useful information for everyone. Same for download URL in the payload?

Yes, that's what I was planing to do, still need to find how to get the Origin URL automatically (the allowed origins can have multiple URLs and I think we're not using the sites framework). So it may need a new setting or something like that ...

> I think my main concern here is by putting everything in the db, if we want to change things or add headers in the future it will require data migrations.

They're custom headers for that reason, if someone upgrades the SS or the customer, they can always go back and edit those headers in the callback form. In the end, after all the transformations and dynamic manipulation in the form, the headers end up in a single field as a JSON string.

> On that note though, if we're going to include a download URL in the payload, then the receiving (SCOPE side) endpoint can't be open as I suggested (or if we're going to build the URL from the Origin header and download, as that's easy to spoof as well).

That's a great point.

> So either we can sign the request, or use the Authorization header. I do think maybe there's an opportunity to have this work as you've planned and make this change a lot simpler by only storing the Authorization header in the db, and setting Origin automatically (ideally from the sites framework, but I haven't checked if that is being used in SS).

I think that combining custom and hard-coded headers will complicate things a bit more.

jraddaoui:
> > On that note though, if we're going to include a download URL in the payload, then the receiving (SCOPE side) endpoint can't be open as I suggested (or if we're going to build the URL from the Origin header and download, as that's easy to spoof as well).

>That's a great point.

Actually, with the Origin validation that we're doing in SCOPE to check the SS credentials I think we could be covered if someone is trying to force us to download bad stuff, but having auth. in there is a great extra layer.

cole:
> Yes, that's what I was planing to do, still need to find how to get the Origin URL automatically (the allowed origins can have multiple URLs and I think we're not using the sites framework).

Hmm, that's tricky, since we're not in a request context.

> They're custom headers for that reason, if someone upgrades the SS or the customer, they can always go back and edit those headers in the callback form. In the end, after all the transformations and dynamic manipulation in the form, the headers end up in a single field as a JSON string.

I see that they are stored as JSON, but that doesn't really make a difference as far as I can see. I'm still not convinced that this flexibility is a win, but since you've already written the code, that's a strong point in favour of configurability. 😁 It also may turn out to be helpful, it's pretty hard to predict. I'm ok to go with this approach for now.

> Actually, with the Origin validation that we're doing in SCOPE to check the SS credentials we could be covered if someone is trying to force us to download bad stuff, but I think that having auth. in there is a great extra layer.

I haven't seen how this is setup but I don't think you can meaningfully validate an `Origin` header (outside of a browser context). Happy to hear if you have a way to do it!

storage_service/locations/models/event.py
cole: I think the main problem here is if you are a third party writing a consumer...how do you know what to expect in the payload? But maybe we'll end up needing the configurability. NHA might even.

storage_service/locations/forms.py
cole: Ok, sounds good.

jraddaoui:
> Hmm, that's tricky, since we're not in a request context.

Not sure if you saw the update in the comment, I added that we could use a setting (defined through the env.) or something like that to configure that "Origin URL" if we need to do so.

> I see that they are stored as JSON, but that doesn't really make a difference as far as I can see. I'm still not convinced that this flexibility is a win, but since you've already written the code, that's a strong point in favour of configurability. grin It also may turn out to be helpful, it's pretty hard to predict. I'm ok to go with this approach for now.

I just don't understand your point about the migrations, but if we think this will be more of a problem than an enhancement in the future I'm happy to remove that code earlier on.

> I haven't seen how this is setup but I don't think you can meaningfully validate an Origin header (outside of a browser context). Happy to hear if you have a way to do it!

To be fair, I could have used a different header, I just thought it makes sense to use that one even if we're not in a browser context. Please, let me know if you think it's better to use a different one to avoid confusion. This is [the work in progress PR](https://github.com/CCA-Public/dip-access-interface/pull/153) in SCOPE if you want to take an early look at it.

cole:
> I just don't understand your point about the migrations, but if we think this will be more of a problem than an enhancement in the future I'm happy to remove that code earlier on.

I think I haven't done a good job of making this clear, but what I've been trying to advocate for is a clearly defined callback from SS to other systems — meaning it has a payload that is consistent and headers that are always present.

The approach this PR takes is much more flexible, which can handle many existing system types, but if you are trying to define the webhook call behaviour (for example, in docs that people integrating with SS would use), the answer is, "it depends on how you set it up".

Both approaches have pros and cons, but on reflection, I think we should go with the approach you've taken here (meaning changes to the PR not necessary), because we don't yet know if this will actually be a problem. It might turn out that we need the flexibility.

Sorry for confusing things!

> To be fair, I could have used a different header, I just thought it makes sense to use that one even if we're not in a browser context. Please, let me know if you think it's better to use a different one to avoid confusion. This is the work in progress PR in SCOPE if you want to take an early look at it.

I think `Origin` is the best header to use, the point I'm trying to make is any header we choose can't be trusted, since this is a server to server request (in a browser context, `Origin` is set by the browser, it which case it is pretty trustworthy) — if I make a curl request to the SCOPE endpoint saying my origin is google.com, how do you know that I'm lying?

cole (approved):

jraddaoui:
Thanks @cole!

No worries, I should also have been clearer about choosing that flexibility to solve the authorization and origin issues I was facing.

The current [documentation](https://www.archivematica.org/en/docs/storage-service-0.14/administrators/#service-callbacks) only includes an integration example for Islandora, but I'm not sure if we should add another example for SCOPE when everything is ready, I think it may be better to just document it in their docs. For new integrations it'll be as you said, which may be good thing for a third party developing an integration, as long as they document properly how everything should be configured.

About the origin header, you're right that it's less trustworthy without coming for a browser, but that shouldn't be a problem with the current implementation. If an unknown origin is used it will fail looking for that origin in the SCOPE configuration and, if someone tries to impersonate a configured origin, they will also need the authorization header.

I'll add some minor tests and ask you for a final review.

cole (approved):
New changes look good.

jraddaoui:
Thanks again @cole! I've created https://github.com/archivematica/Issues/issues/673 for the remaining to-do.

PR #451, opened 2019-05-02
Async: convert buffer to str before pickling

sevein:
Convert `Async.error` to `str` before unserialization.

This was preventing errors from being reported properly.

Connects to https://github.com/archivematica/Issues/issues/671.
Similar to https://github.com/artefactual/archivematica-storage-service/issues/358.


replaceafill (approved):
@sevein :+1: Thanks for the reference to #358 too. I didn't know about memoryview objects :O

PR #436, opened 2019-01-18
Add configurable variable 'as_archival_repo' to DSpace REST location

J4bbi:
allows defining destination ArchivesSpace repository in transfer metadata using the key
'archivesspace_dip_repository'

Connects to https://github.com/archivematica/Issues/issues/458.

PR #452, opened 2019-05-14
DSpace REST: make AS repo configurable

sevein:
Update `_assign_destination` so it learns how to determine the AS repo
by looking up the value in `archivesspace_dip_repository`. This is in
addition to other values already supported such `dspace_dip_collection` or
`dspace_aip_collection`.

`_analyze_md_els` stops populating `None` values when the lookups in the
metadata file do not return matches - this ensures that the config defaults
are used under this scenario where before they were ignored.

Connects to https://github.com/archivematica/Issues/issues/435.
Connects to https://github.com/archivematica/Issues/issues/458.
This replaces https://github.com/artefactual/archivematica-storage-service/pull/436. Coded by @J4bbi.

### DIFF
--- a/storage_service/locations/models/dspace_rest.py
+++ b/storage_service/locations/models/dspace_rest.py
@@ -174,7 +174,6 @@ class DSpaceREST(models.Model):
         """Locate, extract (if necessary), XML-parse and return the METS file
         for this package.
         """
-        mets_path = ""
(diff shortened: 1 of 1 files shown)


cole (approved):
This looks good. I don't have a testing setup, but I don't see any issues with the changes.

Issue #270, opened 2017-11-22
Problem: double replications being created from the first replication, not the original


sromkey:
To reproduce:
- create an AIP storage location which has 2 replicating locations
- Store an AIP in the location

Result:
The second replication will be created from the first replication, rather than from the original AIP. If you look at the pointer files for the original and replicated AIPs:
- original AIP has one replication event resulting in the first replication
- first replication also has a replication event resulting in the second replication

Instead, the original AIP pointer file should have two replication events and two corresponding validation events.

joel-simpson:
@sromkey I tagged this with the triage-release-1.10 label because it was fixed for a client project (NHA) and PR has been completed, so better to have this merged sooner rather than later.

cole:
> Instead, the original AIP pointer file should have two replication events and two corresponding validation events.

After discussion, this isn't quite true for validation events, as the validation events are in the pointer files for the replica AIPs. The behaviour we want (I think) is:
1. The original pointer file should contain one replication event per replica
2. Each replica should contain one validation event
3. Replica pointer files don't contain replication events, creation is used instead.

joel-simpson:
The PR for this is merged so this looks like it can be moved over to 'review'... changing status now.

sallain:
What a weird issue.

I think this is accurate now, though I just want to confirm the acceptance criteria, @sromkey.

My storage setup has a standard AIP storage location, plus two replicator locations which are both connected to the standard location

![Screenshot of Archivematica Storage Service.jpg](https://codetreeusercontent.s3.amazonaws.com/uploads/93454c3c-1cd6-4f41-9b3d-0fa8d8897fe7/Screenshot+of+Archivematica+Storage+Service.jpg)

When I store an AIP in the standard location, two replications are created.

The pointer files contain the following info:

* The pointer file for the original has two replication PREMIS events (one for each replication)
* The pointer file for the original does not have any validation events
* The pointer files for the replications each have a validation event that says they have been validate against the original

That all seems correct to me, but just want a second set of eyes on it!

PR #453, opened 2019-05-16
Fix replication

cole:
Relates to: https://github.com/archivematica/Issues/issues/684
Relates to: https://github.com/archivematica/Issues/issues/463

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -1192,7 +1192,7 @@ class Package(models.Model):
         replication_event = self.create_replication_event(
             replica_package, event_uuid=replication_event_uuid, agents=ss_agents
         )
-        old_premis_events += (replication_event,)
+        old_premis_events.append(replication_event)
         replication_relationship = _get_replication_derivation_relationship(
             replica_package.uuid, replication_event_uuid
         )
@@ -1206,7 +1206,7 @@ class Package(models.Model):
         )
         for ss_agent in ss_agents:
             if ss_agent not in old_premis_agents:
-                old_premis_agents += (ss_agent,)
+                old_premis_agents.append(ss_agent)
         return self.create_pointer_file(
             new_premis_object,
             old_premis_events,
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -138,10 +138,6 @@ class Package(models.Model):
         help_text=_("For storing flexible, often Space-specific, attributes"),
     )
 
-    # Temporary attributes to track path on locally accessible filesystem
-    local_path = None
-    local_path_location = None
-
     PACKAGE_TYPE_CAN_DELETE = (AIP, AIC, TRANSFER)
     PACKAGE_TYPE_CAN_DELETE_DIRECTLY = (DIP,)
     PACKAGE_TYPE_CAN_EXTRACT = (AIP, AIC)
@@ -159,6 +155,14 @@ class Package(models.Model):
     class Meta:
         verbose_name = _("Package")
         app_label = "locations"
+
+    def __init__(self, *args, **kwargs):
+        super(Package, self).__init__(*args, **kwargs)
+
+        # Temporary attributes to track path on locally accessible filesystem
+        self.local_path = None
+        self.local_path_location = None
+        self.origin_location = None
 
     def __unicode__(self):
         return u"{uuid}: {path}".format(uuid=self.uuid, path=self.full_path)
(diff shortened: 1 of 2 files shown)


replaceafill:
@cole Do you think your fix also fixes https://github.com/archivematica/Issues/issues/463?

cole:
@replaceafill it seems like it should, I didn't test that case though.

storage_service/locations/models/package.py
```diff
@@ -1183,12 +1192,12 @@ def create_new_pointer_file_with_replication(
         replication_event = self.create_replication_event(
             replica_package, event_uuid=replication_event_uuid, agents=ss_agents
         )
-        old_premis_events.append(replication_event)
+        old_premis_events += (replication_event,)
```
ross-spencer: I suspect I'm missing something obvious but I'm not sure why this change is needed for this? `old_premis_events` I believe to be a list at this stage so adding a one item tuple, or appending to the list has the same effect? Same q. with the other `+= (tuple, )` lines below.

storage_service/locations/tests/test_package.py
```diff
@@ -357,6 +357,27 @@ def test_run_post_store_callbacks_dip(self):
...
+            relative_path=replication_dir,
+            purpose=models.Location.REPLICATOR,
+        )
+
+        aip.create_replicas()
+        replica = aip.replicated_package
+
+        assert replica is not None
+        # The relationship is a little bit broken here; one would think that the
+        # _original_ AIP should have replicas, and the replicated AIP should not.
```
ross-spencer, line 377: I agree! The database is a little cleaner in this regard, and just has a `replicated_package_id` so I guess we just work it out from that.

ross-spencer (approved):
Just a clarification asked for my own learning, but otherwise looks great. Thanks for the quick fix! 

One point, and this is coming from someone who I think might not be great at this, maybe a note for a potential process in the future, but a diagnosis of the issue, in the PR or the issue itself would be great. I feel like it would speed up CR but also help when it comes to discuss in joint analyst/dev meetings too. 

I think I can see here that I introduced an assumption with https://github.com/artefactual/archivematica-storage-service/commit/3f21a96879103d6e550752770e344cff42cd919a that self.origin_location was initialized, but that isn't the case when creating a replicated package, and so we need to be more careful with that before determining whether a pointer file is created/updated.

ross-spencer:
My test shows archivematica/Issues#463 is fixed as well. A point of testing for the analysts, (based on the chat in Slack between yourself and Evelyn) should be to:

* [ ] create a replicated packaged.
* [ ] ensure the replication entries in the pointer file are correct.

cole:
Thanks for the feedback @ross-spencer! I may have been premature in marking this as ready for CR, as it hasn't yet been tagged with the 1.10 milestone.

storage_service/locations/models/package.py
cole: This is a good point — I was getting a "tuple has no attribute append" here, but looking at the metsrw code again, I don't see it returning a tuple. I'll have another look.

storage_service/locations/tests/test_package.py
cole: It looks like `replicated_package_id` is also backwards (I think unintentionally, based on the [docstring of replicate](https://github.com/artefactual/archivematica-storage-service/blob/stable/0.14.x/storage_service/locations/models/package.py#L468)), that is, the original package will have a value for `replicated_package_id`, and the replica package will not. This seems to me to be the cause of https://github.com/artefactual/archivematica-storage-service/issues/270. It needs to be looked at more though.

cole:
> I think I can see here that I introduced an assumption with 3f21a96 that self.origin_location was initialized, but that isn't the case when creating a replicated package, and so we need to be more careful with that before determining whether a pointer file is created/updated.

Yeah, I think so, but I think it was broken prior to that. As soon as I fixed the origin_location issue, I started getting the error detailed in https://github.com/archivematica/Issues/issues/463; I'm not clear if that was caused by a metsrw update such as f70a99c or something else. I don't see any smoking guns in metsrw or how it is used in storage service though.

I guess it's good to be careful, but to me the takeaway is that we need more meaningful validation of changes (i.e. unit tests). It's easy to miss things when developing! There are a lot of possible configurations, to the point where I don't think everyone can remember to account for all of them (especially new-ish contributors who like me aren't even aware of all features).

ross-spencer:
> I guess it's good to be careful, but to me the takeaway is that we need more meaningful validation of changes (i.e. unit tests). It's easy to miss things when developing! There are a lot of possible configurations, to the point where I don't think everyone can remember to account for all of them (especially new-ish contributors who like me aren't even aware of all features).

Totally!

storage_service/locations/models/package.py
cole: I think I got the lines mixed up here: `new_relationships` is a tuple, but `old_premis_events` and `old_premis_agents` are lists. `+=` also works for lists, but I'll switch back to `append`.

cole:
@replaceafill, can you review the unit test I added to cover https://github.com/archivematica/Issues/issues/463?

replaceafill (commented):
LGTM! :+1:

PR #454, opened 2019-05-21
Compression methods

helenst:
Improvements to support for compression methods in AIP pointer file processing and reingest

* Add support for gzip compressed AIPS https://github.com/archivematica/Issues/issues/681
* Add support for 7z uncompressed AIPS https://github.com/archivematica/Issues/issues/605
* Refactoring and tests around compression work

Connected to https://github.com/archivematica/Issues/issues/681
Connected to https://github.com/archivematica/Issues/issues/605
Requires: https://github.com/artefactual/archivematica/pull/1412

### DIFF
--- a/storage_service/common/tests/test_utils.py
+++ b/storage_service/common/tests/test_utils.py
@@ -1,0 +1,212 @@
+from StringIO import StringIO
+
+import mock
+import pytest
+
+from common import utils
+
+
+@pytest.mark.parametrize(
+    "pronom,algorithm,compression",
+    [
+        (utils.PRONOM_7Z, "bzip2", utils.COMPRESSION_7Z_BZIP),
+        (utils.PRONOM_7Z, "lzma", utils.COMPRESSION_7Z_LZMA),
+        (utils.PRONOM_7Z, "copy", utils.COMPRESSION_7Z_COPY),
+        (utils.PRONOM_7Z, "unknown algo", utils.COMPRESSION_7Z_BZIP),
+        (utils.PRONOM_BZIP2, "", utils.COMPRESSION_TAR_BZIP2),
+        (utils.PRONOM_GZIP, "", utils.COMPRESSION_TAR_GZIP),
+        ("unknown pronom", "", utils.COMPRESSION_7Z_BZIP),
+    ],
+)
+def test_get_compression(pronom, algorithm, compression):
+    xml = (
+        '<?xml version="1.0"?>'
+        '<mets:mets xmlns:mets="http://www.loc.gov/METS/" xmlns:premis="info:lc/xmlns/premis-v2">'
+        " <premis:formatRegistryKey>%s</premis:formatRegistryKey>"
+        ' <mets:transformFile TRANSFORMALGORITHM="%s"></mets:transformFile>'
+        "</mets:mets>"
+    ) % (pronom, algorithm)
+
+    assert (
+        utils.get_compression(StringIO(xml)) == compression
+    ), "Incorrect compression value: {} returned for XML (pointer file) input".format(
+        compression
+    )
+
+
+@pytest.mark.parametrize(
+    "compression,command",
+    [
+        (
+            utils.COMPRESSION_7Z_BZIP,
+            "7z a -bd -t7z -y -m0=bzip2 -mtc=on -mtm=on -mta=on -mmt=on /extract/filename.7z /full/path",
+        ),
+        (
+            utils.COMPRESSION_7Z_LZMA,
+            "7z a -bd -t7z -y -m0=lzma -mtc=on -mtm=on -mta=on -mmt=on /extract/filename.7z /full/path",
+        ),
+        (
+            utils.COMPRESSION_7Z_COPY,
+            "7z a -bd -t7z -y -m0=copy -mtc=on -mtm=on -mta=on -mmt=on /extract/filename.7z /full/path",
+        ),
+        (utils.COMPRESSION_TAR, "tar c -C /full -f /extract/filename.tar path"),
+        (
+            utils.COMPRESSION_TAR_GZIP,
+            "tar c -z -C /full -f /extract/filename.tar.gz path",
+        ),
+        (
+            utils.COMPRESSION_TAR_BZIP2,
+            "tar c -j -C /full -f /extract/filename.tar.bz2 path",
+        ),
+    ],
+)
+def test_get_compress_command(compression, command):
+    cmd, _ = utils.get_compress_command(
+        compression, "/extract/", "filename", "/full/path"
+    )
+    assert (
+        " ".join(cmd) == command
+    ), "Incorrect compression command: {0} returned for compression input {1}".format(
+        cmd, compression
+    )
+
+
+@pytest.mark.parametrize(
+    "compression,command",
+    [
+        (
+            utils.COMPRESSION_7Z_BZIP,
+            '#!/bin/bash\necho program="7z"\\; algorithm="bzip2"\\; version="`7z | grep Version`"',
+        ),
+        (
+            utils.COMPRESSION_7Z_LZMA,
+            '#!/bin/bash\necho program="7z"\\; algorithm="lzma"\\; version="`7z | grep Version`"',
+        ),
+        (
+            utils.COMPRESSION_7Z_COPY,
+            '#!/bin/bash\necho program="7z"\\; algorithm="copy"\\; version="`7z | grep Version`"',
+        ),
+        (
+            utils.COMPRESSION_TAR,
+            'echo program="tar"\\; algorithm=""\\; version="`tar --version | grep tar`"',
+        ),
+        (
+            utils.COMPRESSION_TAR_GZIP,
+            'echo program="tar"\\; algorithm="-z"\\; version="`tar --version | grep tar`"',
+        ),
+        (
+            utils.COMPRESSION_TAR_BZIP2,
+            'echo program="tar"\\; algorithm="-j"\\; version="`tar --version | grep tar`"',
+        ),
+    ],
+)
+def test_get_tool_info_command(compression, command):
+    cmd = utils.get_tool_info_command(compression)
+    assert (
+        cmd == command
+    ), "Incorrect tool info: {0} returned for compression input {1}".format(
+        cmd, compression
+    )
+
+
+@pytest.mark.parametrize(
+    "compression,cmd_output,expected_detail",
+    [
+        (
+            utils.COMPRESSION_7Z_BZIP,
+            "7z command\nVersion 3.0\nsomething else",
+            'program="7z"; version="Version 3.0"',
+        ),
+        (
+            utils.COMPRESSION_7Z_LZMA,
+            "7z command\nVersion 3.0\nsomething else",
+            'program="7z"; version="Version 3.0"',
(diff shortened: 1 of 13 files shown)


storage_service/common/utils.py
```diff
@@ -40,13 +40,22 @@
...
     COMPRESSION_TAR,
     COMPRESSION_TAR_BZIP2,
+    COMPRESSION_ZIP_DEFLATE,
+    COMPRESSION_TAR_GZIP,
 )
 
+PRONOM_7Z = "fmt/484"
+PRONOM_BZIP2 = "x-fmt/268"
+PRONOM_GZIP = "x-fmt/266"
+PRONOM_ZIP = "x-fmt/263"
```
ross-spencer: Good idea, thanks for pulling these out of the main body.

storage_service/common/utils.py
```diff
@@ -38,13 +38,15 @@
 # Compression options for packages
 COMPRESSION_7Z_BZIP = "7z with bzip"
 COMPRESSION_7Z_LZMA = "7z with lzma"
+COMPRESSION_7Z_COPY = "7z without compression"
```
ross-spencer, line 40: Thanks for adding this too!

storage_service/common/tests/test_utils.py
```diff
@@ -0,0 +1,185 @@
...
+def test_get_compression(pronom, algorithm, compression):
+    xml = (
+        '<?xml version="1.0"?>'
+        '<mets:mets xmlns:mets="http://www.loc.gov/METS/" xmlns:premis="info:lc/xmlns/premis-v2">'
+        " <premis:formatRegistryKey>%s</premis:formatRegistryKey>"
+        ' <mets:transformFile TRANSFORMALGORITHM="%s"></mets:transformFile>'
+        "</mets:mets>"
+    ) % (pronom, algorithm)
+
+    assert utils.get_compression(StringIO(xml)) == compression
```
ross-spencer: Nice! This is a big help.

ross-spencer: I find it a nice addition to have some sort of failure message after the assert so that it's absolutely clear what's not happening when the unit test is run, and potentially fails, so for this one:

`assert utils.get_compression(StringIO(xml)) == compression, "Incorrect compression value: {} returned for XML (pointer file) input".format(compression)` 

We could do it for each of the asserts below, but maybe it can be one to keep in mind for future PRs.

storage_service/common/tests/test_utils.py
```diff
@@ -0,0 +1,185 @@
...
+            utils.COMPRESSION_7Z_BZIP,
+            '#!/bin/bash\necho program="7z"\\; algorithm="bzip2"\\; version="`7z | grep Version`"',
+        ),
+        (
+            utils.COMPRESSION_7Z_LZMA,
+            '#!/bin/bash\necho program="7z"\\; algorithm="lzma"\\; version="`7z | grep Version`"',
+        ),
+        (
+            utils.COMPRESSION_7Z_COPY,
+            '#!/bin/bash\necho program="7z"\\; algorithm="copy"\\; version="`7z | grep Version`"',
```
ross-spencer, line 89: The way we build these commands are intense! Not sure they need to be so complex in current environments, but this refactors are a good first step to understanding that and making the calls more cleanly. :+1:

storage_service/locations/models/package.py
```diff
@@ -1568,7 +1555,6 @@ def _create_aip_premis_object(
             ".7z": {"puid": utils.PRONOM_7Z, "name": "7Zip format"},
             ".bz2": {"puid": utils.PRONOM_BZIP2, "name": "BZIP2 Compressed Archive"},
             ".gz": {"puid": utils.PRONOM_GZIP, "name": "GZIP Compressed Archive"},
-            ".zip": {"puid": utils.PRONOM_ZIP, "name": "Zip Compressed Archive"},
```
ross-spencer: :+1:

storage_service/locations/models/package.py
```diff
@@ -2981,109 +2855,9 @@ def _update_pointer_file(self, compression, root=None, path=None):
...
-                x for x in subprocess.check_output("7z").splitlines() if "Version" in x
-            ][0]
-            format_info = {
-                "name": "Zip Compressed Archive",
-                "registry_name": "PRONOM",
-                "registry_key": utils.PRONOM_ZIP,
-                "program_name": "zip",
-                "program_version": version,
-            }
+        transform_file, format_info = utils.get_compression_transforms(
```
ross-spencer: Though I think the chances are slim, should we try/except for `ValueError` here in-case `utils.get_compression_transforms()` returns empty or without enough values? We can raise StorageException with the err?

storage_service/locations/models/package.py
```diff
@@ -2602,20 +2506,16 @@ def finish_reingest(
...
                         "Failed to extract valid compression algorithm from"
                         ' "{}"; does not match any of the following recognized'
                         " options: {}".format(
-                            compression_algorithm, ", ".join(utils.COMPRESSION_ALGORITHMS)
+                            compression_algorithm,
+                            ", ".join(utils.COMPRESSION_ALGORITHMS),
                         )
                     )
+                    LOGGER.error(msg)
+                    raise StorageException(msg)
```
ross-spencer, line 2323: Perhaps remove the logging in favor of `raise StorageException(msg)` my understanding is both should return to the console? What do you think?

ross-spencer (changes requested):
Two minor rfcs but otherwise looks really good @helenst this PR really helps us a lot. Thank you! I've also just approved the corresponding change in Archivematica, so we should be good to pull those into `qa` soon.

storage_service/common/utils.py
```diff
@@ -291,6 +306,253 @@ def get_compression(pointer_path):
...
+            )
+        )
+
+    elif compression == COMPRESSION_TAR_GZIP:
+        transforms.append(
+            etree.Element(
+                PREFIX_NS["mets"] + "transformFile",
+                TRANSFORMORDER=str(transform_order),
+                TRANSFORMTYPE="decompression",
+                TRANSFORMALGORITHM="gzip",
```
ross-spencer: Discussing this internally, we think perhaps representing that tar component of the gzip might be important, is it possible to change this to `tar.gzip` do you think that makes sense for your team as well?

ross-spencer (changes requested):
Following up with just one change to update the `TRANSFORMALGORITHM` field. And following that I'll ask you to rebase the commits, and then we'll get this merged. 

I am tempted to wait until after this is fixed for `1.10` in case there are any corresponding changes required here or in Archivematica: https://github.com/archivematica/Issues/issues/713

storage_service/locations/models/package.py
helenst: I think i put this line in because it seemed that the error was just being swallowed (i.e. I was getting 500 errors on the API call and nothing logged) - this being called by `obj_update_hook` on the tastypie endpoint. It's possible there's something in my setup that's causing this but I'm not sure.

ross-spencer: @helenst I have been having this exact issue today trying to debug S3 so I'd keep it if you haven't removed it already. I hadn't quire realized how little the storage service was logging when things weren't quite performing as expected.

storage_service/locations/models/package.py
helenst: Just realised the calling code is wrong here - I originally wrote `get_compression_transforms` to return the format info too and package them up in a tuple, and somehow didn't update the call. Correcting this now.

storage_service/common/utils.py
helenst: I don't know whether we have a preference. Currently the code adds a separate compression transform element for `tar`. Same as `pbzip2` compression method. But this can change. I developed the gzip additions using the pbzip2 compression as a template (since they both use tar and seem to be very similar) - whatever is decided about this, should `pbzip2` be treated the same way?

helenst:
@ross-spencer Updated and responded to comments - let me know what you think about the tar/gzip thing?

ross-spencer:
Thanks @helenst I should be able to get back to this later today, and if not, first thing tomorrow morning!

storage_service/common/utils.py
ross-spencer: Thanks @helenst "tar gzip" would be great in the `TRANSFORMALGORITHM` too. It wouldn't normally be your responsibility making a PR like this to change `pbzip2` at the same time, but if you had a spare moment. Only annoying thing is that the change needs to be made in Archivematica as far as I can tell.

I had to investigate this today as I noticed pbzip2 is outputting blank: 

```
      <mets:file ID="pbzip-ae0b9327-1b7f-43f9-8e07-4be4830259c5.tar" GROUPID="Group-ae0b9327-1b7f-43f9-8e07-4be4830259c5" ADMID="amdSec_113644">
        <mets:FLocat xlink:href="/var/archivematica/sharedDirectory/www/AIPsStore/ae0b/9327/1b7f/43f9/8e07/4be4/8302/59c5/pbzip-ae0b9327-1b7f-43f9-8e07-4be4830259c5.tar.bz2" LOCTYPE="OTHER" OTHERLOCTYPE="SYSTEM"/>
        <mets:transformFile TRANSFORMTYPE="decompression" TRANSFORMORDER="1" TRANSFORMALGORITHM=""/>
      </mets:file>
```

It turns out there needs to be a second component in `workflow.json`, 

* [pbzip2](https://github.com/artefactual/archivematica/blob/qa/1.x/src/MCPServer/lib/assets/workflow.json#L1056)
* [gzip workflow](https://github.com/artefactual/archivematica/blob/qa/1.x/src/MCPServer/lib/assets/workflow.json#L1044)

Would need to read something like:

* `"pbzip-tar.pbzip2"`,
* `"gzip-tar.gzip":`

So that here: https://github.com/artefactual/archivematica/blob/qa/1.x/src/MCPClient/lib/clientScripts/compress_aip.py#L45 the algorithm is picked up. 

That will automatically be picked up for `pbzip2` but then for gzip, we need to add a string replacement `{}` [here](https://github.com/artefactual/archivematica/blob/9a6fe32f2dc847038acf895a4356d0dbc8e7779d/src/MCPClient/lib/clientScripts/compress_aip.py#L105-L106) to match the other functions. 

I should have picked up on this before. 

If that's okay to make the change, I'll get that PR reviewed as quickly as possible. 

Thank you!

(And I think in future we can probably streamline this in code quite a lot, but it's beyond the scope of what we're doing now)

ross-spencer:
Thanks @helenst I have requested the change to the algorithm, and pbzip2 if you have the chance, if you can make that change, we'll get the two PRs merged side-by-side. 

I'm off shortly today, but around all day today to do that.

helenst:
OK - I've updated this PR and created one on Archivematica: https://github.com/artefactual/archivematica/pull/1435

ross-spencer:
Thanks @helenst. I have had to bring together this work on another branch supporting PREMIS 3: https://github.com/artefactual/archivematica-storage-service/pull/469 most of your work is now over there and the changes are nearly complete. I'll make the final changes there before merging.

ross-spencer:
Closing as superseded, and fulfilled by https://github.com/artefactual/archivematica-storage-service/commit/d4728fb4ae0c7785be5f431980aba9903c59dbc2.

PR #455, opened 2019-05-21
Pin python package requirements using pip-tools

helenst:
Addresses https://github.com/archivematica/Issues/issues/634


cole (approved):
Tested out and looks good!

PR #458, opened 2019-05-27
Fix errors blocking replication to multiple locations

cole:
The changes in PR #453 didn't handle the case where we have more than one replication PREMIS event, so this PR addresses that. Also fixes an issue where the second replica was related to the first replica (and so on, with any subsequent replicas) rather than all replicas being related to the master AIP.

Connects to https://github.com/artefactual/archivematica-storage-service/issues/270

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -553,7 +553,8 @@ class Package(models.Model):
             replicator_location.uuid,
         )
 
-        replica_package = _replicate_package_mdl_inst(self)
+        replica_package = self._clone()
+        replica_package.replicated_package = self
 
         # Remove the /uuid/path from the replica's current_path and replace the
         # old UUID in the basename with the new UUID.
@@ -2792,6 +2793,17 @@ class Package(models.Model):
     def has_been_submitted_for_processing(self):
         return "deposit_completion_time" in self.misc_attributes
 
+    def _clone(self):
+        """Create a new ``Package`` instance that is exactly like ``package`` but has
+        a new primary key and UUID.
+        """
+        clone = copy.deepcopy(self)
+        clone.pk = None
+        clone.uuid = None
+        clone.save()  # Generate a new id and UUID
+
+        return clone
+
 
 def _get_decompr_cmd(compression, extract_path, full_path):
     """Returns a decompression command (as a list), given ``compression``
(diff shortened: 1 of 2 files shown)


storage_service/locations/tests/test_package.py
```diff
@@ -370,13 +370,36 @@ def test_replicate_aip(self):
...
 
         aip.create_replicas()
-        replica = aip.replicated_package
 
+        assert aip.replicas.count() == 1
+
+        replica = aip.replicas.first()
         assert replica is not None
-        # The relationship is a little bit broken here; one would think that the
-        # _original_ AIP should have replicas, and the replicated AIP should not.
```
cole, line 377: Turns out, the relationship was on the right side, the `create_replicas` method was just modifying `aip` in place.

ross-spencer:
Whoops! @replaceafill sorry, ignore my button-mashing here! I was trying to figure out if this work was still unassigned, turns out all is good! :smile:

storage_service/locations/models/package.py
```diff
@@ -1196,14 +1191,22 @@ def create_new_pointer_file_with_replication(
...
         )
-        new_relationships = old_premis_object.findall("relationship")
-        new_relationships += (replication_relationship,)
-        new_premis_object = premisrw.PREMISObject(
-            xsi_type=old_premis_object.xsi_type,
-            object_identifier=old_premis_object.find("object_identifier"),
-            object_characteristics=old_premis_object.find("object_characteristics"),
-            relationship=new_relationships,
-        )
+        new_relationships = [
```
cole, line 1213: @replaceafill this change (using the data attr of the old relationship) should fix the `AttributeError: Instance of <class 'abc.PREMISRelationshipElement'> has no attribute split.` issue — what is happening is using kwargs to construct a PREMISObject doesn't allow multiple values in metsrw right now, unfortunately.

Sorry, I forgot to note that anywhere — it would have been helpful to do so.

replaceafill: Thank you @cole. It's indeed fixed by this.

storage_service/locations/models/package.py
```diff
@@ -3247,18 +3250,17 @@ def _get_compression_details_from_premis_events(premis_events, aip_uuid):
     return compression_event.compression_details
 
 
-def _replicate_package_mdl_inst(package_mdl):
+def _replicate_package_mdl_inst(package):
     """Create a new Django ``Package`` instance that is exactly like
     ``package_mdl`` but which has a new primary key and UUID, and references
     ``package_mdl`` as its ``replicated_package``.
```
replaceafill: Should these `package_mdl`s in the docstring be updated?

replaceafill (commented):
@cole I tried to verify the functionality according to https://github.com/artefactual/archivematica-storage-service/issues/270 I can see the two replication events in the pointer file for the original AIP but I can't find the `two corresponding validation events` mentioned.

storage_service/locations/models/package.py
cole: Good point! I changed the naming of this method in 80efe62, should be clearer now.

replaceafill: It is! Thank you!

replaceafill (approved):
:+1: I can see the creation and validation events in the pointer files of each replica as described in https://github.com/artefactual/archivematica-storage-service/issues/270#issuecomment-501880461

PR #460, opened 2019-05-28
Fix delete links in Packages tab

replaceafill:
This attaches the event handlers of the delete links to the body of
the page. That way table rows past the initial display or loaded
dynamically through the different navigation options get the handlers.

Connects to https://github.com/archivematica/Issues/issues/711


cole (approved):
Looks good!

PR #462, opened 2019-06-01
DuraSpace: fix chunking of contents in upload

sevein:
- Include the ``index`` of each chunk in the manifest, which is a required
  attribute of chunk entries according to the XSD: https://git.io/fjEn5.
  The missing attribute was impeding the download of contents by DuraCloud
  Retrieval Tool.
- Strip leading slash in ``contentId`` and ``chunkId`` which came as a
  recommendation by the DuraCloud team. In my test, this was not the problem
  causing the retrieval tool to fail.
- Update the size of each chunk to 1 GB instead of 1 GiB to mimic the default
  behaviour seen in DuraCloud Sync Tool and the examples seen in their docs.
  This is not strictly necessary.

Connects to https://github.com/archivematica/Issues/issues/574.

### DIFF
--- a/storage_service/locations/models/duracloud.py
+++ b/storage_service/locations/models/duracloud.py
@@ -61,7 +61,15 @@ class Duracloud(models.Model):
     ]
 
     MANIFEST_SUFFIX = ".dura-manifest"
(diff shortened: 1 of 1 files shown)


cole (approved):
🎉

PR #463, opened 2019-06-03
S3 fixes to improve the chances of deletion and user feedback

ross-spencer:
In this commit we test for leading-slashes in paths we are attempting to delete from S3 and remove them where the slash impedes Boto3's ability to find and then delete the package (as per https://github.com/archivematica/Issues/issues/696#issuecomment-498646029). We also try and provide a bit more information to the user when something goes wrong. 

Connected to archivematica/issues#696

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -5,6 +5,7 @@
 import json
 import logging
 import os
+import pprint
 import re
 import shutil
 import urllib
@@ -1659,11 +1660,15 @@ class PackageResource(ModelResource):
     def _attempt_package_request_event(
         self, package, request_info, event_type, event_status
     ):
-
-        import pprint
-
-        LOGGER.info("PACKAGE DELETE REQUEST:")
-        LOGGER.info(pprint.pformat(request_info))
+        """Generic package request handler, e.g. package recovery: RECOVER_REQ,
+        or package deletion: DEL_REQ.
+        """
+        LOGGER.info(
+            "Package event: '{}' requested, with package status: '{}'".format(
+                event_type, event_status
+            )
+        )
+        LOGGER.debug(pprint.pformat(request_info))
 
         pipeline = Pipeline.objects.get(uuid=request_info["pipeline"])
         request_description = event_type.replace("_", " ").lower()
@@ -1684,11 +1689,6 @@ class PackageResource(ModelResource):
                 store_data=package.status,
             )
             request_event.save()
-
-            # Update package status
-            package.status = event_status
-            package.save()
-
             response = {
                 "message": _("%(event_type)s request created successfully.")
                 % {"event_type": request_description.title()},
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 from __future__ import absolute_import
 
 # stdlib, alphabetical
@@ -1935,10 +1936,10 @@ class Package(models.Model):
                 error = lom._delete_update_lom(self, delete_lom_ids)
         try:
             self.current_location.space.delete_path(self.full_path)
-        except Exception as e:
-            error = e.message
-
-        # Remove pointer file, and the UUID quad directories if they're empty
+        except (StorageException, NotImplementedError, ValueError) as err:
+            return False, err
+        # If deleted correctly, remove pointer file, and the UUID quad
+        # directories if they're empty.
         pointer_path = self.full_pointer_file_path
         if pointer_path:
             try:
@@ -1954,8 +1955,6 @@ class Package(models.Model):
                 os.path.dirname(self.pointer_file_path),
                 base=self.pointer_file_location.full_path,
             )
-
-        self.status = self.DELETED
         self.save()
         return True, error
 
--- a/storage_service/locations/models/s3.py
+++ b/storage_service/locations/models/s3.py
@@ -3,6 +3,7 @@ from __future__ import absolute_import
 # stdlib, alphabetical
 import logging
 import os
+import pprint
 
 # Core Django, alphabetical
 from django.db import models
@@ -128,12 +129,30 @@ class S3(models.Model):
         }
 
     def delete_path(self, delete_path):
-        objects = self.resource.Bucket(self._bucket_name()).objects.filter(
+        """Delete an object from an S3 bucket. We assume an object exists, if
+        it doesn't then the generator returned by the S3 library (Boto3) cannot
+        be iterated, and we raise a StorageException.
+        """
+        if delete_path.startswith(os.sep):
+            LOGGER.info(
+                "S3 path to delete {} begins with {}; removing from path prior to deletion".format(
+                    delete_path, os.sep
+                )
+            )
+            delete_path = delete_path.lstrip(os.sep)
+        obj = self.resource.Bucket(self._bucket_name()).objects.filter(
             Prefix=delete_path
         )
-
-        for objectSummary in objects:
-            objectSummary.delete()
+        items = False
+        for object_summary in obj:
+            items = True
+            resp = object_summary.delete()
+            LOGGER.debug("S3 response when attempting to delete:")
+            LOGGER.debug(pprint.pformat(resp))
+        if not items:
+            err_str = "No packages found in S3 at: {}".format(delete_path)
+            LOGGER.warning(err_str)
+            raise StorageException(err_str)
 
     def move_to_storage_service(self, src_path, dest_path, dest_space):
         self._ensure_bucket_exists()
--- a/storage_service/locations/views.py
+++ b/storage_service/locations/views.py
@@ -230,10 +230,11 @@ def _handle_package_request(request, config, view_name):
                 event = req.form.save()
                 event.status_reason = req.form.cleaned_data["status_reason"]
                 event.admin_id = auth.get_user(request)
-
                 # Handle administrator decision and optionally notify remote REST endpoint
                 if "reject" in request.POST:
                     event.status = Event.REJECTED
+                    # Request is rejected so the package status set back to
+                    # what it was stored as previously.
                     event.package.status = event.store_data
                     notification_message = _handle_package_request_remote_result_notification(
                         config, event, False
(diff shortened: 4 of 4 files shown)


storage_service/locations/models/s3.py
```diff
@@ -128,12 +128,29 @@ def browse(self, path):
...
-        for objectSummary in objects:
-            objectSummary.delete()
+        items = False
+        for objectSummary in obj:
+            items = True
+            resp = objectSummary.delete()
+            LOGGER.info("S3 response when attempting to delete: %s", resp)
+        if not items:
+            err_str = "No packages found in S3 at: {}".format(delete_path)
+            LOGGER.warning(err_str)
```
cole, line 154: Aren't exceptions logged anyways? I.e. do we need to both log a warning and raise?

storage_service/locations/models/s3.py
```diff
@@ -128,12 +128,29 @@ def browse(self, path):
...
             Prefix=delete_path
         )
-
-        for objectSummary in objects:
-            objectSummary.delete()
+        items = False
+        for objectSummary in obj:
+            items = True
+            resp = objectSummary.delete()
+            LOGGER.info("S3 response when attempting to delete: %s", resp)
```
cole: Maybe this is more debug level logging? Here's what I get when testing with minio for a success response:

```INFO      2019-06-13 12:35:13  locations.models.s3:s3:delete_path:149:  S3 response when attempting to delete: {'ResponseMetadata': {'HTTPStatusCode': 204, 'RetryAttempts': 0, 'HostId': '', 'RequestId': '15A7D91CB99426C4', 'HTTPHeaders': {'x-xss-protection': '1; mode=block', 'x-minio-deployment-id': 'b7808373-b4c1-41bb-929d-bed6c135c7e6', 'content-security-policy': 'block-all-mixed-content', 'accept-ranges': 'bytes', 'vary': 'Origin', 'server': 'MinIO/RELEASE.2019-05-14T23-57-45Z', 'date': 'Thu, 13 Jun 2019 19:35:13 GMT', 'x-amz-request-id': '15A7D91CB99426C4', 'x-amz-bucket-region': 'minio'}}}```

I can't find the documentation (boto3's is pretty useless), but I believe that if there's an error deleting the object (e.g. permisisons) then object.delete should raise an exception.

File: storage_service/locations/models/s3.py [hunk 3]
```diff
@@ -128,12 +128,29 @@ def browse(self, path):
...
+            )
+            delete_path = delete_path.lstrip(os.sep, "")
+        obj = self.resource.Bucket(self._bucket_name()).objects.filter(
             Prefix=delete_path
         )
-
-        for objectSummary in objects:
-            objectSummary.delete()
+        items = False
+        for objectSummary in obj:
```
cole: Might be a case for snake_case 🐍

cole (approved):
Looks good! I've noted a few nitpicks — I'm ok though if you choose to merge and leave them for now.

storage_service/locations/models/s3.py
ross-spencer: It is a good point. I have been tempted to log an issue about this, it's inconsistent. Because it's the responsibility of the caller to decide what to do with an exception, the caller might not always log it back to `stderr`. In this case, if I force the error, we get a UI output, but no log output, so I create a log at the point before the instead. So we just see the following on the command line:
```
archivematica-storage-service_1  | DEBUG     2019-06-13 22:57:18  locations.models.async_manager:async_manager:_watchdog_loop:122:  Watchdog sees 0 tasks running
archivematica-storage-service_1  | INFO      2019-06-13 22:57:18  locations.models.s3:s3:delete_path:138:  S3 path to delete /will-not-be-deleted/5d93/dda0/055d/4cf9/8304/129f/3ab3/a699/rs-s3-1-5d93dda0-055d-4cf9-8304-129f3ab3a699.7z begins with /; removing from path prior to deletion
archivematica-storage-service_1  | WARNING   2019-06-13 22:57:19  locations.models.s3:s3:delete_path:152:  No packages found in S3 at: will-not-be-deleted/5d93/dda0/055d/4cf9/8304/129f/3ab3/a699/rs-s3-1-5d93dda0-055d-4cf9-8304-129f3ab3a699.7z
```
**Example UI output:**
image

storage_service/locations/models/s3.py
ross-spencer: Good idea!

storage_service/locations/models/s3.py
ross-spencer: It's a good opportunity! Will do!

ross-spencer:
@cole thank you for your comments and review. I've addressed a handful. In the process of considering the PR I was reminded that there is a challenge with S3 and the status of a package as well. It will say deleted even if the deletion failed. I have rectified that in further commits, and so will seek re-review from you if you have time. I have tried to keep the commits as logically separated as possible.

storage_service/locations/models/s3.py
cole: Sounds good. Too bad about the inconsistency, but in that case better to log as well.

storage_service/locations/api/resources.py
```diff
@@ -1685,9 +1687,11 @@ def _attempt_package_request_event(
             )
             request_event.save()
 
-            # Update package status
-            package.status = event_status
-            package.save()
+            # Update package status, if and only if this isn't a deletion
+            # request where we don't want to mistakenly pre-mark it as deleted.
+            if event_status != Package.DELETED:
```
cole: It looks like this would always be true — event status would be `Event.DELETE` for deletion.

storage_service/locations/models/package.py
```diff
@@ -1931,10 +1932,19 @@ def delete_from_storage(self):
...
-        # Remove pointer file, and the UUID quad directories if they're empty
+        except (StorageException, NotImplementedError, ValueError) as err:
+            # The package hasn't been deleted and the deletion request needs
+            # to be resubmitted, so status is UPLOADED as it was prior to the
+            # deletion request. We also want to avoid Schrödinger's package
+            # when trying to delete something already DELETED. Only set status
+            # back to UPLOADED state if it hasn't already been set to DELETED.
+            # This will mean it is still visible in Archivematica.
+            if self.status != self.DELETED:
+                self.status = self.UPLOADED
```
cole: I agree with the reasoning in the comment here, but it seems like there may be some other status cases (VERIFIED, etc). We are already setting package status on success only in the view — maybe it would be simpler to just not update package status at all in this function (remove the update on line 1964).

cole (changes requested):
@ross-spencer np! I have some feedback about the statuses but I may just be misunderstanding how it works.

storage_service/locations/api/resources.py
ross-spencer: I think I could have commented this better. I still haven't found the right balance for comments/not comments. In this instance, I am trying to be as cautious as possible. Because `_attempt_package_request_event` can handle I think three other event types, I want to make sure this change right now doesn't inadvertently change any of the other status values that might be output by those. So the function here can perform entirely as before when a non `DELETE` request comes in. (Providing I've coded that correctly!)

storage_service/locations/models/package.py
ross-spencer: That sounds like a plan. I'll take a look. (I was torn between that and not removing `1964` the second opinion helps!)

ross-spencer: I have logged a ticket against this [here](https://github.com/archivematica/Issues/issues/760). We only have one state in memory at this point in the code's execution, `DEL_REQ` and you are right to say there are more states than `UPLOADED`. As such, I will implement a fix to make sure we set this back to the previous state. That will be `DEL_REQ`. I'm not a fan, but  1. I feel this is worth investigating further without changing too much of the service's behavior and 2. this change now feels out of scope for where this PR began as we reach code-freeze, and it would only be a good idea if it were a genuinely quick fix.

ross-spencer: Previous comment was incorrect. Will remove this here. Importantly, this should ensure that if a package isn't deleted correctly, its status isn't updated to `DELETED`.

storage_service/locations/api/resources.py
ross-spencer: I'll revert this too.

storage_service/locations/models/s3.py
ross-spencer: RE: the exception. In the specific case we're writing this code against, Boto3 isn't throwing an exception. This might be the version of the library at time of writing which is why we're doing this in a more ask-permission kind of way. 
```
   11: boto3==1.6.5
   12: botocore==1.9.23          # via boto3, s3transfer
```
We have another PR in the works where Boto3 is upgraded. Once that is done and we reach code-freeze how we potentially need to handle an exception will come out in testing.

ross-spencer:
I got caught up in [the weeds](https://github.com/archivematica/Issues/issues/760) with this one somewhat but think this should be good to re-review.

replaceafill (approved):
This looks good to me. I checked that [Cole's comments](https://github.com/artefactual/archivematica-storage-service/pull/463#pullrequestreview-250059137) were addressed. I also functionally tested this getting the AIP deleted from my bucket and better logging:

```
archivematica-storage-service_1  | INFO      2019-06-24 21:02:08  locations.models.s3:s3:delete_path:139:  S3 path to delete /abs-path/with/slash/ae34/1db2/3337/4061/a09e/f733/e0c1/467c/DouglasPicturesS3SecondTest-ae341db2-3337-4061-a09e-f733e0c1467c.7z begins with /; removing from path prior to deletion
archivematica-storage-service_1  | DEBUG     2019-06-24 21:02:09  locations.models.s3:s3:delete_path:150:  S3 response when attempting to delete:
archivematica-storage-service_1  | DEBUG     2019-06-24 21:02:09  locations.models.s3:s3:delete_path:151:  {'ResponseMetadata': {'HTTPHeaders': {'date': 'Tue, 25 Jun 2019 04:02:09 GMT',
archivematica-storage-service_1  |                                       'server': 'AmazonS3',
...
archivematica-storage-service_1  |                       'RetryAttempts': 0}}
```

ross-spencer:
Thanks @replaceafill and @cole!

PR #464, opened 2019-06-05
Fix space browser when counter is disabled

mamedin:
Connects to issue https://github.com/archivematica/Issues/issues/657


ross-spencer:
Hi @mamedin it looks like there is a failing unit test that you need to resolve with this PR first. I haven't looked at the cause, but let me know if I can help at all.

sevein:
I'll give it a shot today!

mamedin (commented):
Thanks Jesús, it works fine :)

sevein (approved):
:+1:

PR #465, opened 2019-06-05
Disallow move from transfer source

cole:
Relates to: https://github.com/archivematica/Issues/issues/646

It looks like the defect was introduced when we reduced the number of copies done in https://github.com/artefactual/archivematica-storage-service/pull/320.

This PR builds on work done in https://github.com/artefactual/archivematica-storage-service/pull/422.

### DIFF
--- a/storage_service/locations/models/location.py
+++ b/storage_service/locations/models/location.py
@@ -44,7 +44,7 @@ class Location(models.Model):
     REPLICATOR = "RP"
 
     # List of purposes where moving is not allowed.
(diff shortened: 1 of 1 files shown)


cole:
@hakamine can you please see if you can replicate the issue using the PR branch? Due to docker limitations I couldn't reproduce it exactly, but from the logging in SS I believe this change should fix it.

hakamine (approved):
@cole Thank you for the fix. Testing on a VM it works as expected (transfer source is not deleted). Tried a standard and a zipped bag transfer.

PR #466, opened 2019-06-06
Update pointer file PREMIS version

cole:
Relates to: https://github.com/archivematica/Issues/issues/713

Update metsrw to the latest release for PREMIS 3 compatibility, and update PREMIS generation to use the correct format.

The bulk of this PR is consolidating PREMIS tuple construction in `common.premis` — please note when reviewing that most of the changed code has been moved (mostly as is, with a few changes for version 3 and to accommodate different imports and references to packages) to try to make updates like this a bit easier in future. The consolidation isn't "done" by any means; there's still a lot of PREMIS and METS related logic in the Package model.

### DIFF
--- a/storage_service/common/management/commands/import_aip.py
+++ b/storage_service/common/management/commands/import_aip.py
@@ -63,7 +63,7 @@ from django.db.utils import IntegrityError
 from django.utils.six.moves import input
 
 from administration.models import Settings
-from common import utils
+from common import premis, utils
 from locations import models
 
 
@@ -360,7 +360,7 @@ def compress(aip_model_inst, compression_algorithm):
     """Use the Package model's compress_package method to compress the AIP
     being imported, update the Package model's ``size`` attribute, retrieve
     PREMIS agents and event for the compression (using the package model's
-    ``get_premis_aip_compression_event`` method) and return a 3-tuple:
+    ``create_premis_aip_compression_event`` method) and return a 3-tuple:
     (aip_model_inst, compression_event, compression_agents).
     """
     if not compression_algorithm:
@@ -379,8 +379,8 @@ def compress(aip_model_inst, compression_algorithm):
     aip_model_inst.current_path = new_current_path
     shutil.rmtree(compressed_aip_parent_path)
     aip_model_inst.size = utils.recalculate_size(new_full_path)
-    compression_agents = utils.get_ss_premis_agents()
-    compression_event = aip_model_inst.get_premis_aip_compression_event(
+    compression_agents = [premis.SS_AGENT]
+    compression_event = premis.create_premis_aip_compression_event(
         details["event_detail"],
         details["event_outcome_detail_note"],
         agents=compression_agents,
--- a/storage_service/common/premis.py
+++ b/storage_service/common/premis.py
@@ -1,0 +1,349 @@
+"""
+PREMIS metadata generation.
+"""
+from __future__ import absolute_import, unicode_literals
+
+import uuid
+
+import metsrw
+from django.utils import timezone
+
+from storage_service import __version__ as ss_version
+
+
+PREMIS_META = metsrw.plugins.premisrw.PREMIS_3_0_META
+SS_AGENT = metsrw.plugins.premisrw.PREMISAgent(
+    data=(
+        "agent",
+        PREMIS_META,
+        (
+            "agent_identifier",
+            ("agent_identifier_type", "preservation system"),
+            (
+                "agent_identifier_value",
+                "Archivematica-Storage-Service-{}".format(ss_version),
+            ),
+        ),
+        ("agent_name", "Archivematica Storage Service"),
+        ("agent_type", "software"),
+    )
+)
+
+
+def timestamp():
+    return timezone.now().strftime("%Y-%m-%dT%H:%M:%S")
+
+
+def add_agents_to_event_as_list(event, agents):
+    """Add agents in ``agents`` to the list ``event`` which represents a
+    PREMIS:EVENT.
+    :param list event: a PREMIS:EVENT represented as a list
+    :param iterable agents: an iterable of premisrw.PREMISAgent instances.
+    """
+    for agent in agents:
+        event.append(
+            (
+                "linking_agent_identifier",
+                ("linking_agent_identifier_type", agent.identifier_type),
+                ("linking_agent_identifier_value", agent.identifier_value),
+            )
+        )
+    return event
+
+
+def create_replication_event(
+    original_package_uuid, replica_package_uuid, event_uuid=None, agents=None
+):
+    """Return a PREMISEvent for replication of an AIP.
+    """
+    outcome_detail_note = (
+        "Replicated Archival Information Package (AIP) {} by creating"
+        " replica {}.".format(original_package_uuid, replica_package_uuid)
+    )
+    if not agents:
+        agents = [SS_AGENT]
+    if not event_uuid:
+        event_uuid = str(uuid.uuid4())
+    event = [
+        "event",
+        PREMIS_META,
+        (
+            "event_identifier",
+            ("event_identifier_type", "UUID"),
+            ("event_identifier_value", event_uuid),
+        ),
+        ("event_type", "replication"),
+        ("event_date_time", timestamp()),
+        (
+            "event_detail_information",
+            ("event_detail", "Replication of an Archival Information Package"),
+        ),
+        (
+            "event_outcome_information",
+            ("event_outcome", "success"),
+            (
+                "event_outcome_detail",
+                ("event_outcome_detail_note", outcome_detail_note),
+            ),
+        ),
+    ]
+    event = tuple(add_agents_to_event_as_list(event, agents))
+
+    return metsrw.plugins.premisrw.PREMISEvent(data=event)
+
+
+def create_premis_aip_creation_event(
+    package_uuid, master_aip_uuid=None, agents=None, inst=True
+):
+    """Return a PREMISEvent for creation of an AIP."""
+    if master_aip_uuid:
+        outcome_detail_note = (
+            "Created Archival Information Package (AIP) {} by replicating"
+            " previously created AIP {}".format(package_uuid, master_aip_uuid)
+        )
+    else:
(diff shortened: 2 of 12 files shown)


cole:
Note that due to https://github.com/archivematica/Issues/issues/655, there's still an error when validating the pointer file: "There must be PREMIS elements inside the METS container".

cole:
@ross-spencer FYI these changes may cause conflicts with your work in https://github.com/archivematica/Issues/issues/660 — sorry, I didn't realize you were also working on this area! I think we'll have to see what lands first, but happy to rebase this or help with rebasing, whatever is easier.

storage_service/common/premis.py
```diff
@@ -0,0 +1,335 @@
...
+                ("event_outcome_detail_note", outcome_detail_note),
+            ),
+        ),
+    ]
+    event = tuple(add_agents_to_event_as_list(event, agents))
+
+    return metsrw.plugins.premisrw.PREMISEvent(data=event)
+
+
+def get_replication_derivation_relationship(
```
ross-spencer: I think beyond the scope of the changes you want to make here. With the refactor, in future, perhaps we can take the time to change to `create_` or something approximating that. They don't seem to do a lot of getting! :smile:

storage_service/locations/models/package.py
```diff
@@ -619,8 +618,10 @@ def replicate(self, replicator_location_uuid):
                 replica_package.uuid,
                 master_checksum_algorithm,
             )
-            replication_validation_event = replica_package.get_replication_validation_event(
-                checksum_report=checksum_report, master_aip_uuid=self.uuid
+            replication_validation_event = premis.get_replication_validation_event(
+                replica_package,
```
ross-spencer: For consistency here, I think it would be good to see the form `replica_package_uuid=replica_package.uuid,` as with the other args. Is there a good reason to pass the entire object into the function at the moment, instead of just using the uuid?

storage_service/locations/models/gpg.py
```diff
@@ -293,16 +291,6 @@ def _encr_path2key_fingerprint(encr_path):
         raise GPGException(fail_msg)
 
 
-# This replaces non-unicode characters with a replacement character,
-# and is primarily used for arbitrary strings (e.g. filenames, paths)
-# that might not be valid unicode to begin with.
-# NOTE: non-DRY from archivematicaCommon/archivematicaFunctions.py
-def escape(string):
```
ross-spencer, line 300: I'm not sure the circumstances for this code previously, is it the use of `unicode_literals` that helps us to manage this when it's output in the new module?

storage_service/locations/models/package.py
```diff
@@ -1010,8 +1011,12 @@ def _create_pointer_file_write_to_disk(
         __, compression_program_version, archive_tool = _get_compression_details_from_premis_events(
             premis_events, self.uuid
         )
-        premis_object = self._create_aip_premis_object(
-            checksum_algorithm, checksum, archive_tool, compression_program_version
+        premis_object = premis.get_aip_premis_object(
+            self,
+            checksum_algorithm,
```
ross-spencer, line 1019: Now the package is being passed into the function, are all of these args needed? E.g. will checksum algorithm be available as `package.DEFAULT_CHECKSUM_ALGORITHM`? 

Same question for below's replication package object creation.

storage_service/locations/models/package.py
```diff
@@ -1140,19 +1146,22 @@ def create_replica_pointer_file(
...
             master_checksum_algorithm,
             master_checksum,
             archive_tool,
             compression_program_version,
             premis_relationships=replica_premis_relationships,
         )
-        replica_premis_creation_event = replica_package.get_premis_aip_creation_event(
-            master_aip_uuid=master_aip_uuid, agents=replica_premis_creation_agents
+        replica_premis_creation_event = premis.get_premis_aip_creation_event(
+            replica_package,
```
ross-spencer:
```suggestion
            package=replica_package,
```
Though we're really not very consistent across this module.... :confounded: :nerd_face:

ross-spencer (changes requested):
This is really great @cole. We've a few pieces of work in this area at the minute, my small change to the package name, @helenst has one cleaning up the way the compression artifacts are recorded in the pointer file too as well as adding to the methods. So it's all quite good, even if the timing is a little awkward at times! :wink: 

I think it makes sense to merge yours first, and then we can work out what's needed to rebase the other two. 

With regards to my comments, the code looks great. I've made some very minor style suggestions. I've asked one question about passing the package into the PREMIS functions. Those questions may or may not result in changes.

storage_service/common/utils.py
```diff
@@ -351,63 +349,10 @@ def coerce_str(string):
...
-            )
-        )
-    return event
-
-
 def mets_file_now():
     return datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%S")
 
 
-def get_ss_premis_agents(inst=True):
```
ross-spencer, line 375: I missed this, but there are a couple of left-over references to this function in the code, e.g. when creating replicated packages: 
```python
/home/ross-spencer/git/artefactual-labs/am/src/archivematica-storage-service/storage_service/common/management/commands/import_aip.py:
  380      shutil.rmtree(compressed_aip_parent_path)
  381      aip_model_inst.size = utils.recalculate_size(new_full_path)
  382:     compression_agents = utils.get_ss_premis_agents()
  383      compression_event = premis.get_premis_aip_compression_event(
  384          details["event_detail"],

/home/ross-spencer/git/artefactual-labs/am/src/archivematica-storage-service/storage_service/locations/models/package.py:
 1142  
 1143          # 3. Construct the pointer file and return it
 1144:         replica_premis_creation_agents = utils.get_ss_premis_agents()
 1145          __, compression_program_version, archive_tool = (
 1146              master_compression_event.compression_details
 ....
 1198          old_premis_events = old_fsentry.get_premis_events()
 1199          old_premis_agents = old_fsentry.get_premis_agents()
 1200:         ss_agents = utils.get_ss_premis_agents()
 1201          replication_event = premis.get_replication_event(
 1202              self, replica_package, event_uuid=replication_event_uuid, agents=ss_agents
 ....
 1238          old_premis_agents = old_fsentry.get_premis_agents()
 1239          new_premis_events = list(set(old_premis_events + storage_effects.events))
 1240:         new_premis_agents = list(set(old_premis_agents + utils.get_ss_premis_agents()))
 1241          new_composition_level = old_composition_level
 1242          if storage_effects.composition_level_updater:
```

ross-spencer (changes requested):
Just one more thing! (spotted in testing)

storage_service/locations/models/gpg.py
cole: Exactly. Escape was only used on this line:

```
    detail = escape(
        "program=GPG; version={}; key={}".format(gpg_version, key_fingerprint)
    )
```
So previously it would have converted the bytestring to a unicode string, after the substitution of gpg_version and key_fingerprint. Now we're just using a unicode string directly. Both of the substitution vars should be already safe - version should be something like `1.4.16` and key_fingerprint comes from the database, so [Django should provide us with a unicode string](https://docs.djangoproject.com/en/1.8/ref/unicode/#models).

storage_service/common/utils.py
cole: 😱 oops, that's no good. Thanks for catching! Updated now.

storage_service/locations/models/package.py
cole: I think it makes sense to have a mix of args and kwargs here (args being required kwargs, being optional), but I've removed passing packages at all since they're not really required.

storage_service/locations/models/package.py
cole: Very good point! I've changed to pass just the UUID (or other bits of data) for all PREMIS functions.

storage_service/common/premis.py
cole: Yeah, before it was a mix of get and create, and I standardized on get, but I think you're right, that's the wrong one! I've renamed them all to create_.

storage_service/common/management/commands/import_aip.py
```diff
@@ -360,7 +360,7 @@ def compress(aip_model_inst, compression_algorithm):
     """Use the Package model's compress_package method to compress the AIP
     being imported, update the Package model's ``size`` attribute, retrieve
     PREMIS agents and event for the compression (using the package model's
-    ``get_premis_aip_compression_event`` method) and return a 3-tuple:
+    ``create_premis_aip_compression_event`` method) and return a 3-tuple:
```
ross-spencer, line 363: Bonus changes!! :fireworks:  Thanks Cole!

ross-spencer (commented):
Hi @cole, the code looks good, but I'm getting one last failure on reingest with this branch. It's showing up as a 500 internal server error in Archivematica and I'm not getting any additional logs in the Storage Service. If you've time can you take a look at that and double check if it is introduced here? 

I've rebased on top of 5fb80c554d3685dd3a81d182d64e70092ddcaa13 and tried that commit on its own as well. So I think the vanilla `qa/0.x` branch at the moment is okay.

cole:
It seems that during reingest processing the PREMIS 3 namespace was causing errors. I've updated the pointer file reingest to use metsrw as well, as I'd missed that and it was resulting in an odd mix of PREMIS 2 and 3.

There is a change to the resulting PREMIS output though, due to metsrw's behaviour — instead of being replaced, the old techMD containing the premis:object is now superseded ([example output](https://gist.github.com/cole/3578516be27039716043409cc13ad19e)).

If we think that's a problem I think I know how to work around it — but it may actually be better? I guess that's worth some discussion. I'm not available tomorrow, and I'm aware this is blocking other work, so @ross-spencer if you need to press on I'd suggest we can accept this PR as is with the understanding that revisions can happen in a separate one (provided reingest is now working for you of course!). Otherwise I can pick it up on the 13th.

ross-spencer:
Thanks @cole merging now seems a good idea as we will be trying to bring together a handful of storage service changes. I've logged the issue with the superseded section [here](https://github.com/archivematica/Issues/issues/738), and noticed one possible issue with it if we keep this functionality.

ross-spencer (approved):
:+1: LGTM. Thanks Cole. Will rebase and merge!

PR #470, opened 2019-06-14
Use server-side processing for packages table

replaceafill:
This PR changes the implementation of the two tables that display packages in the Storage Service. It adds a new view (`/packages_ajax`) that uses the [server-side processing options](http://legacy.datatables.net/usage/server-side) of the DataTable library to retrieve small batches of packages through AJAX requests instead of the full list of packages.

It also addresses the following changes proposed by Sara and the rest of the team in https://github.com/archivematica/Issues/issues/676#issuecomment-497077314 and below:

* Deleted the Description column which was set always to None. I think this is caused by not having the `description` field included in `storage_service.locations.api.PackageResource.Meta.fields`
* Changed the link in the Current location column to allow downloading the package (similar to the Download action)
* Moved the Pointer File link under the Actions column
* Removed the Replicas column and retitled the "Is Replica Of" column
* Most of the columns are now sortable and searchable and sorting by size uses the actual package size now
* Made the Packages tab and Location views fluid (expand the whole viewport width)

Connected to https://github.com/archivematica/Issues/issues/676
Connected to https://github.com/archivematica/Issues/issues/678

### DIFF
--- a/storage_service/locations/datatable_utils.py
+++ b/storage_service/locations/datatable_utils.py
@@ -1,0 +1,160 @@
+"""DataTable utils
+
+Implements the server-side processing options needed by the jQuery
+DataTables v1.9 library used in the packages tables of the storage
+service (in the Packages and the Location detail views).
+
+The request and response details are documented in:
+http://legacy.datatables.net/usage/server-side
+"""
+
+import os
+
+from django.db.models import Q
+from django.utils import timezone
+
+from .models import Package
+
+
+class DataTable(object):
+    """This class parses the parameters sent by the JS client code and
+    builds a filtered and sorted list of `Package` objects.
+    """
+
+    # number of rows to display on a single page when using pagination
+    DEFAULT_DISPLAY_LENGTH = 10
+
+    def __init__(self, query_dict):
+        self.total_records = Package.objects.count()
+        self.params = self.parse_datatable_parameters(query_dict)
+        self.echo = self.params["echo"]
+        search_filter = Q()
+        if self.params["search"]:
+            search = self.params["search"]
+            # remove any leading slashes so we can search in relative paths
+            search_as_path = search.lstrip(os.path.sep)
+            search_filter |= (
+                Q(uuid__icontains=search)
+                | Q(description__icontains=search)
+                | Q(origin_pipeline__uuid__icontains=search)
+                | Q(origin_pipeline__description__icontains=search)
+                | Q(current_location__relative_path__icontains=search_as_path)
+                | Q(current_location__space__path__icontains=search_as_path)
+                | Q(current_path__icontains=search_as_path)
+                | Q(package_type__icontains=search)
+                | Q(status__icontains=search)
+            )
+        queryset = Package.objects.filter(search_filter)
+        self.total_display_records = queryset.count()
+        self.packages = self.get_packages(queryset)
+
+    def _get_int_parameter(self, query_dict, param, default=0):
+        """Get an integer parameter from the request QueryDict.
+
+        Return `default` on failure
+        """
+        try:
+            return int(query_dict.get(param))
+        except (ValueError, TypeError):
+            return default
+
+    def sorting_column(self, query_dict):
+        """Return a dictionary with the column index and sorting direction of
+        the column used for sorting the table.
+
+        The DataTable library can sort by multiple columns but this
+        implementation is very simple and limited to a single column sort.
+        """
+        result = {}
+        sorting_columns_count = self._get_int_parameter(query_dict, "iSortingCols")
+        if sorting_columns_count == 1:
+            sorting_column_index = self._get_int_parameter(
+                query_dict, "iSortCol_0", None
+            )
+            if sorting_column_index is not None:
+                is_sortable = (
+                    query_dict.get("bSortable_{}".format(sorting_column_index))
+                    == "true"
+                )
+                if is_sortable:
+                    sort_direction = query_dict.get("sSortDir_0", "asc")
+                    result = {
+                        "index": sorting_column_index,
+                        "direction": sort_direction,
+                    }
+        return result
+
+    def parse_datatable_parameters(self, query_dict):
+        return {
+            # query for text based search
+            "search": query_dict.get("sSearch", ""),
+            # options for limiting result display
+            "display_start": self._get_int_parameter(query_dict, "iDisplayStart"),
+            "display_length": self._get_int_parameter(
+                query_dict, "iDisplayLength", default=self.DEFAULT_DISPLAY_LENGTH
+            ),
+            # sorting information
+            "sorting_column": self.sorting_column(query_dict),
+            # parameter sent by the client that the server needs
+            # to return back in order to identify each single rendering
+            "echo": self._get_int_parameter(query_dict, "sEcho", default=-1),
+        }
+
+    def sort(self, queryset):
+        sorting_column = self.params["sorting_column"]
+        if not sorting_column or sorting_column.get("index") is None:
+            return queryset
+        # the columns represented by these indexes can be sorted
+        # through the queryset using the order_by method
+        ORDER_BY_MAPPING = {
+            0: "uuid",
+            1: "origin_pipeline__description",
+            3: "size",
+            5: "replicated_package__uuid",
+        }
+        # these columns instead can't be sorted directly with the
+        # queryset so the queryset needs to be converted into a list
+        # and then sorted using helper methods
+        SORT_KEY_HELPERS_MAPPING = {
+            2: self.sort_by_full_path_key,
+            4: self.sort_by_package_type_key,
+            6: self.sort_by_status_key,
+            7: self.sort_by_fixity_date_key,
+            8: self.sort_by_fixity_status_key,
+        }
+        sort_descending = sorting_column.get("direction") == "desc"
+        if sorting_column["index"] in ORDER_BY_MAPPING:
+            field = ORDER_BY_MAPPING[sorting_column["index"]]
+            if sort_descending:
+                field = "-{}".format(field)
+            return queryset.order_by(field)
+        elif sorting_column["index"] in SORT_KEY_HELPERS_MAPPING:
+            key = SORT_KEY_HELPERS_MAPPING[sorting_column["index"]]
+            return sorted(list(queryset), key=key, reverse=sort_descending)
+        else:
+            return queryset
+
+    def get_packages(self, queryset):
+        result = self.sort(queryset)
+        display_start = self.params["display_start"]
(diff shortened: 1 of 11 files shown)


storage_service/common/datatable_utils.py
```diff
@@ -0,0 +1,159 @@
+"""DataTable utils
```
cole: It's really a matter of opinion, but I'd advocate for putting this file in locations, just on the idea that locations should import from common, but not the other way around.

storage_service/common/datatable_utils.py
```diff
@@ -0,0 +1,159 @@
...
+                | Q(origin_pipeline__uuid__icontains=search)
+                | Q(origin_pipeline__description__icontains=search)
+                | Q(current_location__relative_path__icontains=search_as_path)
+                | Q(current_location__space__path__icontains=search_as_path)
+                | Q(current_path__icontains=search_as_path)
+                | Q(package_type__icontains=search)
+                | Q(status__icontains=search)
+            )
+        else:
+            queryset = Package.objects.filter()
```
cole: The results are the same but maybe `Package.objects.all()` is clearer?

storage_service/common/datatable_utils.py
```diff
@@ -0,0 +1,159 @@
...
+        }
+        sort_descending = sorting_column.get("direction") == "desc"
+        if sorting_column["index"] in ORDER_BY_MAPPING:
+            field = ORDER_BY_MAPPING[sorting_column["index"]]
+            if sort_descending:
+                field = "-{}".format(field)
+            return queryset.order_by(field)
+        elif sorting_column["index"] in SORT_KEY_HELPERS_MAPPING:
+            key = SORT_KEY_HELPERS_MAPPING[sorting_column["index"]]
+            queryset = sorted(list(queryset), key=key, reverse=sort_descending)
```
cole: This case needs a return value (or just move the return below outside of the else).

storage_service/locations/views.py
```diff
@@ -69,10 +72,41 @@ def package_list(request):
...
     return render(request, "locations/package_list.html", context)
 
 
+def package_list_ajax(request):
+    datatable = datatable_utils.build_table(request.GET)
+    data = []
+    csrf_token = get_token(request)
+    for package in datatable.packages:
+        data.append(
+            get_template("snippets/package_row.html")
```
cole, line 87: Why not return the table data as JSON instead of the rendered row? It should be faster over the wire, and it's more flexible if we want to change the client side in future.

storage_service/templates/snippets/packages_table.html
```diff
@@ -1,120 +1,22 @@
...
-                  <button class="btn" data-dismiss="modal" aria-hidden="true">{% trans "Close" %}</button>
-                  <button class="btn btn-danger" type="submit">{% trans "Delete" %}</button>
-                </div>
-              </div>
-            </form>
-          {% endif %}
-
-        </td>
-      </tr>
+      {% include "snippets/package_row.html" %}
```
cole: It looks like on initial load we are rendering the full table, then immediately updating it via datatables — can we skip the initial render and display a "Loading..." row? Or, defer the datatable ajax request until the table needs updating (I couldn't find an easy option for that, but there may be one).

storage_service/common/datatable_utils.py
```diff
@@ -0,0 +1,159 @@
...
+    def sort_by_fixity_date_key(self, package):
+        # avoid comparing datetimes with None
+        return package.latest_fixity_check_datetime or timezone.now()
+
+    def sort_by_fixity_status_key(self, package):
+        return package.latest_fixity_check_result
+
+
+def build_table(query_dict):
+    return DataTable(query_dict)
```
cole: I think you can just skip this function.

storage_service/common/datatable_utils.py
```diff
@@ -0,0 +1,159 @@
...
+            key = SORT_KEY_HELPERS_MAPPING[sorting_column["index"]]
+            queryset = sorted(list(queryset), key=key, reverse=sort_descending)
+        else:
+            return queryset
+
+    def get_packages(self, queryset):
+        result = self.sort(queryset)
+        display_start = self.params["display_start"]
+        display_length = self.params["display_length"]
+        if display_start and display_length:
```
cole: `display_start` is 0 on the first page, so this will return all results for that page 😁. I think we need to check for `is not None`.

cole (changes requested):
Nice! This looks overall very good, a few questions and small bugs noted.

It's kind of a shame we're in an odd place between tastypie and adopting DRF (which would possibly give us a lot of this work for free via [django-rest-framework-datatables](https://github.com/izimobil/django-rest-framework-datatables). For consistency it might make more sense to implement the AJAX view via tastypie...but both tastypie and datatables are quite opinionated in the way they structure JSON responses, so I think that's probably a lot more work than it's worth.

storage_service/common/datatable_utils.py
replaceafill: Ah! At some point I thought about using `get_model` to avoid the import from locations, but I wasn't sure about what app label or AppConfig to use and couldn't find much (or any) use of it in the codebase. Maybe it's not a good pattern? I'd appreciate your opinion on that?

storage_service/common/datatable_utils.py
replaceafill: I was wondering if all those `Q` objects could be grouped together and passed as `kwargs` like `filter` expects. Something along the lines of:

```
kwargs = {}
if self.params["search"]:
     # add all the Qs to kwargs here
queryset = Package.objects.filter(**kwargs)
```

I can live with the `all` branch. Just curious.

storage_service/common/datatable_utils.py
replaceafill: :man_facepalming: Thanks!

storage_service/common/datatable_utils.py
replaceafill: Looks like I need to improve the tests too!

storage_service/common/datatable_utils.py
replaceafill: :+1:

storage_service/locations/views.py
replaceafill: I tried it at the beginning, but some rows have a modal form that made the JS a bit complex. The modal repetition is inefficient too and I think instead we could have a single modal in the body of page and just populate and display it when the user clicks on the action link.

I also agree with your point about flexibility. Will see what I can do.

storage_service/common/datatable_utils.py
cole: I think that's a good pattern in some cases (mostly to avoid circular imports) but not necessary here since the datatable implementation is package specific. I haven't seen much of that in the codebase either 🙂

storage_service/common/datatable_utils.py
cole: I think you can do something like:
```
search_filter = Q()
if self.params["search"]:
    search_filter = search_filter | Q(uuid__icontains=search) | ...
queryset = Package.objects.filter(search_filter)
```

storage_service/locations/views.py
cole: It may be too complicated/too late for this PR... if so that's fine!

storage_service/templates/snippets/packages_table.html
replaceafill: Ah yes! I think it's [iDeferLoading](http://legacy.datatables.net/usage/options#iDeferLoading)

storage_service/common/datatable_utils.py
replaceafill: Thank you for explaining. Makes sense. I moved the module to the `locations` package.

storage_service/common/datatable_utils.py
replaceafill: :+1: Also used Jesus suggestion about `search_filter |= ...`

storage_service/locations/views.py
replaceafill: Thanks. I decided to leave it like this for now. I plan to update the issue with things I didn't handle in this implementation. I'll make a note about this.

storage_service/templates/snippets/packages_table.html
replaceafill: `iDeferLoading` presented the problem that it needs the total number of records for the other navigations options to work even if the datatable hasn't been built. So as you suggested, instead I set a `Loading data from server` message and let the table to handle the initial rendering. I also verified the behavior in the Network tab of the devtools. Thank you for catching it!

storage_service/common/datatable_utils.py
replaceafill: In https://github.com/artefactual/archivematica-storage-service/pull/470/commits/d291a7b18cd7222b327d34c7098cdaa23852c822 I set a default number of rows to display and changed the condition to verify if some result exists.

storage_service/locations/datatable_utils.py
```diff
@@ -0,0 +1,160 @@
...
+            key = SORT_KEY_HELPERS_MAPPING[sorting_column["index"]]
+            return sorted(list(queryset), key=key, reverse=sort_descending)
+        else:
+            return queryset
+
+    def get_packages(self, queryset):
+        result = self.sort(queryset)
+        display_start = self.params["display_start"]
+        display_length = self.params["display_length"]
+        if result:
```
cole: `result` is a queryset here, so this will evaluate it (`SELECT *`). I would use `result.count()`, since we don't need all the queryset results (just the slice below). This is like the number one queryset gotcha 🙂

cole (changes requested):
Looks good! Just one small thing.

storage_service/locations/datatable_utils.py
replaceafill: Ah! Thanks for the pointer. I just found in the docs about [when querysets are evaluated](https://docs.djangoproject.com/en/1.8/ref/models/querysets/#when-querysets-are-evaluated). Then I assume this also applies to templates, right? If that's the case maybe we should also update the [package_list template](https://github.com/artefactual/archivematica-storage-service/blob/17b1fce2993c07fa11d9c11e8b1ac03c5c7535e4/storage_service/templates/locations/package_list.html#L15) and/or the [view](https://github.com/artefactual/archivematica-storage-service/blob/17b1fce2993c07fa11d9c11e8b1ac03c5c7535e4/storage_service/locations/views.py#L68).

cole: Doh! Totally missed that. Yes, exactly. It looks like it's not actually evaluated in the view right now, but it is in the template. I'd pass a `package_count` through in the template context.

cole (approved):

PR #472, opened 2019-06-21
Remove extra superseded techMD from reingested pointer file

cole:
Relates to: https://github.com/archivematica/Issues/issues/738

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -2634,9 +2634,6 @@ class Package(models.Model):
             # what AM uses
             checksum = utils.generate_checksum(path, "sha512")
 
-        aip = mets.get_file(type="Archival Information Package")
-        aip.path = self.full_path
-
         transform_types = set(
             [transform.get("type") for transform in aip.transform_files]
         )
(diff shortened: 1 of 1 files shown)


ross-spencer (approved):
I see no concerns with the code, and this works, so LGTM. :+1: 

If I understand correctly, we'll have two techMDs at this point. The latter one is appended to the amdSec, and so to remove the unnecessary techMD, we need to find the first instance in the amdSec, we record its index, and then remove it. In the example where were use superseded and current this would have been the superseded techMD.

Hopefully that's right :crossed_fingers: 

This has been useful to see and work through, I think I have spotted some more reingest issues (not related to this commit) and will follow up with a ticket for those later today/tomorrow.

ross-spencer:
Hi @cole I wonder if you can hold tight on merging this. I was testing again today, and I think we'll inadvertently break reingest in the use case where:

1. User creates an AIP using 7z (uncompressed as an example).
2. User changes the processing config to Gzip (Parallel bz might be the same).
3. User reingests.

I think in that case the final pointer file will still only have two techMDs, but one for each composition level:
```xml
<premis:compositionLevel>1</premis:compositionLevel>
<premis:compositionLevel>2</premis:compositionLevel>
```
* Example: https://gist.github.com/ross-spencer/36f648ad92e1ad6c097c5247f1c8f436

In which case we wouldn't want to delete either. 

I don't know if there is an easy heuristic. Let me know what you think. We may need to discuss more as a team. 

One alternative I think would be worth considering is removing nearly-all reingest pointer file logic, destroying the pointer file each reingest, and creating the ability to create a new pointer file from scratch once we've recovered all the past-details we need (which seems to just be information about encryption/compression/possibly replication?). And seems to be used to accurately extract what we've got. We don't need to specifically preserve anything like you've noted in the issue related to this one. If we did this we wouldn't need functions like `_update_pointer_file`. We could create a cleaner mechanism. 

Anyway, thinking too far ahead, and it feels like such a broad change we'd need good discussion if it is at all a valid option. (We have an [epic for reingest](https://github.com/archivematica/Issues/issues/751) in Codetree).

cole:
@ross-spencer thanks for catching! I think we can handle both cases pretty easily — please let me know if you see any issues with the changes in a63c2fb.

ross-spencer:
Thanks @cole this looks to do the trick :+1:

PR #474, opened 2019-06-25
Improve the package deletion mechanism around S3

ross-spencer:
**NB.** This is a direct copy of https://github.com/artefactual/archivematica-storage-service/pull/463 which was accidentally closed before merging. Spotted by @jraddaoui. 

A couple of small issues were effecting our ability to effectively
delete S3 packages. Those are resolved here.

The commit history for these changes are as follows:

* Test generator provided by Boto3

  Tests the generator provided by Boto3. We expect there to be
  something to iterate over if a deletion has been requested. If
  there is nothing to iterate over a StorageException is returned and
  the reason returned to the package.py.

* Manage exceptions from deletion requests

  Except Exception has the potential to suck up a number of other
  errors raised by Python so look specifically at what might be
  returned from various storage adapters.

  Additionally, return immediately if there has been a problem with a
  deletion request instead of letting the code continue to delete the
  pointer file for a package and set its status as deleted when there
  is likely a higer probability that the package hasn't been deleted.

* Remove leading slashes from s3 deletion requests

  When an S3 deletion request is made and the space is configured
  incorrectly, i.e. using a leading slash for a Path, then it will not
  interfere with the deletion. The slash is removed so that the
  package can be found using Boto3 and then deleted.

* Add logging

  Raising a storage service exception isn't guaranteed to appear as a
  log output for the administrator. Ensure that happens here.

We correct some of the logging elsewhere in the code along the way as
well.

### DIFF
--- a/storage_service/locations/api/resources.py
+++ b/storage_service/locations/api/resources.py
@@ -5,6 +5,7 @@
 import json
 import logging
 import os
+import pprint
 import re
 import shutil
 import urllib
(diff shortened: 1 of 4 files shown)


jraddaoui (approved):
Thanks @ross-spencer!!

PR #477, opened 2019-07-12
Expose LDAP TLS options via env variables

cole:
It seems that most users will need to set some TLS options (certfiles, etc), so exposing those via env vars is helpful.

Connected to: archivematica/Issues#680

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -188,6 +188,138 @@ This is the current list of strings supported:
     - **Type:** `string`
     - **Default:** `archivematica-storage-service`
 
+### LDAP-specific environment variables
+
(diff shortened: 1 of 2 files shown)


sevein:
Looks good.

We should be adding the envs to [install/README.md](https://github.com/artefactual/archivematica-storage-service/blob/qa/0.x/install/README.md) even if it's just a comment referring to the settings file? Same in [dashboard/install/README.md](https://github.com/artefactual/archivematica/blob/qa/1.x/src/dashboard/install/README.md) at some point?

cole:
@sevein good point! I have updated the SS readme, and filed https://github.com/archivematica/Issues/issues/798 for adding documentation to the dashboard.

sevein (approved):
Please update the TOC in install/README.md if you can!

PR #479, opened 2019-07-23
Include replicas in package search

replaceafill:
This PR allows to search for a package UUID and get its replicas returned in the packages table. Searching for a replica UUID will also return the replicated package.

Connected to https://github.com/archivematica/Issues/issues/801


ross-spencer (approved):
This is pretty cool code @replaceafill. Looks great to me. :+1:

PR #485, opened 2019-10-07
arkivum: get email sender from SS settings

hakamine:
Connected to https://github.com/archivematica/Issues/issues/921 and https://github.com/archivematica/Issues/issues/1298.


sevein (approved):
Sorry for taking so long!

PR #486, opened 2019-10-10
import_aip: add config parameters

hakamine:
- Add an argument to be able to specify a directory where the AIP is to be decompressed
- Add an argument to be able to skip checking of AIP existence in the SS database

Connected to https://github.com/archivematica/Issues/issues/940

### DIFF
--- a/storage_service/common/management/commands/import_aip.py
+++ b/storage_service/common/management/commands/import_aip.py
@@ -123,6 +123,15 @@ class Command(BaseCommand):
             default=DEFAULT_UNIX_OWNER,
         )
 
+        parser.add_argument(
+            "--tmp-dir",
+            help="Temporary directory for processing, passed as dir parameter"
+            " to tempfile.mkdtemp(), e.g,'/var/archivematica/sharedDirectory/tmp'."
+            " Default: None ",
+            default=None,
+            required=False,
+        )
+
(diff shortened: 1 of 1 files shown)


ross-spencer (approved):
Really nice improvement @hakamine thanks for this. LGTM! :+1: 

For testers I used a blank install on docerk-compose, and ran the command:
```
sudo make manage-ss ARG='import_aip /home/import-aip/my-aip.7z \
 --decompress-source --compression-algorithm="7z with bzip" \
 --tmp-dir /var/archivematica/sharedDirectory/ --force'
```
Where `--tmp-dir` is a known location to compose, and `/home/import-aip/my-aip.7z` was an AIP I had previously downloaded to my transfer source. 

I ran it a second time with the same args and it re-uploaded without error. A third time without --force, and without the tmp dir. Again, the functionality worked as anticipated - in the last case, asking if I wanted to replace the AIP. In the first two cases I could see the tmp dir being used when I monitored it.

ross-spencer:
oh @hakamine I just noticed the failing build, I think you might need to force push after a rebase to make it work and then it can be merged. I can't see any flake8 or black errors otherwise.

hakamine:
@ross-spencer thank you very much for the CR. Travis tests were green after rebasing. Will merge.

PR #488, opened 2019-10-10
Fix for selecting region us-east-1 with S3

fitnycdigitalinitiatives:
Addresses [archivematica/Issues#922](https://github.com/archivematica/Issues/issues/922)

Before creating a bucket checks to see if the region is us-east-1. If true, then creates bucket without LocationConstraint. Otherwise, creates bucket with a LocationConstraint supplied region.

### DIFF
--- a/storage_service/locations/models/s3.py
+++ b/storage_service/locations/models/s3.py
@@ -101,10 +101,8 @@ class S3(models.Model):
                 raise StorageException(err)
             LOGGER.info("Creating S3 bucket '%s'", self._bucket_name())
             # LocationConstraint cannot be specified if it us-east-1 because it is the default, see: https://github.com/boto/boto3/issues/125
-            if self.region == 'us-east-1':
-                self.client.create_bucket(
-                    Bucket=self._bucket_name()
-                )
+            if self.region == "us-east-1":
+                self.client.create_bucket(Bucket=self._bucket_name())
             else:
                 self.client.create_bucket(
                     Bucket=self._bucket_name(),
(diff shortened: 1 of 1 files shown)


storage_service/locations/models/s3.py
```diff
@@ -100,10 +100,16 @@ def _ensure_bucket_exists(self):
...
                 raise StorageException(err)
             LOGGER.info("Creating S3 bucket '%s'", self._bucket_name())
-            self.client.create_bucket(
-                Bucket=self._bucket_name(),
-                CreateBucketConfiguration={"LocationConstraint": self.region},
-            )
+            # LocationConstraint cannot be specified if it us-east-1 because it is the default, see: https://github.com/boto/boto3/issues/125
+            if self.region == 'us-east-1':
+                self.client.create_bucket(
+                    Bucket=self._bucket_name()
```
ross-spencer: This looks good, can you run `black` against the changed files, I think it will convert this line to something like:
```
+            if self.region == "us-east-1":
+                self.client.create_bucket(Bucket=self._bucket_name())
```
And then I think the build will pass okay! 

Black is found here: https://github.com/psf/black

ross-spencer (commented):
This is nice Joseph. Thanks for taking the time to make this submission. I've marked the original ticket with a `1.11` triage label, and so I'll ask the product owner about putting them in with the next release. Once you've run the black formatter, can you rebase the code so it remains a single commit, and then I'll approve this and it will be ready to go!

storage_service/locations/models/s3.py
fitnycdigitalinitiatives: I totally messed up the rebasing process. I'm just going to resubmit the pull request.

PR #489, opened 2019-10-14
Correct user admin editing behavior

ross-spencer:
1. Prevents non-superusers editing permissions
2. Prevents the only-superuser from locking themselves out of editing permissions.

Connected to archivematica/issues#948

### DIFF
--- a/storage_service/administration/forms.py
+++ b/storage_service/administration/forms.py
@@ -222,9 +222,31 @@ class UserChangeForm(auth.forms.UserChangeForm):
     """ Modifys an existing user.  Inherits from django's UserChangeForm. """
 
     def __init__(self, *args, **kwargs):
+        current_user = kwargs.pop("current_user", None)
+        self.user_being_edited = kwargs["instance"]
+        self.superusers = auth.get_user_model().objects.filter(is_superuser=True)
         super(UserChangeForm, self).__init__(*args, **kwargs)
         self.fields["is_superuser"].label = _("Administrator?")
+        if not (current_user and current_user.is_superuser):
+            # If current user is not super, do not permit editing of that.
+            del self.fields["is_superuser"]
+        elif self.superusers.count() == 1 and current_user == self.user_being_edited:
+            # Provide some indication that this is undesirable.
+            self.fields["is_superuser"].widget.attrs["readonly"] = True
         del self.fields["password"]
+
+    def clean(self):
+        """Validate the form to protect against potential user errors."""
+        if self.superusers.count() > 1:
+            return self.cleaned_data
+        try:
+            # Protect field from being reverted if only one superuser.
+            if self.user_being_edited.is_superuser:
+                self.cleaned_data["is_superuser"] = True
+        except KeyError:
+            # Field isn't being modified, nothing to do.
+            pass
+        return self.cleaned_data
 
     class Meta:
         model = auth.get_user_model()
--- a/storage_service/administration/views.py
+++ b/storage_service/administration/views.py
@@ -80,7 +80,9 @@ def user_edit(request, id):
 
     action = _("Edit User")
     edit_user = get_object_or_404(get_user_model(), id=id)
-    user_form = settings_forms.UserChangeForm(request.POST or None, instance=edit_user)
+    user_form = settings_forms.UserChangeForm(
+        request.POST or None, instance=edit_user, current_user=request.user
+    )
     password_form = SetPasswordForm(data=request.POST or None, user=edit_user)
     if "user" in request.POST and user_form.is_valid():
         user_form.save()


storage_service/administration/forms.py
```diff
@@ -222,8 +222,17 @@ class UserChangeForm(auth.forms.UserChangeForm):
...
         super(UserChangeForm, self).__init__(*args, **kwargs)
+        currently_editing = getattr(kwargs["instance"], "username")
+        superusers = auth.get_user_model().objects.filter(is_superuser=True)
         self.fields["is_superuser"].label = _("Administrator?")
+        if len(superusers) >= 1:
+            if (
+                not getattr(kwargs["instance"], "is_superuser")
+                and self.current_user == currently_editing
+            ):
+                del self.fields["is_superuser"]
```
cole: I think this is the right approach; the form is the right place to make this change. It's kind of too bad there isn't a better solution for this in Django but none of the permissions logic gets as granular as form fields. 

I would simplify it though so that only superusers can change the attribute, they can change it for anyone, and we don't care if there are any existing superusers or not. 

something like:
```
current_user = kwargs.pop("current_user", None)
...
if not (current_user and current_user.is_superuser):
    del self.fields["is_superuser"]
```

I'm guessing the check for existing superusers is to prevent being locked out of the system? Django provides a management command (`createsuperuser`) for that case. It's not very user friendly but it seems like it would be ok in the event someone locks out all accounts. Alternatively I think more secure approach would be to disable the superuser field on the last super user. In that case we can disable the form field:

```
superusers = auth.get_user_model().objects.filter(is_superuser=True)
...
if not (current_user and current_user.is_superuser):
    del self.fields["is_superuser"]
elif superusers.count() == 1:
    self.fields['is_superuser'].widget.attrs['readonly']
```

storage_service/administration/forms.py
```diff
@@ -222,8 +222,17 @@ class UserChangeForm(auth.forms.UserChangeForm):
     """ Modifys an existing user.  Inherits from django's UserChangeForm. """
 
     def __init__(self, *args, **kwargs):
+        self.current_user = kwargs.pop("current_user", None).username
         super(UserChangeForm, self).__init__(*args, **kwargs)
+        currently_editing = getattr(kwargs["instance"], "username")
+        superusers = auth.get_user_model().objects.filter(is_superuser=True)
         self.fields["is_superuser"].label = _("Administrator?")
+        if len(superusers) >= 1:
```
cole: Minor note because this is a super common Django ORM gotcha/thing: `len(superusers)` selects all user data for super users, builds a list, and checks the list length, while `superusers.count()` will do a `SELECT COUNT...`.  Not really an issue here, moreso if the number of items in the list is likely to be large.

storage_service/administration/forms.py
```diff
@@ -222,8 +222,17 @@ class UserChangeForm(auth.forms.UserChangeForm):
...
+            if (
+                not getattr(kwargs["instance"], "is_superuser")
+                and self.current_user == currently_editing
```
cole: I don't think we need this comparison (see note on review) but also a common Django thing here: you can just compare user objects with `==`, so `kwargs.pop("current_user", None) == kwargs["instance"]` (as long as they're not both None). It's mostly a matter of preference, but that method will compare the primary keys.

cole (changes requested):
I think this is the right approach; the form is the right place to make this change. It's kind of too bad there isn't a better solution for this in Django but none of the permissions logic gets as granular as form fields. 

I would simplify it though so that only superusers can change the attribute, they can change it for anyone, and we don't care if there are any existing superusers or not. 

something like:
```
current_user = kwargs.pop("current_user", None)
...
if not (current_user and current_user.is_superuser):
    del self.fields["is_superuser"]
```

I'm guessing the check for existing superusers is to prevent being locked out of the system? Django provides a management command (`createsuperuser`) for that case. It's not very user friendly but it seems like it would be ok in the event someone locks out all accounts. Alternatively I think more secure approach would be to disable toggling the superuser field on the last super user. In that case we can disable the form field (though that may also need some explanation to users):

```
superusers = auth.get_user_model().objects.filter(is_superuser=True)
...
if not (current_user and current_user.is_superuser):
    del self.fields["is_superuser"]
elif superusers.count() == 1:
    self.fields['is_superuser'].widget.attrs['readonly'] = True
```

ross-spencer:
Hi @cole thank you for the tips. I've come up with a bit of a Frankenstein's monster of what was there before perhaps? It provides some feedback and protection from the user, and I *think* the logic holds up for the end-user too. Have thrown it into 'read-for-review' and await the verdict! :slightly_smiling_face:

storage_service/administration/forms.py
```diff
@@ -222,10 +222,30 @@ class UserChangeForm(auth.forms.UserChangeForm):
...
+        current_user = kwargs.pop("current_user", None)
+        user_being_edited = kwargs["instance"]
+        self.superusers = auth.get_user_model().objects.filter(is_superuser=True)
         super(UserChangeForm, self).__init__(*args, **kwargs)
         self.fields["is_superuser"].label = _("Administrator?")
+        if not (current_user and current_user.is_superuser):
+            # If current user is not super, do not permit editing of that.
+            del self.fields["is_superuser"]
+        elif self.superusers.count() == 1 and not (current_user != user_being_edited):
+            # Provide some indication that this is undesirable."""
```
cole: Trailing `"""`

storage_service/administration/forms.py
```diff
@@ -222,10 +222,30 @@ class UserChangeForm(auth.forms.UserChangeForm):
...
 
+    def clean(self):
+        """Validate the form to protect against potential user errors."""
+        if self.superusers.count() > 1:
+            return self.cleaned_data
+        try:
+            # Protect field from being reverted if only one superuser.
+            self.cleaned_data["is_superuser"] = True
+        except KeyError:
+            """Field isn't being modified, nothing to do."""
```
cole: It's v minor but I would say this is a place for a pass and a comment. Strings are code, even if they are no-ops, comments aren't (docstrings are an exception). [This SO answer is pretty good on that topic](https://stackoverflow.com/questions/708649/python-comments-vs-strings#answer-708674).

storage_service/administration/forms.py
```diff
@@ -222,10 +222,30 @@ class UserChangeForm(auth.forms.UserChangeForm):
...
+            self.fields["is_superuser"].widget.attrs["readonly"] = True
         del self.fields["password"]
 
+    def clean(self):
+        """Validate the form to protect against potential user errors."""
+        if self.superusers.count() > 1:
+            return self.cleaned_data
+        try:
+            # Protect field from being reverted if only one superuser.
+            self.cleaned_data["is_superuser"] = True
```
cole: What about raising a [ValidationError](https://docs.djangoproject.com/en/1.8/ref/forms/validation/#raising-validationerror) here instead? That should give the user some feedback as to what's happening.

I think if you go that route it actually removes the need for the readonly widget I suggested before (sorry!).

storage_service/administration/forms.py
```diff
@@ -222,10 +222,30 @@ class UserChangeForm(auth.forms.UserChangeForm):
...
     def __init__(self, *args, **kwargs):
+        current_user = kwargs.pop("current_user", None)
+        user_being_edited = kwargs["instance"]
+        self.superusers = auth.get_user_model().objects.filter(is_superuser=True)
         super(UserChangeForm, self).__init__(*args, **kwargs)
         self.fields["is_superuser"].label = _("Administrator?")
+        if not (current_user and current_user.is_superuser):
+            # If current user is not super, do not permit editing of that.
+            del self.fields["is_superuser"]
+        elif self.superusers.count() == 1 and not (current_user != user_being_edited):
```
cole: `not (current_user != user_being_edited)` can be written without the double negative as `current_user == user_being_edited`.

cole (approved):
Approved; this PR is a big improvement on the current behaviour. I've left some comments though, up to you if you want to make changes.

storage_service/administration/forms.py
ross-spencer: I think this is a good suggestion, but have left it as is for now. The validation error looks a bit messy as it will complain password fields need changing too, and I think possibly protects the form from being changed in other ways. 

**NB.** for transparency as I merge, I put in one additional check here after the try to make sure this only impacts the user being edited if they are already a super user as I spotted all users would otherwise get made super if editing.

PR #490, opened 2019-10-16
 Increase default workers to 3 

cole:
Connects to: https://github.com/archivematica/Issues/issues/944

It's hard to determine an ideal number of workers given the number of other processes on the system. It seems clear though that 2-4 is a good minimum, so start at 3.

Also increases the default SQLite timeout to try to mitigate concurrent write errors.

I'm not sure if this should be merged as it may cause more trouble than it's worth due to the use of SQLite.

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -147,12 +147,14 @@ This is the current list of strings supported:
     - **Default:** `127.0.0.1:8001`
 
 - **`SS_GUNICORN_WORKERS`**:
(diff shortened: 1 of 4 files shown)


sevein:
Not merging this to avoid potential issues with SQLite. But I'll make a recommendation in our distribution channels to increase the default number of workers where MySQL is used. Our [scaling docs](https://www.archivematica.org/en/docs/archivematica-1.11/admin-manual/installation-setup/customization/scaling-archivematica/#optimising-the-number-of-storage-service-workers) also mention it.

PR #459, opened 2019-05-28
Improvements to s3 plugin

helenst:
Addresses https://github.com/archivematica/Issues/issues/712 - also related to https://github.com/archivematica/Issues/issues/639

I've also added some tests for the browse feature, and added an optional bucket name field (which makes it easier to set up policies etc. when the system is being automatically deployed, rather than use the space id)

I've bundled these few small fixes in together but please let me know if there's anything you'd like separately?

Connected to archivematica/Issues#558
Connected to archivematica/Issues#975
Connected to archivematica/Issues#712

PR #497, opened 2019-11-14
S3 plugin: allow named bucket to be specified

helenst:
Splitting out https://github.com/artefactual/archivematica-storage-service/pull/459 into separate PRs - this addresses https://github.com/archivematica/Issues/issues/558 and adds a bucket name field to the S3 storage plugin so that a pre-existing named bucket can be used, rather than auto creating one to match the space id.

### DIFF
--- a/requirements/test.in
+++ b/requirements/test.in
@@ -10,3 +10,4 @@ tox
 vcrpy>=1.0.0
 ipdb
 pip-tools==3.7.0
+moto==1.3.8
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -57,6 +57,7 @@ metsrw==0.3.13
 mock==3.0.5               # via pytest-mock, vcrpy
 monotonic==1.5
 more-itertools==5.0.0     # via pytest, zipp
+moto==1.3.8
 msgpack==0.6.2
 mysqlclient==1.4.6
 ndg-httpsclient==0.4.2
--- a/storage_service/locations/constants.py
+++ b/storage_service/locations/constants.py
@@ -100,6 +100,12 @@ PROTOCOL = {
     models.Space.S3: {
         "model": models.S3,
         "form": forms.S3Form,
-        "fields": ["endpoint_url", "access_key_id", "secret_access_key", "region"],
+        "fields": [
+            "endpoint_url",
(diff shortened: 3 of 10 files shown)


replaceafill (approved):
LGTM :+1: and works great. I just replicated an AIP in an existing S3 bucket.

ross-spencer:
Thanks for picking up on this @replaceafill and your time again @helenst. I'm just back from a brief period of leave and should be able to help merge this now. I am afraid there were a few commits in the last week, so it needs another rebase, are you okay to do that @helenst? I think a rebase will be fine. If you find any difficulties, then you might need to run the requirements `Makefile`, that seems to have been a pain-point a few times recently. But all might be okay.

helenst:
@ross-spencer Rebased - the merge conflict was Ok to fix :)

ross-spencer:
Perfect! (I was definitely over-thinking!) Thanks @helenst that's merged. We'll get cracking on the other two reviews and get back to you on the next merges. Sarah R posted [this](https://groups.google.com/forum/#!topic/archivematica/j0jXK9lArF4) yesterday as we try to open up contributions a bit more this release-sprint, and also keep on top of them. Might be worth keeping in mind. Thanks again!

helenst:
There is an issue with the way we check that the bucket exists... `get_bucket_location` can only be used by the bucket owner, and otherwise returns `AccessDenied` - this is quite restrictive and would not support our workflow in which the owner of the bucket is different from the credentials that Archivematica runs with. Given the rest of the plugin allows this, it seems a bit restrictive. I saw a comment in the code about using `head_bucket` - might be worth trying this with more recent boto versions.

ross-spencer:
@helenst definitely worth a try - I was just looking back at the commit history as this rang a few bells, but I can see our strategy here is one of growing sophistication which is nice and we definitely haven't tried this approach. If head_bucket can take care of it for most cases that would be great. Do you see a case where bucket creation should only ever be done by the owner of the bucket? Or perhaps, do you feel there is enough granularity in the storage service permissions to do what it needs to here? i.e. should the storage service mirror levels of permissions where certain roles perform certain tasks, or is this caught under the umbrella of who might administer the system in the first place?

PR #498, opened 2019-11-14
S3 plugin: allow blank authentication details

helenst:
Splitting out #459  into separate PRs - this addresses https://github.com/archivematica/Issues/issues/712 and allows the authentication fields on the S3 storage plugin to be left blank so that authentication information can be taken from the environment instead.

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -328,6 +328,21 @@ This is the current list of strings supported:
     - **Type:** `string`
     - **Default:** ``
 
+### AWS-specific environment variables
+
+    These variables can be set to allow AWS authentication for S3 storage spaces as an alternative
+	to providing these details via the user interface. See [AWS CLI Environment Variables](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-envvars.html) for details.
+
+- **`AWS_ACCESS_KEY_ID`**:
+    - **Description:** Access key for AWS authentication
+    - **Type:** `string`
+    - **Default:** ``
+
+- **`AWS_SECRET_ACCESS_KEY`**:
+    - **Description:** Secret key associated with the access key
+    - **Type:** `string`
+    - **Default:** ``
+
 ## Logging configuration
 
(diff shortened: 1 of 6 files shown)


ross-spencer (changes requested):
This looks great @helenst and the removal of the `client` property is a nice simplification. Adding `botocore` for the exceptions also. Great work.

I was taking a look at: https://github.com/artefactual/archivematica-storage-service/pull/499 but that looks like it might need changes too. I'm happy to coordinate with either this or 499 with you yet to get the PRs into the storage service. 

For this one can you:

* Update the migration number, following the previous PR where migration `0023_` was created.
* Can you amend the [environment variables](https://github.com/artefactual/archivematica-storage-service/blob/302c8825a35173df886003694afac5c60059e83b/install/README.md#environment-variables) section of the installation README to provide details of:

```
   AWS_ACCESS_KEY_ID=...
   AWS_SECRET_ACCESS_KEY=...
```

For running S3 with authentication details provided by the environment? 

It would also be great if you or one of your team can make a reference to that in the S3 docs online as well:

* [Here](https://www.archivematica.org/en/docs/storage-service-0.15/administrators/#s3-amazon)

I don't think that would need to meet the December deadline for code, we tend to write documentation later, but it would be great to get an indication y'all can commit to it. 

Do you think `botocore` needs to be requested explicitly in the `requitements.txt` or are you happy that it is pinned via `boto3`?

Otherwise, code-wise, I have tried this out, it works nicely, and I can see no other changes to make inline. So expect to be able to merge what's here. The docs changes can be done asynchronously.

helenst:
OK - I've updated the migration number, added to the install docs and pinned the botocore requirement.

Docs should be possible, my priority for this week is to meet the code deadlines but I should be able to do the docs after christmas.

ross-spencer (approved):
Perfect! Thanks @helenst :+1:

PR #500, opened 2019-11-20
Add package_name placeholder to post-store AIP/AIC/DIP callback

tw4l:
This PR addresses issue https://github.com/archivematica/Issues/issues/978.

It adds a `_replace_callback_placeholders` method to the Package model, which replaces the `<package_uuid>` and new `<package_name>` placeholders with appropriate values. It also adds mention and description of the new `<package_name>` placeholder to `callback_form.html` and extends the existing test `test_run_post_store_callbacks_aip` to check that placeholder values have been replaced in the callback URL and body. The latter required adding a new package fixture with a realistic `current_path` value and subsequently making some changes in `test_datatable.py` to account for the new fixture. I have made this new fixture a gzipped tar package to check that `_replace_callback_placeholders` works appropriately for packages with multiple extensions (e.g. `.tar.gz`).

I went with extending the existing testing strategy but would be happy to write a unit test for `_replace_callback_placeholders` if desired. Happy to make any other changes as well. Thanks!


storage_service/locations/models/package.py
```diff
@@ -1426,6 +1426,33 @@ def create_replicas(self):
...
+        equivalent to the post-sanitization Transfer name.
+        """
+        current_package = os.path.split(self.current_path)[1]
+        current_package_root = os.path.splitext(current_package)[0]
+        # Split on extension twice if package compressed as .tar.gz
+        if current_package_root.endswith(".tar"):
+            package_name_with_uuid = os.path.splitext(current_package_root)[0]
+        else:
+            package_name_with_uuid = current_package_root
+        package_name = package_name_with_uuid[:-37]
```
sevein: This section that identifies the name of the package could be moved to a class @property.

We've started using pathlib recently. This could be a good place to use it:

```python
>>> p = pathlib.Path('/var/lib/foo.bar.tar.gz')
>>> p.name
'foo.bar.tar.gz'
>>> p.suffixes
['.bar', '.tar', '.gz']
```

We could find the actual extension using .suffixes. You'd discard all the items off that list until you run into the first known extension. I think that we have a module somewhere defining known extensions, in the utils module?

storage_service/locations/models/package.py
```diff
@@ -1426,6 +1426,33 @@ def create_replicas(self):
...
+
+        uri = uri.replace("<package_uuid>", self.uuid).replace(
+            "<package_name>", package_name
+        )
+        body = body.replace("<package_uuid>", self.uuid).replace(
+            "<package_name>", package_name
+        )
```
sevein: It's common in Python to find these expressive ways of doing things. I find it hard but it's fun. Perhaps, you could use a list comprehension, e.g.:

```python
return [
    item.replace("<package_uuid>", self.uuid).replace("<package_name>", self.name)
    for item in (uri, body)
]
```

storage_service/locations/tests/test_package.py
```diff
@@ -328,12 +328,16 @@ def test_extract_file_aip_from_compressed_aip(self):
...
-        uuid = "0d4e739b-bf60-4b87-bc20-67a379b28cea"
+        uuid = "473a9398-0024-4804-81da-38946040c8af"
         aip = models.Package.objects.get(uuid=uuid)
         with mock.patch("locations.models.Callback.execute") as mocked_execute:
             aip.run_post_store_callbacks()
             # Only `post_store_aip` callbacks are executed
             assert mocked_execute.call_count == 1
+            # Placeholders replaced in URI and body
+            url = "http://consumer.com/api/v1/aip/%s/" % uuid
+            body = '{"name": "tar_gz_package"}'
```
sevein: Is there a reason why you didn't test the UUID placeholder?

storage_service/locations/models/package.py
```diff
@@ -1426,6 +1426,33 @@ def create_replicas(self):
         for replicator_loc in replicator_locs:
             self.replicate(replicator_loc)
 
+    def _replace_callback_placeholders(self, uri, body):
+        """
+        Replaces post store callback placeholders with values.
```
sevein: Please use the imperative mood. See PEP257:

> The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".

I think that we're not doing it consistently though.

sevein: Cosmetic, it should be:

        """Replace post store callback placeholders with values.

(without the line break)

sevein (changes requested):
Thanks!

storage_service/locations/models/package.py
sevein: Or...

 ```python
repls = ("<package_uuid>", "UUID"), ("<package_name>", "NAME")
return [
    reduce(lambda a, kv: a.replace(*kv), repls, item)
    for item in (uri, body)
]
```

I actually find these solutions so hard to read! 🤣

**Review by sevein — CHANGES_REQUESTED** [review]

File: storage_service/locations/models/package.py (hunk 1 above)
Comment (tw4l) [reply storage_service/locations/models/package.py]: Oh, that is nice. I've implemented similarly to how you suggest and added pathlib2 as a base requirement.

File: storage_service/locations/models/package.py (hunk 2 above)
Comment (tw4l) [reply storage_service/locations/models/package.py]: I find the lambda hard to read (I love an arrow function in JS, but in Python anonymous functions still seem weird to me 😂) so went with the list comprehension. Thanks!

File: storage_service/locations/tests/test_package.py (hunk 3 above)
Comment (tw4l) [reply storage_service/locations/tests/test_package.py]: The UUID placeholder is being tested in the URL (the placeholder uri is `""http://consumer.com/api/v1/aip/<package_uuid>/"`) but you're right it would be better to test in the body as well.

File: storage_service/common/utils.py [hunk 5]
```diff
@@ -533,6 +533,18 @@ def get_compression_transforms(aip, compression, transform_order):
...
+def get_compression_extensions():
+    """ Return list of compression and archive file extensions employed. """
+    extensions = [".tar"]
+    for ext in (
+        COMPRESS_EXTENSION_7Z,
+        COMPRESS_EXTENSION_BZIP2,
+        COMPRESS_EXTENSION_GZIP,
+    ):
+        extensions.append(ext)
+    return extensions
```
sevein: A better/simpler alternative could be to work just with tuples:

```python
COMPRESS_EXTENSIONS = (
    COMPRESS_EXTENSION_7Z,
    COMPRESS_EXTENSION_BZIP2,
    COMPRESS_EXTENSION_GZIP,
)

KNOWN_EXTENSIONS = (
    '.tar',
) + COMPRESS_EXTENSIONS
```

storage_service/locations/models/package.py
```diff
@@ -190,6 +195,20 @@ def full_pointer_file_path(self):
...
+
+        For AIPs created directly from a transfer, this will be
+        equivalent to the post-sanitization Transfer name.
+        """
+        full_name = Path(self.current_path).name
+        known_extensions = utils.get_compression_extensions()
+        for suffix in full_name.suffixes:
+            if suffix in known_extensions:
+                full_name.replace(suffix, '')
+        return full_name[:-37]
```
sevein: CI isn't running (will look into that later), but there is something wrong with this code. `suffixes` is a property of the Path object, `name` returns a string. Also `replace` does not mutate, it returns a new string so you need an assignment.

But I get the idea! Thanks. I'd definitely write a test for this with a bunch of different samples. You may find one case where it doesn't work very well, e.g.:

    /var/test/dataset.with.7z.files-a081c05e-ccc2-4bb6-a201-9562a261050a.tar.bz2

I'd expect `dataset.with.7z.files` instead of `dataset.with.files`.

I'm not sure if it's worth the effort, but here's an alternative:

```python
    @property
    def name(self):
        path = Path(self.current_path)
        name, ext = path.name, ''
        for (i, suffix) in reversed(list(enumerate(path.suffixes))):
            if suffix not in KNOWN_EXTENSIONS:
                ext = ''.join(path.suffixes[i+1:])
                break
        return name.replace(ext, '')[:-37]
```

sevein:
Oh I see what's going on with Travis. Your branch merges into qa/1.x but it should be merged into qa/0.x instead. Our .travis.yml is not including qa/1.x yet, but we should do that at some point.

https://travis-ci.org/artefactual/archivematica-storage-service/requests

storage_service/locations/models/package.py
tw4l: Thanks @sevein! Yeah, this is still very much a work in progress. I pushed the changes so far so I can work across computers but will definitely fix, add tests, and take your notes here into account. I'll let you know when it's ready - probably Monday, as I have some conflicting obligations today.

tw4l:
My mistake! I've rebased the PR on `qa/0.x`.

storage_service/locations/models/package.py
tw4l: Hi @sevein - coming back to this for a moment. I think the code snippet above has the same problem for some paths. E.g. for `/var/test/dataset.with.7z.files-a081c05e-ccc2-4bb6-a201-9562a261050a.7z`, the `.replace(ext, '')` will remove both `.7z` substrings, including the one that's not really a file extension and we want to keep.

I think a solution that avoids this issue might be something like:

```python
path = Path(self.current_path)
name, chars_to_remove = path.name, 37
for suffix in reversed(path.suffixes):
    if suffix not in KNOWN_EXTENSIONS:
        break
    chars_to_remove += len(suffix)
return name[:-chars_to_remove]
```

I'll make time Monday to properly test and make /push the appropriate changes, but any thoughts you have are welcome!

sevein: Nice! Perhaps I should have used rstrip(ext)?
But yours is great! Thanks! 👏

tw4l:
Hi @sevein, with the changes I pushed on Friday the Travis build is passing and it all looks good to me. I've added a test case `test_run_post_store_callbacks_aip_tricky_name` - between that and the existing `test_run_post_store_callbacks_aip` test, I think our coverage of edge cases for the name property is pretty good. Let me know if you'd like any further changes.

sevein (approved):
Thanks!

PR #501, opened 2019-11-27
Ensure pip requirements are up to date

ross-spencer:
We are seeing Jenkins build failures with `pathlib2` not being included in the requirements, it's not entirely clear if this is related to 978 where it was included, so should work? Or if it is related to the later addition of b22d605d2d8101ea0f70727eb3fbbd3c261b8ff3 but there are updates to the requirements when we run the [Makefile](https://github.com/artefactual/archivematica-storage-service/blob/26e473ded1db0c4a0d00e7730f5a48f136995111/requirements/Makefile) which you can see in the diff, so this seems a good point to run maintenance on that and update it. 

Connected to archivematica/issues#978


ross-spencer:
Closed due to issues getting the CI running on changing base branch.

PR #503, opened 2019-12-04
Filter packages in location detail view

replaceafill:
This PR fixes the location detail view to list only packages contained in that location.

The JS code that does the ajax requests now passes the location UUID back to the server using [the `fnServerParams` callback](http://legacy.datatables.net/ref).

Connected to https://github.com/archivematica/Issues/issues/953

### DIFF
--- a/storage_service/static/js/project.js
+++ b/storage_service/static/js/project.js
@@ -1,5 +1,7 @@
 $(document).ready(function() {
-  var uri = $("#user-data-packages").data("uri") || "/";
+  var $userDataEl = $("#user-data-packages");
+  var uri = $userDataEl.data("uri") || "/";
+  var location = $userDataEl.data("location-uuid");
   var dataTableOptions = {
     // List of language strings from https://datatables.net/reference/option/language
     oLanguage: {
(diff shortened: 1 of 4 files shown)


storage_service/static/js/project.js
```diff
@@ -1,5 +1,6 @@
 $(document).ready(function() {
   var uri = $("#user-data-packages").data("uri") || "/";
+  var location = $("#user-data-packages").data("location-uuid");
```
sevein: IIRC, it's preferably to cache the object (`var xyz = $("#user-data-packages")`) because the jq constructor is expensive.

storage_service/static/js/project.js
```diff
@@ -36,6 +37,12 @@ $(document).ready(function() {
   packagesDataTableOptions["bServerSide"] = true;
   packagesDataTableOptions["bProcessing"] = true;
   packagesDataTableOptions["sAjaxSource"] = uri + "packages_ajax";
+  if (location !== null && location !== "") {
```
sevein: Would the following be a better option?

```js
if (typeof(location) === "string" && location.length) { ... }
```

Since `data` may return `undefined` (not `null`?) but we'd only care when it's a `string`.

sevein (approved):
A couple of minor comments, but :+1:.

storage_service/static/js/project.js
replaceafill: :+1: Will fix.

PR #505, opened 2019-12-10
Use PREMIS 3 only in pointer files

replaceafill:
This sets the PREMIS version used in pointer files of encrypted and
replicated AIPs to be the one used by the `common.premis` module of
the Storage Service.

Connected to https://github.com/archivematica/Issues/issues/820

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -1237,7 +1237,7 @@ class Package(models.Model):
             for relationship in old_premis_object.findall("relationship")
         ]
         new_relationships.append(replication_relationship)
-        new_premis_meta = premisrw.PREMIS_META.copy()
+        new_premis_meta = premis.PREMIS_META.copy()
         new_premis_meta["xsi:type"] = old_premis_object.attributes.get(
             "xsi_type", "premis:file"
         )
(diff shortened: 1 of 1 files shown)


ross-spencer (approved):
Looks great. Thanks @replaceafill. One thing to note is that this change will break the AMAUAT tests for encryption which work primarily in the Docker environment as AMAUAT tests will be looking for PREMIS2. Do you consider fixing those in scope for this issue? 

Docker compose is a useful note for testers as a good way to setup and test this fix. The commands in the `makefile`:

* Encryption in src AIP: `sudo make test-at-behave TAGS=aip-encrypt BROWSER=Firefox`
* Relationship in src AIP and repl, + encryption in repl: `sudo make test-at-behave TAGS=aip-encrypt-mirror BROWSER=Firefox`

Once any of these tests complete, the pointer files can be downloaded manually as usual too so it's great to configure your environment, and doubly verify the result.

replaceafill:
Ah! Thanks for pointing that out @ross-spencer. I'll take a look at the broken AMAUATs :+1:

replaceafill:
@ross-spencer I updated the AMAUATs in https://github.com/artefactual-labs/archivematica-acceptance-tests/commit/e169fbe07c71ce2e1ecf4bf85a60b13acc9d8af9

PR #507, opened 2019-12-11
Upgrade metsrw

replaceafill:
This increases the version of the `mets-reader-writer` library so that pointer files are created with the newest METS version.

Connected to archivematica/Issues#949


ross-spencer (approved):
:+1:

PR #511, opened 2019-12-17
Ignore protected directories when counting objects

replaceafill:
This PR fixes object counting when access-protected directories exist in the transfer source. Under those circumstances `scandir.scandir` raises an `OSError` exception which is now handled and tested properly.

Connected to https://github.com/archivematica/Issues/issues/1019

### DIFF
--- a/storage_service/locations/tests/test_space.py
+++ b/storage_service/locations/tests/test_space.py
@@ -1,0 +1,131 @@
+import pytest
+from scandir import scandir
+
+from locations.models.space import path2browse_dict
+
+
+def _restrict_access_to(restricted_path):
+    """Simulate OSError raised by scandir when it cannot access a path."""
+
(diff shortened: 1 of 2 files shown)


storage_service/locations/tests/test_space.py
```diff
@@ -0,0 +1,131 @@
...
+        "properties": {
+            "empty": {"object count": 0},
+            "error.txt": {"size": 8},
+            "first": {"object count": 2},
+            "second": {"object count": 0},
+            "tree_a.txt": {"size": 6},
+        },
+    }
+
+    # restrict read access to thrid directory (child of second)
```
sevein: https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

sevein (approved):
LGTM

PR #512, opened 2019-12-24
Update urllib3 from 1.24.1 to 1.25.7

ablwr:
Connected to https://github.com/archivematica/Issues/issues/1025
and
https://github.com/artefactual/archivematica/pull/1530


ross-spencer (changes requested):
@ablwr we need to be careful with how requirements.txt is updated, see the top of the file [here](https://github.com/artefactual/archivematica-storage-service/blob/2cb8ed297c9951817637717486dcab8ecf61f6f6/requirements/base.txt#L1-L6
).

I think `urllib3` will either need to be updated by one of the two libs that are referenced in the comment, so you'll need to update requests maybe [here](https://github.com/artefactual/archivematica-storage-service/blob/dev/issue-1025-update-urllib3/requirements/base.in#L24) and then run compile. or I imagine you can pin a more up to date `urllib` specifically as well by adding it to `base.in`. There may be other tricks worth reading up on here too. 

If we merge as is, then I believe we are in danger of overwriting the requirements the next time someone runs pip-compile.

PR #515, opened 2020-01-07
check region when creating s3 bucket

ross-spencer:
Cherry-picked from https://github.com/artefactual/archivematica-storage-service/pull/494 for ease of review and merging cc. @fitnycdigitalinitiatives @replaceafill 

**NB.** _Comment to be added to 494_

Connected to archivematica/Issues#922

PR #516, opened 2020-01-13
Fix S3 bucket creation

replaceafill:
This fixes bucket creation in a S3 space when the bucket doesn't exist.

There were two problems:

- The `bucket_name` property of the model was outdated
- https://github.com/artefactual/archivematica-storage-service/pull/515 reintroduced an outdated call to the `client` property which was removed in https://github.com/artefactual/archivematica-storage-service/pull/498

Both of these problems could have been caught by the existing tests, but the mock setup in the S3 tests was also incorrect.

Connected to https://github.com/archivematica/Issues/issues/1048


ross-spencer (approved):
LGTM :+1:

PR #518, opened 2020-02-05
Restore GPG transform keys

replaceafill:
This PR sets a [custom version of the `metsrw`](https://github.com/artefactual-labs/mets-reader-writer/pull/78) that parses `mets:transformFile` subelements when the pointer file is read during reingest.

It also resets previous decompression transforms since a new compression event is generated at the end of reingest. This matches the old behaviour explained in this https://github.com/archivematica/Issues/issues/803#issuecomment-579890818

An old deprecated AMAUAT has also been restored to help testing this in https://github.com/artefactual-labs/archivematica-acceptance-tests/pull/156

Connected to https://github.com/archivematica/Issues/issues/803

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -2655,7 +2655,7 @@ class Package(models.Model):
             )
             # Since a new event has been created based on the new compression
             # mechanism we reset the old decompression transforms
-            self._remove_decompression_transforms(aip)
+            self._filter_and_remove_decompression_transforms(aip)
             # Add the new compression event
             aip.add_premis_event(compression_event)
             self._update_pointer_file(compression, mets, path=updated_aip_path)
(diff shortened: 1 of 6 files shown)


storage_service/locations/models/package.py
```diff
@@ -2665,6 +2669,23 @@ def _process_pointer_file_for_reingest(
...
+    def _get_transform_file_type(self, transform_file):
+        """Get the type of a transform file entry.
+
+        Keys are converted to uppercase by metsrw.FSEntry.__init__
+        but they are handled in lowercase in other modules here in the
+        Storage Service.
+        """
+        return transform_file.get("TYPE") or transform_file.get("type")
+
+    def _remove_decompression_transforms(self, aip):
```
ross-spencer: I'd be tempted to elaborate on the mechanics here in the function name, `_filter_and_remove_decompression_transform` which describes a little more about what I should expect without reading the function. But maybe it's not important -- reading the function just sent me down the path of wait, how many more transform types are we preserving, and why don't we remove them all? that question probably doesn't change -- the answer being `decryption` :slightly_smiling_face:

ross-spencer (approved):
Nice work @replaceafill one comment, but not hugely important. Haven't tested this outside of AMAUAT but it looks like a good fix.

ross-spencer:
I'll get `metsrw` reviewed so that the PIP requirements can be modified too.

PR #519, opened 2020-02-07
Merge Python 3 compatible changes

replaceafill:
This PR cherry picks most of the Python 3 compatible changes from https://github.com/artefactual/archivematica-storage-service/tree/dev/issue-806-python-3-upgrade. The only two commits left out are:

- [Update base image and requirements](https://github.com/artefactual/archivematica-storage-service/commit/0772315a721946c3b9a5d86a2aef67cd9b9aace3)
- [Disable sword2](https://github.com/artefactual/archivematica-storage-service/commit/36c9f7b16bac0b7713e67f7acf9b33fc54f99388)

An extra commit was created to fix linting.

I tested this in Jenkins using the `matrix-qa` features and even though I got a GPG space related error in the `aip-encryption.feature`, a local test run with docker using the same feature file passes. The `black-box` tag also passes in docker, so I'm fairly confident this will pass AMAUATs if it's merged.

Connected to https://github.com/archivematica/Issues/issues/806


install/make_key.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
+from __future__ import print_function
```
ross-spencer, line 2: I wonder if we should try `isort` or consider waiting until it is in CI, we would end up with `from __future__ import absolute_import, print_function` here I think. There may be other cases below.

ross-spencer, line 2: Reading the patterns in the migration, it seems that certain future imports are brought in as a matter of course to increase Py3 compatibility? I can see the benefit of that. Just looking to clarify as I begin the CR  where I don't believe the `absolute_import` has an impact on this particular script -- I think this would be hard to determine across the entire code-base so another + for doing it.

storage_service/administration/forms.py
```diff
@@ -1,10 +1,12 @@
+from __future__ import absolute_import
 from django import forms
 from django.contrib import auth
 from django.utils.translation import ugettext_lazy as _
 
 from common import utils
 
 from locations.models import Location, Space
+from six.moves import zip
```
ross-spencer, line 9: :+1:

storage_service/common/management/commands/import_aip.py
```diff
@@ -61,11 +62,11 @@
 import scandir
 from django.core.management.base import BaseCommand
 from django.db.utils import IntegrityError
-from django.utils.six.moves import input
```
ross-spencer, line 64: Oh interesting. Looking at `six` vs. `django.utils.six` is that the former is better maintained and more suited for our purpose?

storage_service/common/management/commands/import_aip.py
```diff
@@ -369,7 +370,7 @@ def check_if_aip_already_exists(aip_uuid):
             " want to import this AIP anyway (and destroy the existing one),"
             ' then enter "y" or "yes": '.format(aip_uuid)
         )
-        user_response = input(prompt)
+        user_response = eval(input(prompt))
```
ross-spencer: I wasn't sure the impact of `eval` here so I gave it a try with the management command. I am seeing an error, but perhaps I am missing something?

**Command**

```
python manage.py import_aip --compression-algorithm "7z with bzip" --decompress-source --pipeline 15affa1a-0bb8-499b-ad98-ce18b5b0f494 --aip-storage-location a272a155-2cfc-4a4d-92c9-613ebfd3394d demo_csv-ebca30c2-285a-4835-8722-de29981afdf5.7z
```

**Error**

```
An AIP with UUID ebca30c2-285a-4835-8722-de29981afdf5 already exists in this Storage Service? If you want to import this AIP anyway (and destroy the existing one), then enter "y" or "yes": y
  File "manage.py", line 11, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 394, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python2.7/site-packages/django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "/src/storage_service/common/management/commands/import_aip.py", line 153, in handle
    options["tmp_dir"],
  File "/src/storage_service/common/management/commands/import_aip.py", line 429, in import_aip
    check_if_aip_already_exists(aip_uuid)
  File "/src/storage_service/common/management/commands/import_aip.py", line 375, in check_if_aip_already_exists
    user_response = eval(input(prompt))
  File "<string>", line 1, in <module>
NameError: name 'y' is not defined
```

storage_service/locations/fixtures/vcr_cassettes/duracloud_move_from_ss_chunked_resume.yaml
```diff
@@ -196,27 +196,25 @@ interactions:
...
+        ZmVuZCBpcHN1bS4gRG9uZWMgaW4gcGhhcmV0cmEgbGlndWxhLgpNYWVjZW5hcyB2aXRhZSBpcHN1
+        bSBub24gbG9yZW0gY29udmFsbGlzIGVnZXN0YXMuIFNlZCBuaXNsIG5lcXVlLCBhY2N1bXNhbgph
+        dCBkaWFtIHF1aXMsIGFsaXF1ZXQgdmFyaXVzIGVyb3MuIFV0IHJ1dHJ1bSBsb3JlbSBzZWQgdGlu
+        Y2lkdW50IHN1c2NpcGl0LgpTZWQgdGVtcG9yIHRyaXN0aXF1ZSBpcHN1bSwgbmVjIHNvZGFsZXMg
+        c2FwaWVuIG9ybmFyZSB1dC4gQ3JhcyBuZXF1ZSBtZXR1cywKc29kYWxlcyBub24gbGFjdXMgc2l0
+        IGFtZXQsIG1heGltdXMgZWdlc3RhcyBhdWd1ZS4gQ3JhcyBzZWQgc2VtIG9yY2kuCgoKRHVpcyB1
+        bHRyaWNpZXMgZmF1Y2lidXMgbWV0dXMsIGF0IHVsdHJpY2VzIG5lcXVlIGdyYXZpZGEgYS4gVXQg
+        ZXggZXJhdCwgc2VtcGVyCmVnZXN0YXMgdm9sdXRwYXQgYWMsIHJ1dHJ1bSBub24gcG9zdWVyZS4K
+      - 0
+      - null
```
ross-spencer, line 217: Scalpel like accuracy in your changes here. Nice!

storage_service/common/utils.py
```diff
@@ -432,6 +435,18 @@ def get_tool_info_command(compression):
     return tool_info_command
 
 
+def get_7z_version():
+    return [
+        line
+        for line in subprocess.check_output("7z").splitlines()
+        if b"Version" in line
+    ][0].decode("utf8")
```
ross-spencer, line 443: :+1:

storage_service/locations/migrations/0001_initial.py
```diff
@@ -294,7 +295,7 @@ class Migration(migrations.Migration):
                         blank=True,
                         help_text=b"Identifier for the Archivematica pipeline",
                         unique=True,
-                        verbose_name=b"UUID",
+                        verbose_name="uuid",
```
ross-spencer: Is the case-change significant here? `UUID` vs. `uuid`?

storage_service/locations/migrations/0022_update_pipeline_help_text.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 from __future__ import unicode_literals
 
+from __future__ import absolute_import
```
ross-spencer, line 4: I think this might be last migration in the list, just noting that these all apply okay when run.

storage_service/locations/models/package.py
```diff
@@ -45,13 +46,15 @@
...
 from .fixity_log import FixityLog
+from six.moves import range
 
 __all__ = ("Package",)
 
 
 LOGGER = logging.getLogger(__name__)
 
 
+@six.python_2_unicode_compatible
```
ross-spencer, line 57: Useful decorator to know about.

storage_service/locations/tests/test_duracloud.py
```diff
@@ -308,6 +309,7 @@ def test_move_from_ss_chunked_file(self):
         shutil.copy(os.path.join(FIXTURES_DIR, "chunk_file.txt"), str(self.tmpdir))
         file_path = str(self.tmpdir / "chunk_file.txt")
         self.ds_object.CHUNK_SIZE = 10 * 1024  # Set testing chunk size
+        self.ds_object.BUFFER_SIZE = 1
```
ross-spencer, line 312: Oh, interesting, were these `BUFFER_SIZE` additions (`403`) determined by the Python 3 migration?

ross-spencer (changes requested):
Excellent work @replaceafill -- just one potential bug noticed. I should note as well, that like you I was running this code throughout a good part of the `py3` work so there's some good confidence built. 

I ran `pylint` and I wonder if you would address the `absolute_import` changes mentioned here, and feel free to leave this out, but if you had the time, could you maybe address the final compatibility issue recognized by the linter? I believe (I could be wrong) this is because we're reusing `e` for a variable we might otherwise call `entry` when `e` was previously defined as an exception object `e`. I think then `pylint --py3k` will return a huge thumbs up! 

```
pylint --py3k $(find * | grep py$)
************* Module locations.migrations.0024_allow_blank_aws_auth
[W1618 no-absolute-import] import missing `from __future__ import absolute_import` File: storage_service/locations/migrations/0024_allow_blank_aws_auth.py, line 4, in 
************* Module locations.migrations.0023_s3_bucket_field
[W1618 no-absolute-import] import missing `from __future__ import absolute_import` File: storage_service/locations/migrations/0023_s3_bucket_field.py, line 4, in 
************* Module locations.migrations.0025_update_package_size
[W1618 no-absolute-import] import missing `from __future__ import absolute_import` File: storage_service/locations/migrations/0025_update_package_size.py, line 4, in 
************* Module locations.tests.test_s3
[W1618 no-absolute-import] import missing `from __future__ import absolute_import` File: storage_service/locations/tests/test_s3.py, line 1, in 
************* Module locations.tests.test_space
[W1618 no-absolute-import] import missing `from __future__ import absolute_import` File: storage_service/locations/tests/test_space.py, line 1, in 
************* Module locations.models.space
[W1661 exception-escape] Using an exception object that was bound by an except handler File: storage_service/locations/models/space.py, line 757, in Space.browse_rsync

-----------------------------------
Your code has been rated at 9.99/10
```

Thank you too for the Jenkins report! I really appreciate its use in this context. I think Santi would too! We just need to figure out a way to make it trigger on command via the PR!

install/make_key.py
replaceafill: Yeah, I agree the imports look strange. But I'd rather we wait to define what style we're going to use for `isort` and then run it once accross all projects like we did with black.

replaceafill: During the Python 3 sprint I noticed several conversations in the python-modernize project regarding this:

- https://github.com/python-modernize/python-modernize/issues/142
- https://github.com/python-modernize/python-modernize/pull/165

I think that if the imports don't hurt and linting doesn't complain then it's +1 from me :blush:

storage_service/common/management/commands/import_aip.py
replaceafill: It's the same module, but Django's version is older:

- [Django 1.8 includes `six 1.10.0`](https://github.com/django/django/blob/6a0dc2176f4ebf907e124d433411e52bba39a28e/django/utils/six.py#L32)
- [We're using six 1.13](https://github.com/artefactual/archivematica-storage-service/blob/a5045c356466dbe117e397fc0b68663f14bc90b1/requirements/base.txt#L78)

storage_service/common/management/commands/import_aip.py
replaceafill: Really nice catch! The module was already using `six` and it seems like modernize considered the input to be the Python keyword so it reconverted it.

storage_service/locations/migrations/0001_initial.py
replaceafill: I think this was a mistake. Thanks for catching it!

storage_service/locations/tests/test_duracloud.py
replaceafill: If I remember correctly this was needed because the default buffer size is `1MB` and it also affects the chunking functionality in Duracloud.

The files used in the test fixtures are very small (a few KBs) so they didn't really trigger chunking. Setting the buffer size this small guarantees we'll hit that logic.

storage_service/locations/models/space.py
```diff
@@ -740,8 +740,8 @@ def browse_rsync(
             if assume_rsync_daemon:
                 env["RSYNC_PASSWORD"] = rsync_password
             output = subprocess.check_output(command, env=env)
-        except Exception as e:
-            LOGGER.warning("rsync list failed: %s", e, exc_info=True)
+        except Exception as error:
+            LOGGER.warning("rsync list failed: %s", error, exc_info=True)
```
ross-spencer, line 744: Really appreciate this one, thanks @replaceafill!

storage_service/locations/tests/test_duracloud.py
ross-spencer: Nice.

ross-spencer (approved):
I think this all looks good @replaceafill. And what does pylint say?

```
archivematica-storage-service$ pylint --py3k $(find * | grep py$)

-------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 9.99/10, +0.01)
```

💯👍

PR #520, opened 2020-02-09
Remove redundant compression event on reingest

replaceafill:
This PR removes a redundant PREMIS `compression` event in the pointer file of reingested AIPs.

An extra commit has been added to update the name and docstring of a helper used to set compression transforms on the AIP's `FSEntry`.

Connected to https://github.com/archivematica/Issues/issues/1062

### DIFF
--- a/storage_service/common/tests/test_utils.py
+++ b/storage_service/common/tests/test_utils.py
@@ -254,7 +254,7 @@ def test_get_format_info(compression, version, extension, program_name, transfor
     consistent.
     """
     fsentry = FSEntry()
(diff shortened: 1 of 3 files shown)


ross-spencer (approved):
I haven't tested this but the changes look good and I appreciate your work on the docstring. (This is a bit of a blind-spot for me when reading code.) @replaceafill can you get clarification from @sallain before merging, or as well-as merging? I am reading the comment on archivematica/issues#1062 differently in that while it is talking about events and we do have PREMIS events, it is also requesting the deletion of the activity/action performed on the file which is the driver for the event. If that is the case this can still be merged as a part-fix for that, or it can be updated. Either way. It reads as solid code.

PR #521, opened 2020-02-24
Replicate AIP bag contents not container dir

ross-spencer:
Ensure that during replication we're copying the AIP (bag) contents
and not the container directory. Copying the container directory
pushes the replicated structure one level further down than we expect
and so when we come to test for something, e.g. fixity, we cannot
find the structural aspects of the bag, e.g. manifests, to perform
validation.

Connected to archivematica/issues#1054

### DIFF
--- a/storage_service/locations/tests/test_package.py
+++ b/storage_service/locations/tests/test_package.py
@@ -11,6 +11,7 @@ from django.contrib.messages import get_messages
 from django.core.urlresolvers import reverse
 from django.test import TestCase
 
+from common import utils
 from locations import models
 
 import bagit
@@ -388,6 +389,30 @@ class TestPackage(TestCase):
             body = '{"download_url": "http://ss.com/api/v2/file/%s/download/"}' % uuid
             mocked_execute.assert_called_with(url, body)
 
+    @staticmethod
+    def _test_bagit_structure(replica, replication_dir):
+        """Ensure the replicated bagit structure remains."""
+        bag_contents = [
+            "tagmanifest-md5.txt",
+            "bagit.txt",
+            "manifest-md5.txt",
+            "bag-info.txt",
+            os.path.join("data", "test.txt"),
+        ]
+        expected_bag_path = os.path.join(
+            replication_dir, utils.uuid_to_path(replica.uuid), "working_bag"
+        )
+        expected_bagit_structure = [
+            os.path.join(expected_bag_path, bag_path) for bag_path in bag_contents
+        ]
+        for subdir, _, files in os.walk(replica.current_location.full_path):
+            for file_ in files:
+                file_path = os.path.join(subdir, file_)
+                if file_path not in expected_bagit_structure:
+                    pytest.fail(
+                        "Unexpected file in Bagit structure: {}".format(file_path)
+                    )
+
     def test_replicate_aip(self):
         space_dir = tempfile.mkdtemp(dir=self.tmp_dir, prefix="space")
         replication_dir = tempfile.mkdtemp(dir=self.tmp_dir, prefix="replication")
@@ -401,13 +426,12 @@ class TestPackage(TestCase):
         )
 
         aip.create_replicas()
-
         assert aip.replicas.count() == 1
-
         replica = aip.replicas.first()
         assert replica is not None
         assert replica.origin_pipeline == aip.origin_pipeline
         assert replica.replicas.count() == 0
+        self._test_bagit_structure(aip.replicas.first(), replication_dir)
 
     def test_replicate_aip_twice(self):
         space_dir = tempfile.mkdtemp(dir=self.tmp_dir, prefix="space")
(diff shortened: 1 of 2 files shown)


storage_service/locations/tests/test_package.py
```diff
@@ -378,6 +379,35 @@ def test_run_post_store_callbacks_dip(self):
             body = '{"download_url": "http://ss.com/api/v2/file/%s/download/"}' % uuid
             mocked_execute.assert_called_with(url, body)
 
+    @staticmethod
+    def _test_bagit_structure(replicas, repl_uuid, replication_dir):
```
sevein: I think you mean `replica` instead of `replicas`?

storage_service/locations/tests/test_package.py
```diff
@@ -378,6 +379,35 @@ def test_run_post_store_callbacks_dip(self):
...
+            os.path.join("data", "test.txt"),
+        ]
+        expected_bag_path = os.path.join(
+            replication_dir, utils.uuid_to_path(repl_uuid), "working_bag"
+        )
+        expected_bagit_structure = [
+            os.path.join(expected_bag_path, bag_path) for bag_path in bag_contents
+        ]
+        bag_structure = True
+        replica_ = replicas
```
sevein: This `replica_` might be leftovers from an older version that you were trying to write?

storage_service/locations/tests/test_package.py
sevein: Since you're already accepting a replica, you could just use `replica.uuid` instead of having the additional `repl_uuid` argument.

storage_service/locations/tests/test_package.py
```diff
@@ -378,6 +379,35 @@ def test_run_post_store_callbacks_dip(self):
...
+        path_ = replica_.current_location.full_path
+        for subdir, _, files in os.walk(path_):
+            for file in files:
+                file_path = os.path.join(subdir, file)
+                if file_path not in expected_bagit_structure:
+                    bag_structure = False
+        assert (
+            bag_structure is not False
+        ), "Unexpected file in Bagit structure: {}".format(file_path)
```
sevein: Your loop continues unnecessarily after the first anomaly is found.
You could do the following:

```python
for subdir, _, files in os.walk(replicas.current_location.full_path):
    for file in files:
        file_path = os.path.join(subdir, file)
        if file_path not in expected_bagit_structure:
            pytest.fail(
                "Unexpected file in Bagit structure: {}".format(file_path))
```

storage_service/locations/tests/test_package.py
```diff
@@ -378,6 +379,35 @@ def test_run_post_store_callbacks_dip(self):
...
+            replication_dir, utils.uuid_to_path(repl_uuid), "working_bag"
+        )
+        expected_bagit_structure = [
+            os.path.join(expected_bag_path, bag_path) for bag_path in bag_contents
+        ]
+        bag_structure = True
+        replica_ = replicas
+        path_ = replica_.current_location.full_path
+        for subdir, _, files in os.walk(path_):
+            for file in files:
```
sevein: `file` is a built-in function.

storage_service/locations/tests/test_package.py
```diff
@@ -422,6 +451,12 @@ def test_replicate_aip_twice(self):
         assert aip.replicas.count() == 2
         for replica in aip.replicas.all():
             assert replica.replicas.count() == 0
+            repl_dir = ""
+            for repl_path in (replication_dir, replication_dir2):
+                if replica.current_location.full_path.startswith(repl_path):
+                    repl_dir = repl_path
+                    break
+            self._test_bagit_structure(replica, replica.uuid, repl_dir)
```
sevein: I think that our tests would be more readable at a glance if we avoided loops.
That may not always going to be the case, but in this particular scenario what do you think about this alternative? I haven't tested, it may require some tweaking.

```python
assert aip.replicas.count() == 2
assert aip.replicas[0].replicas.count() == 0
assert aip.replicas[1].replicas.count() == 0
self._test_bagit_structure(
    replica[0], aip.current_location.replicators[0].relative_path)
self._test_bagit_structure(
    replica[1], aip.current_location.replicators[1].relative_path)
```

storage_service/locations/tests/test_package.py
```diff
@@ -378,6 +379,35 @@ def test_run_post_store_callbacks_dip(self):
             body = '{"download_url": "http://ss.com/api/v2/file/%s/download/"}' % uuid
             mocked_execute.assert_called_with(url, body)
 
+    @staticmethod
```
sevein, line 392: I'm wondering what are the implications in using staticmethod in TestCase. One issue is that you won't have access to its [methods](https://docs.python.org/3/library/unittest.html#unittest.TestCase) but that's obviously not a problem since we use pytest most of the time. But I'm wondering if there are unknowns like problems around transaction management, probably not but I figured I'd raise it just in case you know more.

sevein (changes requested):
Thanks Ross, nice test. I've made some suggestions.

storage_service/locations/tests/test_package.py
ross-spencer: Yeah, I tried to avoid them, but we can't access these objects through indices. I've proposed a solution that works, and I think might fall nicely in-between. It works because we're only dealing with two objects, but let me know if it works for you too.

storage_service/locations/tests/test_package.py
ross-spencer: I think interesting to meditate on, but I don't think we'll see an impact here.

storage_service/locations/tests/test_package.py
ross-spencer: Nice thinking, old habits here.

sevein (approved):

PR #522, opened 2020-02-26
Remove deleted package size from related resources

jraddaoui:
Refs https://github.com/archivematica/Issues/issues/312#issuecomment-589491356.

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -1980,11 +1980,14 @@ class Package(models.Model):
         Returns (True, None) on success, and (False, error_msg) on failure.
         """
         error = None
+        location = self.current_location
+        space = location.space
+
         # LOCKSS must notify LOM before deleting
-        if self.current_location.space.access_protocol == Space.LOM:
+        if space.access_protocol == Space.LOM:
             # Notify LOM that files will be deleted
             if "num_files" in self.misc_attributes:
-                lom = self.current_location.space.get_child_space()
+                lom = space.get_child_space()
                 lom.update_service_document()
                 delete_lom_ids = [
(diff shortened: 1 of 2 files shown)


storage_service/locations/models/package.py
```diff
@@ -2014,6 +2017,14 @@ def delete_from_storage(self):
             )
         self.status = self.DELETED
         self.save()
+
+        # Remove size from location and space used values. Like the initial
+        # checks and updates, this is not considering race conditions.
+        space.used -= self.size
+        space.save()
+        location.used -= self.size
+        location.save()
```
sevein, line 2026: Not sure how often we do this in practice, but should these three updates be part of a single transaction?

```python
with transaction.atomic():
    ...
```

storage_service/locations/tests/test_package.py
```diff
@@ -48,6 +48,16 @@ def setUp(self):
...
 
+    def test_model_delete_from_storage(self):
+        package = models.Package.objects.get(
+            uuid="88deec53-c7dc-4828-865c-7356386e9399"
+        )
+        with mock.patch("locations.models.Space.delete_path") as mocked_delete:
+            package.delete_from_storage()
+            assert mocked_delete.called
+            assert package.current_location.used == -595
+            assert package.current_location.space.used == -595
```
sevein: Nit: replace `-595` with `package.size`.

sevein (changes requested):

storage_service/locations/models/package.py
sevein: Discussed this offline. Radda reminded me that we're going to be addressing this issue and similar issues in Storage Service as part of the async refactoring.

sevein (approved):

PR #523, opened 2020-02-26
Package: convert size property to integer

sevein:
Manually dehydrate `BigIntegerField` as an integer since Tastypie can't do
that for us (https://github.com/django-tastypie/django-tastypie/pull/299).

Connects to https://github.com/archivematica/Issues/issues/1094.


jraddaoui (approved):

sevein:
I've added `KeyError` to the list although I don't know if `bundle.data` will ever do that in practice.

PR #533, opened 2020-05-25
Add integration tests

sevein:
This commit adds an additional level of testing that does not require us to set
up Archivematica pipelines. It is run on every commit and pull request made to
this repository using a GitHub workflow.

Presumably, this could be incorporated to our existing pytest suite using
custom markers. My suggestion is to decide that later on.

Some of the test scenarios will not pass until https://github.com/artefactual/archivematica-storage-service/pull/524 is merged.

Connected to https://github.com/archivematica/Issues/issues/1149.


integration/test_integration.py
```diff
@@ -0,0 +1,410 @@
...
+        resp = self.client.add_space(self.SPACES[self.dst])
+        assert resp.status_code == 201
+        space = json.loads(resp.content)
+
+        # 2. Add replicator location.
+        resp = self.client.add_location(
+            {
+                "relative_path": "aips",
+                "staging_path": "",
+                "purpose": "RP",
```
replaceafill: Could we use `Location` attributes in all of these `purpose` parameters?

integration/run.sh
```diff
@@ -0,0 +1,11 @@
...
+
+__dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+cd ${__dir}
+
+env COMPOSE_DOCKER_CLI_BUILD=1 DOCKER_BUILDKIT=1 docker-compose build minio archivematica-storage-service
+
+docker-compose run --rm archivematica-storage-service
+
+docker-compose down
```
replaceafill: Could `--volumes` be used here like in the action?

replaceafill (approved):
LGTM :+1:

PR #534, opened 2020-05-26
Update how integration worfklow triggers

sevein:
Change the way that the integration workflow is triggered. We want to trigger
the workflow on push and pull requests but only for the qa and stable branches.

Connects to https://github.com/archivematica/Issues/issues/1149.


jraddaoui (approved):
:ok_hand:

PR #536, opened 2020-06-03
Update link to contributing agreement (SS)

sallain:
Now points to a fillable PDF rather than the .txt file we had before.

I copied the sections verbatim from `artefactual/archivematica/CONTRIBUTING.md` so the formatting is a little different, but the text should be identical.


sromkey (approved):
:+1:

PR #537, opened 2020-07-13
Update requirements against pip-compile 5.1.2

ross-spencer:
Importlib-resources have been pinned per the advice [here](https://github.com/jazzband/pip-tools/issues/1174)

The commit doesn't update any other output from pip-compile other than those resolved by the pip-tools/pip-compile utility. 

Connected to archivematica/Issues#1252
Connected to archivematica/Issues#1253


sevein (approved):
:clap:

PR #539, opened 2020-07-29
Replace use of whitelist with allowlist

sevein:
Fixes https://github.com/archivematica/Issues/issues/1226.


tox.ini
```diff
@@ -37,7 +37,7 @@ setenv =
 basepython = python2
 skip_install = true
 deps = -rrequirements/test.txt
-whitelist_externals = bash
+allowlist_externals = bash
```
jraddaoui, line 40: :astonished: Both work?

jraddaoui (approved):

tox.ini
sevein: Yes, fixed just a few days ago! https://tox.readthedocs.io/en/latest/changelog.html#v3-18-0-2020-07-23

PR #543, opened 2020-08-24
Ensure cas_user_authenticated signal receiver is called

tw4l:
This commit moves the signal receiver for `cas_user_authenticated` to the `common` app and adds a custom AppConfig which imports `common.signals` to ensure that this and any future Django signal receivers in the app work as expected.

I propose moving the signal receiver to `common` so it is easily located next to the other SSO backends and middleware, including the custom CAS backend.

Connected to https://github.com/archivematica/Issues/issues/1289

### DIFF
--- a/storage_service/common/__init__.py
+++ b/storage_service/common/__init__.py
@@ -1,0 +1,1 @@
+default_app_config = "common.apps.CommonAppConfig"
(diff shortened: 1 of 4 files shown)


sevein (approved):
Thanks!

PR #544, opened 2020-10-08
Add management task to (re)create AIP replicas

tw4l:
Connected to https://github.com/archivematica/Issues/issues/1209

This commit introduces a new management command for the Storage Service which can be used to  create new replicas for all AIPs in an AIP Store location (specified with `--location`, defaults to default AIP Store). When called with the `-d/--delete` option, the command will first delete existing replicas of AIPs prior to creating new ones. This should allow Support to fix the issue in the linked issue by creating new AIP replicas.

Two questions for the reviewer that I'd like a second opinion/pointer for:

1. Currently, when called with the `--delete` argument, this command deletes the replicas from both the filesystem and the database. Would it be better to keep the entries for the deleted replicas in the database and change their status to "DELETED", rather than delete them outright from the database?
2. I'm suppressing most of the Storage Service log messages to make the terminal output for this command more clear, but these DEBUG messages are slipping through: `DEBUG     2020-10-13 09:12:37  locations.models.async_manager:async_manager:_watchdog_loop:123:  Watchdog sees 0 tasks running`. Any idea how to successfully suppress these as well?

### DIFF
--- a/storage_service/common/management/commands/create_aip_replicas.py
+++ b/storage_service/common/management/commands/create_aip_replicas.py
@@ -1,0 +1,158 @@
+# -*- coding: utf-8 -*-
+"""(Re)create replicas in an AIP store
+
+When called with the -d/--delete argument, this command will attempt to
+delete existing replicas of AIPs in the specified AIP Store location
+prior to creating new replicas.
+
+If an AIP Store location is not specified with the --location argument,
+the task will create replicas for all AIPs in the default AIP Storage
+location.
+
+Execution example:
+./manage.py create_aip_replicas --location <UUID>
+"""
+from __future__ import absolute_import, print_function
+import logging
+
+from django.core.management.base import CommandError
+from django.db import transaction
+
+from administration.models import Settings
+from common.management.commands import StorageServiceCommand
+from locations.models.package import Package
+
+# Suppress the logging from models/package.py.
+logging.config.dictConfig({"version": 1, "disable_existing_loggers": True})
+
+
+class ReplicaDeleteException(Exception):
+    pass
+
+
+def get_replica_count(aip_uuid):
+    """Return number of replicas for given AIP
+
+    :param aip_uuid: AIP UUID
+
+    :returns: Number of replicas (int)
+    """
+    return Package.objects.filter(replicated_package=aip_uuid).count()
+
+
+class Command(StorageServiceCommand):
+
+    help = __doc__
+
+    def add_arguments(self, parser):
+        """Entry point to add custom arguments"""
+        parser.add_argument(
+            "-d",
+            "--delete",
+            action="store_true",
+            help="Delete existing replicas in AIP Store location"
+            " prior to creating new replicas.",
+        )
+        parser.add_argument(
+            "--location",
+            help="UUID for AIP Storage location to create replicas",
+            default=None,
+        )
+
+    def handle(self, *args, **options):
+        delete_existing_replicas = False
+        if options["delete"]:
+            delete_existing_replicas = True
+
+        # Determine AIP storage location for which we'll be creating
+        # replicas. Use default AIP Store if user did not supply a UUID.
+        aip_store_uuid = options["location"]
+        if aip_store_uuid is None:
+            try:
+                default_aip_store = Settings.objects.get(name="default_AS_location")
+                aip_store_uuid = default_aip_store.value
+            except Settings.DoesNotExist:
+                raise CommandError("No AIP Store location specified or set as default.")
+
+        aips = Package.objects.filter(
+            current_location=aip_store_uuid,
+            package_type=Package.AIP,
+            replicated_package=None,
+        ).all()
+        if not aips:
+            raise CommandError(
+                "No AIPs to replicate in location {}".format(aip_store_uuid)
+            )
+
+        aips_count = len(aips)
+        success_count = 0
+        deleted_count = 0
+
+        self.info("AIPs to replicate: {}".format(aips_count))
+
+        for aip in aips:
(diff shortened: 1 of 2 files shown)


storage_service/common/management/commands/create_aip_replicas.py
```diff
@@ -0,0 +1,150 @@
...
+        # Determine AIP storage location for which we'll be creating
+        # replicas. Use default AIP Store if user did not supply a UUID.
+        aip_store_uuid = options["location"]
+        if aip_store_uuid is None:
+            try:
+                default_aip_store = Settings.objects.get(name="default_AS_location")
+                aip_store_uuid = default_aip_store.value
+            except Settings.DoesNotExist:
+                self.error("No AIP Store location specified or set as default.")
+                sys.exit(1)
```
sevein: One pattern in Django that we should consider adopting is to raise `CommandError`. See for example [sqlmigrate](https://github.com/django/django/blob/d88365708c554efe3c786c3be6da1d9de916360f/django/core/management/commands/sqlmigrate.py).

storage_service/common/management/commands/create_aip_replicas.py
```diff
@@ -0,0 +1,150 @@
...
+
+        aips = Package.objects.filter(
+            current_location=aip_store_uuid, package_type="AIP", replicated_package=None
```
sevein: You could use the `Package.AIP` constant here.

storage_service/common/management/commands/create_aip_replicas.py
```diff
@@ -0,0 +1,150 @@
...
+        ).all()
+        if not aips:
+            self.error("No AIPs to replicate in location {}".format(aip_store_uuid))
+            sys.exit(1)
+
+        aips_count = len(aips)
```
sevein, line 87: I was going to suggest to use [qs.iterator()](https://docs.djangoproject.com/en/3.0/ref/models/querysets/#iterator) and you may want to give it a shot, the truth is that in MySQL the benefit is quite not the expected. But it could be beneficial when using SQLite.

storage_service/common/management/commands/create_aip_replicas.py
```diff
@@ -0,0 +1,150 @@
...
+                        replica.uuid, aip_uuid, err
+                    )
+                )
+        return deleted_count
+
+    def _delete_replica(self, replica, aip_uuid):
+        """Delete replica from filesystem and database
+
+        :param replica: Package object of replicated AIP to delete.
+        :param aip_uuid: UUID of AIP whose replicas we are deleting.
```
sevein: `aip_uuid` is unused.

storage_service/common/management/commands/create_aip_replicas.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+        # Delete from filesystem.
+        fs_deletion = replica.delete_from_storage()
+        fs_deletion_success = fs_deletion[0]
+        if not fs_deletion_success:
+            error_msg = fs_deletion[1]
+            raise ReplicaDeleteException(error_msg)
```
sevein, line 153: I'm starting to think that we don't need custom exceptions when they don't have any specific behaviour/state or there is no specific control flow up in the stack. You could argue it's dead code after all and raising Exception would be just fine. What do you think?

sevein (changes requested):
Looking good, just a few minor comments.

sevein (commented):
> Currently, when called with the --delete argument, this command deletes the replicas from both the filesystem and the database. Would it be better to keep the entries for the deleted replicas in the database and change their status to "DELETED", rather than delete them outright from the database?

This is maybe a good question for @sromkey?

> I'm suppressing most of the Storage Service log messages to make the terminal output for this command more clear, but these DEBUG messages are slipping through: DEBUG 2020-10-13 09:12:37 locations.models.async_manager:async_manager:_watchdog_loop:123: Watchdog sees 0 tasks running. Any idea how to successfully suppress these as well?

I think that log event should eventually become a Prometheus-exported metric. You may want to do it as part of this pull request. Tons of examples in metrics.py (AM), I think you'd need a histogram metric for this? The alternative is probably to set up your local logging config to INFO or just assuming in production users don't set their loggers to the DEBUG level.

sromkey:
> Currently, when called with the --delete argument, this command deletes the replicas from both the filesystem and the database. Would it be better to keep the entries for the deleted replicas in the database and change their status to "DELETED", rather than delete them outright from the database?

@tw4l were just discussing and we're leaning towards the more conservative approach of changing the status rather than deleting the entry.

storage_service/common/management/commands/create_aip_replicas.py
tw4l: Yes, thanks! That was a leftover from an earlier iteration

storage_service/common/management/commands/create_aip_replicas.py
tw4l: Thanks @sevein, this is good to know! It looks like support for SQLite streaming wasn't added until Django 2.2, so we wouldn't see the benefits of this yet: https://docs.djangoproject.com/en/2.2/ref/models/querysets/#iterator

storage_service/common/management/commands/create_aip_replicas.py
tw4l: I don't have a strong opinion on this but suppose I'd slightly lean toward keeping the custom exception because it indicates that the exception was an anticipated one, and offers a bit of a hint to the end user about what happened. But I do see your point and would be fine with changing it if you prefer.

sevein (approved):

PR #545, opened 2020-10-15
Remove decoding on Duracloud chunk generation

jraddaoui:
Refs https://github.com/archivematica/Issues/issues/1314.

### DIFF
--- a/storage_service/locations/models/duracloud.py
+++ b/storage_service/locations/models/duracloud.py
@@ -319,9 +319,9 @@ class Duracloud(models.Model):
     def _process_chunk(self, f, chunk_path):
         bytes_read = 0
 
-        with open(chunk_path, "w") as fchunk:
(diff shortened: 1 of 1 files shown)


storage_service/locations/models/duracloud.py
```diff
@@ -321,7 +321,7 @@ def _process_chunk(self, f, chunk_path):
 
         with open(chunk_path, "w") as fchunk:
```
sevein: @jraddaoui, you may want to try opening it with modes `wb`. In binary mode, you should be able to write the byte sequence without having to do the conversion to unicode.

jraddaoui: Thanks @sevein!

sevein (approved):

sromkey:
@jraddaoui when you have time please go ahead and merge!

PR #553, opened 2020-12-14
Add configurable password validation to Storage Service

tw4l:
Connected to https://github.com/archivematica/Issues/issues/1332

This PR adds password validation to the Storage Service. The validators employed include:
- Django's MinimumLengthValidator (default length: 8 characters)
- Django's CommonPasswordValidator
- Django's UserAttributeSimilarityValidator
- A custom PasswordComplexityValidator, which checks that passwords contain at least 3 of: lower-case characters, upper-case characters, numbers, special characters.

The validators are enabled by default but can be disabled individually via the environment. Since the client requirement for password complexity is not universally seen as a positive and some clients may want to disable it (see for instance the newest [NIST password guidelines](https://en.wikipedia.org/wiki/Password_policy#NIST_guidelines), which discourage composition rules), we decided to make each validator individually configurable rather than providing one option to enable/disable all of them at once.

Password validation is enforced by Django when creating or editing users in the Storage Service GUI as well as through the `createsuperuser` and `changepassword` management commands.

Commit 7788241 disables password validation for the local environment. To test that password validation is working in a Docker development environment, either drop that commit locally or comment out `AUTH_PASSWORD_VALIDATORS = []` in `  storage_service/storage_service/settings/local.py`.

See also the accompanying PR for the Archivematica dashboard: https://github.com/artefactual/archivematica/pull/1661

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -70,6 +70,26 @@ of these settings or provide values to mandatory fields.
     - **Type:** `string`
     - :red_circle: **Mandatory!**
 
+- **`SS_AUTH_PASSWORD_MINIMUM_LENGTH`**:
+    - **Description:** sets minimum length for user passwords.
+    - **Type:** `integer`
+    - **Default:** `8`
+
+- **`SS_AUTH_PASSWORD_DISABLE_COMMON_VALIDATION`**:
+    - **Description:** disables password validation that prevents users from using passwords that occur in a list of common passwords.
+    - **Type:** `boolean`
+    - **Default:** `false`
+
+- **`SS_AUTH_PASSWORD_DISABLE_USER_ATTRIBUTE_SIMILARITY_VALIDATION`**:
+    - **Description:** disables password validation that prevents users from using passwords that are too similar to their username and other user attributes.
+    - **Type:** `boolean`
+    - **Default:** `false`
+
+- **`SS_AUTH_PASSWORD_DISABLE_COMPLEXITY_VALIDATION`**:
+    - **Description:** disables password validation that checks that passwords contain at least three of: lower-case characters, upper-case characters, numbers, special characters.
+    - **Type:** `boolean`
+    - **Default:** `false`
+
 - **`SS_SHIBBOLETH_AUTHENTICATION`**:
     - **Description:** enables the Shibboleth authentication system. Other settings related to Shibboleth cannot be defined via environment variables at the moment, please edit [storage_service.settings.base](../storage_service/storage_service/settings/base.py) manually.
     - **Type:** `boolean`
--- a/storage_service/administration/validators.py
+++ b/storage_service/administration/validators.py
@@ -1,0 +1,50 @@
+from __future__ import absolute_import
+import six
+
+from django.core.exceptions import ValidationError
+from django.utils.translation import ugettext as _
+
+
+class PasswordComplexityValidator(object):
+    """Custom password complexity validator.
+
+    To pass validation, passwords must contain at least three of:
+    - uppercase characters
+    - lowercase characters
+    - numbers
+    - special characters
+    """
+
+    HELP_TEXT = _(
+        "Your password must contain at least 3 of: uppercase characters, "
+        "lowercase characters, numbers, and special characters (symbols or "
+        "non-alphanumeric Unicode characters)."
+    )
+
+    def validate(self, password, user=None):
+        type_count = 0
+
+        has_lower = False
+        has_upper = False
+        has_number = False
+        has_special = False
+
+        unicode_password = six.ensure_text(password, errors="ignore")
+        for char in unicode_password:
+            if char.islower():
+                has_lower = True
+            elif char.isupper():
+                has_upper = True
+            elif char.isdigit():
+                has_number = True
+            else:
+                has_special = True
+
+        CHARACTER_TYPES = (has_lower, has_upper, has_number, has_special)
+        type_count = sum(val is True for val in CHARACTER_TYPES)
+
+        if type_count < 3:
+            raise ValidationError(self.HELP_TEXT, code="notcomplex")
+
+    def get_help_text(self):
+        return self.HELP_TEXT
--- a/storage_service/static/css/project.css
+++ b/storage_service/static/css/project.css
@@ -43,6 +43,8 @@ form ul {
 form li {
   list-style-type: none;
   padding-left: 0px;
+  font-size: 14px;
+  line-height: 20px;
 }
 
 .btn {
--- a/storage_service/storage_service/settings/base.py
+++ b/storage_service/storage_service/settings/base.py
@@ -13,20 +13,13 @@ from sys import path
 from django.core.exceptions import ImproperlyConfigured
 from django.utils.translation import ugettext_lazy as _
 
+from storage_service.settings.helpers import get_env_variable, is_true
+
 try:
     import ldap
     from django_auth_ldap import config as ldap_config
 except ImportError:
     ldap, ldap_config = None, None
-
-
-def get_env_variable(var_name):
-    """ Get the environment variable or return exception """
-    try:
-        return environ[var_name]
-    except KeyError:
-        error_msg = "Set the %s environment variable" % var_name
-        raise ImproperlyConfigured(error_msg)
 
 
 # ######## PATH CONFIGURATION
@@ -179,8 +172,10 @@ TEMPLATES = [
 
 # ######### AUTHENTICATION CONFIGURATION
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#authentication-backends
-
 AUTHENTICATION_BACKENDS = ["django.contrib.auth.backends.ModelBackend"]
+
+from .components.auth import *
+
 # ######### END AUTHENTICATION CONFIGURATION
 
 # ######### MIDDLEWARE CONFIGURATION
@@ -314,10 +309,6 @@ WSGI_APPLICATION = "%s.wsgi.application" % SITE_NAME
 # ######## END WSGI CONFIGURATION
 
 ALLOW_USER_EDITS = True
-
-
-def is_true(env_str):
-    return env_str.lower() in ["true", "yes", "on", "1"]
 
 
 ######### LDAP CONFIGURATION #########
--- a/storage_service/storage_service/settings/components/auth.py
+++ b/storage_service/storage_service/settings/components/auth.py
@@ -1,0 +1,46 @@
+# flake8: noqa
+
+"""Settings for basic user authentication."""
+
+from os import environ
+
+from storage_service.settings.helpers import is_true
+
+
+PASSWORD_MINIMUM_LENGTH = 8
+try:
+    PASSWORD_MINIMUM_LENGTH = int(
+        environ.get("SS_AUTH_PASSWORD_MINIMUM_LENGTH", PASSWORD_MINIMUM_LENGTH)
+    )
+except ValueError:
+    pass
+
+AUTH_PASSWORD_VALIDATORS = [
+    {
+        "NAME": "django.contrib.auth.password_validation.MinimumLengthValidator",
+        "OPTIONS": {"min_length": PASSWORD_MINIMUM_LENGTH},
+    }
+]
+
+# Additional password validation is enabled by default. To disable, set the
+# following AUTH_PASSWORD_* settings to false in the environment.
+if not is_true(environ.get("SS_AUTH_PASSWORD_DISABLE_COMMON_VALIDATION", "false")):
+    AUTH_PASSWORD_VALIDATORS.append(
+        {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator"}
+    )
+
+if not is_true(
+    environ.get(
+        "SS_AUTH_PASSWORD_DISABLE_USER_ATTRIBUTE_SIMILARITY_VALIDATION", "false"
+    )
+):
+    AUTH_PASSWORD_VALIDATORS.append(
+        {
+            "NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator"
+        }
+    )
+
+if not is_true(environ.get("SS_AUTH_PASSWORD_DISABLE_COMPLEXITY_VALIDATION", "false")):
+    AUTH_PASSWORD_VALIDATORS.append(
+        {"NAME": "administration.validators.PasswordComplexityValidator"}
+    )
--- a/storage_service/storage_service/settings/helpers.py
+++ b/storage_service/storage_service/settings/helpers.py
@@ -1,0 +1,16 @@
+from os import environ
+
+from django.core.exceptions import ImproperlyConfigured
+
+
+def get_env_variable(var_name):
+    """ Get the environment variable or return exception """
+    try:
+        return environ[var_name]
+    except KeyError:
+        error_msg = "Set the %s environment variable" % var_name
+        raise ImproperlyConfigured(error_msg)
+
+
+def is_true(env_str):
+    return env_str.lower() in ["true", "yes", "on", "1"]
--- a/storage_service/storage_service/settings/local.py
+++ b/storage_service/storage_service/settings/local.py
@@ -61,6 +61,11 @@ CACHES = {"default": {"BACKEND": "django.core.cache.backends.locmem.LocMemCache"
 # )
 # ########## END TOOLBAR CONFIGURATION
 
+# ######## AUTHENTICATION CONFIGURATION
+# Disable password validation in local development environment.
+AUTH_PASSWORD_VALIDATORS = []
+# ######## END AUTHENTICATION CONFIGURATION
+
 ######### LDAP CONFIGURATION #########
 if LDAP_AUTHENTICATION and DEBUG:
     # Don't validate certs if debug is on
--- a/storage_service/storage_service/tests/test_password_validation.py
+++ b/storage_service/storage_service/tests/test_password_validation.py
@@ -1,0 +1,46 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import
+
+import pytest
+
+from django.contrib.auth.password_validation import (
+    get_password_validators,
+    validate_password,
+)
+from django.core.exceptions import ValidationError
+
+
+@pytest.mark.parametrize(
+    "password,raises_exception",
+    [
+        # Passwords with only one type raise ValidationError.
+        ("abcdefghijk", True),
+        ("ABCDEFGHIJK", True),
+        ("12345678", True),
+        ("@#!_-=^$", True),
+        # Passwords with only two types raise ValidationError. Accented upper
+        # and lower case letters are treated as their ASCII equivalents and do
+        # not count as a special character.
+        ("abcdeFGHIJK", True),
+        ("ABÇDEF1234", True),
+        ("àbcdef!$@", True),
+        # Passwords with three types pass validation.
+        ("abcdef123!@#", False),
+        ("ABCdef12345", False),
+        ("#@!abcdefGHIJK", False),
+        # Passwords with all four types pass validation.
+        ("abcDEF123!@#", False),
+        ("#l33t#PASSWORD!", False),
+    ],
+)
+def test_password_complexity_validator(password, raises_exception):
+    VALIDATOR_CONFIG = [
+        {"NAME": "administration.validators.PasswordComplexityValidator"}
+    ]
+    validators = get_password_validators(VALIDATOR_CONFIG)
+
+    if raises_exception:
+        with pytest.raises(ValidationError):
+            _ = validate_password(password, password_validators=validators)
+    else:
+        _ = validate_password(password, password_validators=validators)
--- a/storage_service/locale/en/LC_MESSAGES/django.po
+++ b/storage_service/locale/en/LC_MESSAGES/django.po
@@ -8,7 +8,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2018-10-24 09:08-0700\n"
+"POT-Creation-Date: 2021-01-11 16:23-0800\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
 "Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
 "Language-Team: LANGUAGE <LL@li.org>\n"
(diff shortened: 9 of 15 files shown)


storage_service/storage_service/settings/base.py
```diff
@@ -181,6 +189,39 @@ def get_env_variable(var_name):
...
+    AUTH_PASSWORD_VALIDATORS += [
+        {
+            "NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator"
+        }
+    ]
+
+if not is_false(environ.get("SS_AUTH_PASSWORD_COMPLEXITY_VALIDATION", "")):
+    AUTH_PASSWORD_VALIDATORS += [
+        {"NAME": "administration.validators.PasswordComplexityValidator"}
+    ]
```
ross-spencer: This is a pretty gnarly file already. I'm not sure why it hasn't been treated like first-class code before and cared for like the rest of the code-base. As an example, I'm not sure why LDAP or shibboleth don't have clearly defined modules that are imported which would clean everything up, and also reduce the number of lines we have to read through. Authentication could have its own module too and be imported into `base` to be read by the app. 

That being said I understand the decision making here is to be consistent and it's looking good from that perspective. There's certainly no fault here. 

Can I suggest if we're not as an organisation in a place to start refactoring this module, to maybe introduce a pattern that could clean things up slightly and at least collect complex management of configuration in a separate function? e.g. `get_my_settings()` returns a list back to the configuration variable? 

I have created a diff to demonstrate a very basic approach below. There will likely be further optimizations. 

```diff
diff --git a/storage_service/storage_service/settings/base.py b/storage_service/storage_service/settings/base.py
index 287b673..258bf5e 100644
--- a/storage_service/storage_service/settings/base.py
+++ b/storage_service/storage_service/settings/base.py
@@ -37,6 +37,47 @@ def is_false(env_str):
     return env_str.lower() in ["false", "no", "off", "0"]
 
 
+def get_my_auth_settings():
+    """Collect the work needed to go to configure the storage service's
+    authentication mechanisms...
+    """
+
+    PASSWORD_MINIMUM_LENGTH = 8
+    PASSWORD_MINIMUM_LENGTH = int(
+        environ.get("SS_AUTH_PASSWORD_MINIMUM_LENGTH", PASSWORD_MINIMUM_LENGTH)
+    )
+
+    AUTH_PASSWORD_VALIDATORS = [
+        {
+            "NAME": "django.contrib.auth.password_validation.MinimumLengthValidator",
+            "OPTIONS": {"min_length": PASSWORD_MINIMUM_LENGTH},
+        }
+    ]
+
+    # Additional password validation is enabled by default. To disable, set the
+    # following AUTH_PASSWORD_* settings to false in the environment.
+    if not is_false(environ.get("SS_AUTH_PASSWORD_COMMON_VALIDATION", "")):
+        AUTH_PASSWORD_VALIDATORS.append(
+            {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator"}
+        )
+
+    if not is_false(
+        environ.get("SS_AUTH_PASSWORD_USER_ATTRIBUTE_SIMILARITY_VALIDATION", "")
+    ):
+        AUTH_PASSWORD_VALIDATORS.append(
+            {
+                "NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator"
+            }
+        )
+
+    if not is_false(environ.get("SS_AUTH_PASSWORD_COMPLEXITY_VALIDATION", "")):
+        AUTH_PASSWORD_VALIDATORS.append(
+            {"NAME": "administration.validators.PasswordComplexityValidator"}
+        )
+
+    return AUTH_PASSWORD_VALIDATORS
+
+
 # ######## PATH CONFIGURATION
 # Absolute filesystem path to the Django project directory:
 DJANGO_ROOT = dirname(dirname(abspath(__file__)))
@@ -187,43 +228,11 @@ TEMPLATES = [
 
 # ######### AUTHENTICATION CONFIGURATION
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#authentication-backends
-
 AUTHENTICATION_BACKENDS = ["django.contrib.auth.backends.ModelBackend"]
-
-try:
-    PASSWORD_MINIMUM_LENGTH = int(environ.get("SS_AUTH_PASSWORD_MINIMUM_LENGTH", 8))
-except ValueError:
-    PASSWORD_MINIMUM_LENGTH = 8
-
-AUTH_PASSWORD_VALIDATORS = [
-    {
-        "NAME": "django.contrib.auth.password_validation.MinimumLengthValidator",
-        "OPTIONS": {"min_length": PASSWORD_MINIMUM_LENGTH},
-    }
-]
-
-# Additional password validation is enabled by default. To disable, set the
-# following AUTH_PASSWORD_* settings to false in the environment.
-if not is_false(environ.get("SS_AUTH_PASSWORD_COMMON_VALIDATION", "")):
-    AUTH_PASSWORD_VALIDATORS += [
-        {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator"}
-    ]
-
-if not is_false(
-    environ.get("SS_AUTH_PASSWORD_USER_ATTRIBUTE_SIMILARITY_VALIDATION", "")
-):
-    AUTH_PASSWORD_VALIDATORS += [
-        {
-            "NAME": "django.contrib.auth.password_validation.UserAttributeSimilarityValidator"
-        }
-    ]
-
-if not is_false(environ.get("SS_AUTH_PASSWORD_COMPLEXITY_VALIDATION", "")):
-    AUTH_PASSWORD_VALIDATORS += [
-        {"NAME": "administration.validators.PasswordComplexityValidator"}
-    ]
+AUTH_PASSWORD_VALIDATORS = get_my_auth_settings()
 # ######### END AUTHENTICATION CONFIGURATION
 
+
 # ######### MIDDLEWARE CONFIGURATION
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#middleware-classes
 MIDDLEWARE = [
```

There are a couple of benefits I see to this including not having to use `### COMMENT BLOCKS ###` to describe a setting block in the code. And adding more documentation to functions that describe what's going on. Incidentally it also becomes testable if we were being particularly thorough.

storage_service/storage_service/settings/base.py
```diff
@@ -181,6 +189,39 @@ def get_env_variable(var_name):
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#authentication-backends
 
 AUTHENTICATION_BACKENDS = ["django.contrib.auth.backends.ModelBackend"]
+
+try:
+    PASSWORD_MINIMUM_LENGTH = int(environ.get("SS_AUTH_PASSWORD_MINIMUM_LENGTH", 8))
+except ValueError:
+    PASSWORD_MINIMUM_LENGTH = 8
```
ross-spencer: An approach that makes sure there's something clearly on the happy path and also encodes the default in a const might be:
```
    PASSWORD_MINIMUM_LENGTH = 8
    try:
        PASSWORD_MINIMUM_LENGTH = int(
            environ.get("SS_AUTH_PASSWORD_MINIMUM_LENGTH", PASSWORD_MINIMUM_LENGTH)
        )
    except ValueError:
        pass
```
I find it reads a little easier than a left-aligned `try/except` like this.

storage_service/storage_service/settings/base.py
```diff
@@ -181,6 +189,39 @@ def get_env_variable(var_name):
...
+
+# Additional password validation is enabled by default. To disable, set the
+# following AUTH_PASSWORD_* settings to false in the environment.
+if not is_false(environ.get("SS_AUTH_PASSWORD_COMMON_VALIDATION", "")):
+    AUTH_PASSWORD_VALIDATORS += [
+        {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator"}
+    ]
+
+if not is_false(
+    environ.get("SS_AUTH_PASSWORD_USER_ATTRIBUTE_SIMILARITY_VALIDATION", "")
```
ross-spencer: Creating the concept of `is_false` `is_true` creates somewhat of a quasi-type. Given that it's a small jolt to the eye to read the default as simply `""`. I feel it might help if this reads `is_false(
    environ.get("SS_AUTH_PASSWORD_USER_ATTRIBUTE_SIMILARITY_VALIDATION", True)` which I believe would still be valid according to your intentions?

storage_service/storage_service/settings/base.py
```diff
@@ -29,6 +29,14 @@ def get_env_variable(var_name):
         raise ImproperlyConfigured(error_msg)
 
 
+def is_true(env_str):
+    return env_str.lower() in ["true", "yes", "on", "1"]
+
+
+def is_false(env_str):
+    return env_str.lower() in ["false", "no", "off", "0"]
```
ross-spencer: Rather than introducing another handler here, I'm just wondering if we need to change the names of the variables being used. 

According to the PR, the new settings will default to on. If we name the variables and update the descriptions to let users know they "disable" not "enable" we can still use: `is_true`. 

```
- **`SS_AUTH_PASSWORD_COMMON_VALIDATION`**:>> SS_AUTH_PASSWORD_COMMON_VALIDATION_OFF
- **`SS_AUTH_USER_ATTRIBUTE_SIMILARITY_VALIDATION`**: >> SS_AUTH_USER_ATTRIBUTE_SIMILARITY_VALIDATION_OFF
- **`SS_AUTH_PASSWORD_COMPLEXITY_VALIDATION`**: >> SS_AUTH_PASSWORD_COMPLEXITY_VALIDATION_OFF
```

Given it seems that we seem to only use these values to disable, I think that would be preferable to do. I think that also makes sense in-line as well where testing for `not false` is more complicated that truthiness/falseness.

ross-spencer: That aside, I was trying to work through some of the complexity in `base.py` and answer this question and wrote-up a sample wrapper for the `environ` function. For my money, `is_true` and our work casting and watching for `ValueErrors` inline is getting pretty heavy here. It's a clue for me that a different abstraction is needed (and I am sure must be used in other projects or modules available). I wonder if we'd be better served by a `get_envvar` wrapper for `environ.get`. Something like:

https://gist.github.com/ross-spencer/b30115cea336016e34897bb4e1d4db04

I don't do a lot of work in these areas, so it's interesting to come in cold with a microscope when doing CR. 

This is probably out of scope. But if you think it worth it that code should capture what I can see is most cases in this module.

storage_service/administration/validators.py
```diff
@@ -0,0 +1,44 @@
...
+    To pass validation, passwords must contain at least three of:
+    - uppercase characters
+    - lowercase characters
+    - numbers
+    - special characters
+    """
+
+    HELP_TEXT = _(
+        "Your password must contain at least 3 of: uppercase characters, "
+        "lowercase characters, numbers, and special characters."
```
ross-spencer: I can imagine a user asking what a special character is. Do you see that as an issue here as well? Is it everything else on the keyboard? The shifted characters running 0-9 on the number pas? Is there anything possible to create on the keyboard that OWASP wouldn't like at all? https://owasp.org/www-community/password-special-characters - a small number of examples here might be enough.

storage_service/storage_service/tests/test_password_validation.py
```diff
@@ -0,0 +1,44 @@
...
+        ("12345678", True),
+        ("@#!_-=^$", True),
+        # Passwords with only two types raise ValidationError.
+        ("abcdEFGHIJK", True),
+        ("ABCDEF1234", True),
+        ("abcdef!$@", True),
+        # Passwords with three types pass validation.
+        ("abcdef123!@#", False),
+        ("ABCdef12345", False),
+        ("#@!abcdefGHIJK", False),
```
ross-spencer, line 30: It looks like accented characters work for upper case characters and I didn't realize regex was that agnostic, so that's cool! I wonder if you can perhaps add a few examples with different upper-case Unicode characters, e.g. some of `Á Â Ç Ë Ê Ñ Ö Ú Û` as a demonstration?

storage_service/static/css/project.css
```diff
@@ -43,6 +43,8 @@ form ul {
 form li {
   list-style-type: none;
   padding-left: 0px;
+  font-size: 14px;
+  line-height: 20px;
```
ross-spencer, line 47: This looks sensible to me on-screen.

install/README.md
```diff
@@ -70,6 +70,26 @@ of these settings or provide values to mandatory fields.
     - **Type:** `string`
     - :red_circle: **Mandatory!**
 
+- **`SS_AUTH_PASSWORD_MINIMUM_LENGTH`**:
+    - **Description:** sets minimum length for user passwords.
+    - **Type:** `integer`
+    - **Default:** `8`
+
+- **`SS_AUTH_PASSWORD_COMMON_VALIDATION`**:
+    - **Description:** enables password validation that prevents users from using passwords that occur in a list of common passwords.
```
ross-spencer: I left a note in the `base.py` module, but if we're only using these to disable rather than enable, I feel we can rename these, and make that clearer.

storage_service/storage_service/settings/local.py
```diff
@@ -61,6 +61,11 @@
 # )
 # ########## END TOOLBAR CONFIGURATION
 
+# ######## AUTHENTICATION CONFIGURATION
+# Disable password validation in local development environment.
+AUTH_PASSWORD_VALIDATORS = []
```
ross-spencer, line 66: This was helpful for testing! Thanks for thinking of it.

storage_service/storage_service/tests/test_password_validation.py
```diff
@@ -0,0 +1,44 @@
...
+
+from django.contrib.auth.password_validation import (
+    get_password_validators,
+    validate_password,
+)
+from django.core.exceptions import ValidationError
+
+
+@pytest.mark.parametrize(
+    "password,raises_exception",
```
ross-spencer, line 14: These all look like good tests. Nicely done.

storage_service/storage_service/settings/base.py
```diff
@@ -181,6 +189,39 @@ def get_env_variable(var_name):
...
+        "OPTIONS": {"min_length": PASSWORD_MINIMUM_LENGTH},
+    }
+]
+
+# Additional password validation is enabled by default. To disable, set the
+# following AUTH_PASSWORD_* settings to false in the environment.
+if not is_false(environ.get("SS_AUTH_PASSWORD_COMMON_VALIDATION", "")):
+    AUTH_PASSWORD_VALIDATORS += [
+        {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator"}
+    ]
```
ross-spencer: `AUTH_PASSWORD_VALIDATORS.append({ ... })` would be a little more idiomatic here unless there's a reason I am missing for not using it? It will also reduce the number of lines a little.

storage_service/administration/validators.py
```diff
@@ -0,0 +1,44 @@
+import re
```
ross-spencer: We're missing `from __future__ import absolute_import` at the top of this file which we need to add for later python3 compatibility. Most of the storage service modules should have this. I believe it's the only one impacted in your changes? 

This comes from testing against `pylint --py3k <module_name>`

ross-spencer (commented):
This looks good @tw4l. Nice and pragmatic, and simpler than I had imagined in my head. It's nice too that it doesn't affect existing passwords. I have left some questions and comments, and some prompts for cleaning up `base.py` a little bit. A question I was wrestling with the logic for a bit around variable naming. I have reviewed the storage service first, so that anything we learn here, you might also want to port over to Archivematica proper. Which seems like a sensible decision, but maybe I could have worked on them both in parallel? 

It confused me too that I think Archivematica has an 8 character password minimum already? And just the storage service doesn't. Is that what you know to the case also?

storage_service/storage_service/settings/base.py
tw4l: Yeah, agreed, the current `base.py` settings file here is not great. I'll look at refactoring this into modules, more akin to what's being done in the Dashboard, as at a quick glance I don't think that would take very much time or confer much risk. I also like the approach you're suggesting here so I'll keep an eye toward either mixing the two, or deferring to something like your diff if a proper refactor seems like more risk/time once I get into it.

storage_service/administration/validators.py
tw4l: My first thought was to be prescriptive about what characters were allowed, but I kind of like the permissive definition of special characters being anything that's not a lower-case character, upper-case character, or number. It does mean we allows other characters/code points such as emojis in passwords as well, so that a valid password might be, e.g., `abcd9876🐶`. This seems acceptable to Django in my testing and by my reading of the [docs](https://docs.djangoproject.com/en/1.11/ref/contrib/auth/#django.contrib.auth.models.User.password), which say that "Raw passwords can be arbitrarily long and can contain any character". OWASP is new to me - if you think it's better that we limit to that set, I think that would be fine as well.

What I'm reading as your main thrust here - that we should give some examples - is also very reasonable I think.

storage_service/storage_service/tests/test_password_validation.py
tw4l: Sure thing!

install/README.md
tw4l: Makes sense to me! That should remove the need for `is_false()` as well. I'd like to be as clear as possible to avoid confusion around the booleans (i.e. `True` turning off a validator) - how does `SS_AUTH_PASSWORD_DISABLE_COMMON_VALIDATION` sound as a name to you?

storage_service/storage_service/settings/local.py
tw4l: Glad you liked it! I wrestled with whether this was a good idea, but I think ultimately it's more useful than not.

storage_service/storage_service/settings/base.py
tw4l: `.append()` is certainly more idiomatic and I'm happy to switch it - we seem to use `+=` more commonly in the rest of this module and the Dashboard's `base.py` settings module but I don't see any reason why that needs to be the case or continue. I'm curious how it will reduce the number of lines though? Unless I'm missing something the lines required for the two seem the same, i.e.:

```
AUTH_PASSWORD_VALIDATORS.append(
    {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator"}
)
```

vs

```
AUTH_PASSWORD_VALIDATORS += [
    {"NAME": "django.contrib.auth.password_validation.CommonPasswordValidator"}
]
```

storage_service/storage_service/tests/test_password_validation.py
tw4l: Interesting - I decided to test this via CI with commit https://github.com/artefactual/archivematica-storage-service/pull/553/commits/95feb94c5bc222e35faf3e6526e63e622a171b4d as I'm having trouble running tests in my dev environment at the moment (I think due to my failed Ubuntu upgrade before the break) and it looks like the tests are now failing. From a quick reading of the [output of the failed test run](https://github.com/artefactual/archivematica-storage-service/pull/553/checks?check_run_id=1659239104), it looks like the accented characters are not being counted as upper/lower-case letters by regex. Curious that it's different than what you saw.

tw4l:
Thanks for the review @ross-spencer! I started replying to some of your comments and adding some commits, but have more work to do before it's ready for re-review. To respond to some of your questions/comments in the main thread:

> I have reviewed the storage service first, so that anything we learn here, you might also want to port over to Archivematica proper. Which seems like a sensible decision, but maybe I could have worked on them both in parallel?

That seems sensible to me! There are some very minor differences between the two, but largely just based on the existing state of the settings files. The core logic is mostly copy-pasted from one to the other.

> It confused me too that I think Archivematica has an 8 character password minimum already? And just the storage service doesn't. Is that what you know to the case also?

Yeah, it seems that [the Dashboard has an 8 character password minimum but it's only enforced when creating users through the GUI](https://github.com/artefactual/archivematica/blob/a31f218ceabc5bbe996775b82b9c9861b2f406d9/src/dashboard/src/components/accounts/forms.py#L90-L97), and the Storage Service [doesn't enforce a minimum length at all](https://github.com/artefactual/archivematica-storage-service/blob/f02b9149863efe3e883806b7d470ab3c55273d03/storage_service/administration/forms.py). I'm glad we're getting this chance to shore this up and get the two more in line with each other.

storage_service/storage_service/tests/test_password_validation.py
ross-spencer: Ack, sorry @tw4l I was modifying that test when reviewing so I may have unhelpfully modified it to pass (I actually can't remember what I did, but I do remember I was testing something around the raises exception part of the code, so might have messed that up. Or got the intention of a test wrong.) 

I can verify that I see the same behavior - accented characters are not treated as upper/lower-case.

My personal feeling is that it would be good to recognize them as such, but I think that's an important question for you and the project team, and it largely out of scope for me in CR. So I'll trust your judgement on an approach.

Tests haven't been working for me in `tox` either with a storage service virtual environment. Because I think the other side of our testing woes is that they can fail inconsistently in Docker as a suite I've resorted to running them one by one, so I'm running: `docker-compose run --workdir /src --rm --no-deps --entrypoint py.test -e "DJANGO_SETTINGS_MODULE=storage_service.settings.test" archivematica-storage-service -p no:cacheprovider -p no:warnings -v storage_service/storage_service/tests/test_password_validation.py`. I haven't looked into the standalone environment.

storage_service/administration/validators.py
ross-spencer: We've occasionally had access to OWASP reports on AtoM but nothing in Archivematica. It seems generally useful. My question was broadly around wondering what a special character was which you've helped me to understand. And looking in more detail at OWASP guidelines beyond that kind of glossary, they seem pretty comfortable with any and everything including white-space to that's cool: [OWASP cheat sheet - password controls](https://cheatsheetseries.owasp.org/cheatsheets/Authentication_Cheat_Sheet.html#implement-proper-password-strength-controls).

RE: examples - would be lovely! Although I'm not sure what direction you might take on that given such a broad set of characters.

install/README.md
ross-spencer: That seems like a good approach :+1:

storage_service/storage_service/settings/base.py
ross-spencer: > I'm curious how it will reduce the number of lines though?

In hindsight - I've no idea what made me think it was fewer lines sorry for the additional noise! :innocent:

storage_service/storage_service/settings/base.py
tw4l: `is_true` expects a string as input but I've modified it to the following to help with readability:

```python
if not is_true(environ.get("SS_AUTH_PASSWORD_DISABLE_COMMON_VALIDATION", "false")):
    ...
```

storage_service/administration/validators.py
tw4l: The examples are tricky. I tried a couple approaches but the help text kept getting pretty long or hard to read, so I've tried going with more of a brief explanation instead in commit https://github.com/artefactual/archivematica-storage-service/pull/553/commits/9cd4a3e9f4b88a2bf74a28fa6a45237f5007c8df:

"Your password must contain at least 3 of: uppercase characters, lowercase characters, numbers, and special characters (symbols or other non-alphanumeric Unicode characters)."

LMK if you'd rather see proper examples.

tw4l: As of commit https://github.com/artefactual/archivematica-storage-service/pull/553/commits/39599627341e4d1a3f852520e16b8dd5422c21b2 this description is now accurate, because accented alphabetical characters are treated as their ASCII equivalents.

storage_service/storage_service/tests/test_password_validation.py
tw4l: I agree with you that it's more correct to view accented alphabetical characters like their ASCII equivalents, so I've reworked the validator in commit https://github.com/artefactual/archivematica-storage-service/pull/553/commits/39599627341e4d1a3f852520e16b8dd5422c21b2 to achieve that. I looked at the [`regex` library on PyPI](https://pypi.org/project/regex/) per the top answer on [this SO post](https://stackoverflow.com/questions/36187349/python-regex-for-unicode-capitalized-words) but ultimately went with using Python's built-in unicode methods like `isupper()` and `isdigit()`, since those allow us to achieve the same result without bringing in an external library. With the help of six, I think the code there should be Python 2/3 compatible, and I checked with `pylint --py3k storage_service/administration/validators.py` for good measure.

The validator is slightly more complex now, but I think still pretty minimal and effective. I'm curious to hear your thoughts on the approach taken.

tw4l: Also thanks for the pointer for now to run tests individually in Docker! That's really helpful

storage_service/storage_service/settings/base.py
tw4l: In commit https://github.com/artefactual/archivematica-storage-service/pull/553/commits/02327ed6a395fb95975795745c57b0426b975cca, I've moved the auth settings other than `AUTHENTICATION_BACKENDS` into a new `components.auth` module, similar to how the LDAP/Shibboleth/CAS settings are organized in Archivematica proper. I opted for this instead of something like the suggested `get_my_auth_settings()` function (despite seeing some advantages there, such as for testing), as I think the existing Archivematica approach will extend better to other components such as LDAP/Shibboleth/CAS, which set a number of settings variables that won't necessary be easily returned as a list, for example. Does that seem reasonable to you?

After looking at it a bit, I'd rather not touch the other settings quite yet, as I feel like there's the potential for accidental regressions and the time required to set up environments to test and make sure everything is working correctly seems significant enough to be out-of-scope for this project. That said, hopefully you agree that what I have done here suffices as a first step in the direction of more modularity.

storage_service/storage_service/settings/base.py
tw4l: The proposed approach makes sense to me and I think it's good insight, but I'm inclined to agree that it's a bit out of scope. Especially given that we already have a [get_env_variable helper](https://github.com/artefactual/archivematica-storage-service/blob/stable/0.17.x/storage_service/storage_service/settings/base.py#L23-L29) with a slightly different focus that's in use in parts of `local.py` and `production.py` that are unrelated to this work but would need to be changed.

If we approach this in a separate piece of work, the [django-environ](https://github.com/joke2k/django-environ) library seems like something that might be worth looking into.

storage_service/storage_service/settings/base.py
tw4l: (edited the above to point out that it's `is_true`, not `environ.get`, that's the limiting factor here)

tw4l:
Hey @ross-spencer - I think this is now ready for re-review. Thanks!

Edit: After this next round of code review, I'll go back and update the Archivematica branch as appropriate. Just wanted to see your feedback on this work first in case we end up making more changes.

storage_service/storage_service/tests/test_password_validation.py
ross-spencer: Nice! I had started looking, but hadn't found a satisfying solution. I also leaned away from bringing in another library. This is great!

install/README.md
```diff
@@ -70,6 +70,26 @@ of these settings or provide values to mandatory fields.
...
+    - **Description:** disables password validation that prevents users from using passwords that occur in a list of common passwords.
+    - **Type:** `boolean`
+    - **Default:** `false`
+
+- **`SS_AUTH_PASSWORD_DISABLE_USER_ATTRIBUTE_SIMILARITY_VALIDATION`**:
+    - **Description:** disables password validation that prevents users from using passwords that are too similar to their username and other user attributes.
+    - **Type:** `boolean`
+    - **Default:** `false`
+
+- **`SS_AUTH_PASSWORD_DISABLE_COMPLEXITY_VALIDATION`**:
```
ross-spencer, line 88: Nice. These all ready well to me.

storage_service/storage_service/settings/components/auth.py
```diff
@@ -0,0 +1,42 @@
+from os import environ
```
ross-spencer, line 5: This is much cleaner now, and hopefully starts to pave a way for a more significant refactor of the original file. Thank you! 

If you feel so inclined it would be great if you could add a module docstring - I think it just helps adds the belt and braces to making storage service settings become more descriptive. Otherwise, I really appreciate this.

storage_service/storage_service/settings/helpers.py
```diff
@@ -0,0 +1,16 @@
+from os import environ
```
ross-spencer, line 1: `helpers.py` <-- nice addition! :+1:

storage_service/administration/validators.py
```diff
@@ -0,0 +1,52 @@
...
+    - lowercase characters
+    - numbers
+    - special characters
+    """
+
+    HELP_TEXT = _(
+        "Your password must contain at least 3 of: uppercase characters, "
+        "lowercase characters, numbers, and special characters (symbols or "
+        "non-alphanumeric Unicode characters)."
+    )
```
ross-spencer, line 22: This string would be fine with `black` on a single line. I prefer to do it like that nowadays because it avoids having to ensure a string terminates with a space, or begins in a space depending on who is encoding the string, i.e. written as a single stream, it tends to be self correcting. Not a huge deal if you prefer it this way. It's just me asking _"wait, did `black` format this like that?"_

This reminds me though - you'll need to make translations now you're adding this nice message! 

(The message looks good to me, even though it's verbose - and I tried a password with emoji and it worked!! :sweat_smile:)

storage_service/administration/validators.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def validate(self, password, user=None):
+        type_count = 0
+
+        has_lower = False
+        has_upper = False
+        has_number = False
+        has_special = False
+
+        unicode_password = six.ensure_text(password, errors="ignore")
```
ross-spencer, line 32: Nice. I don't think I would have remembered to use this!

storage_service/administration/validators.py
```diff
@@ -0,0 +1,52 @@
...
+                has_upper = True
+            elif char.isdigit():
+                has_number = True
+            else:
+                has_special = True
+
+        CHARACTER_TYPES = (has_lower, has_upper, has_number, has_special)
+        for character_type in CHARACTER_TYPES:
+            if character_type:
+                type_count += 1
```
ross-spencer: This shortcut might be of interest to you: `sum(val is True  for val in [True, True, True, False])` otherwise this is just fine too.

ross-spencer (approved):
Awesome @tw4l - I left one question about a module level docstring and a few notes/hints/positive thoughts along the way. I noted you'll need to run extract strings for translations - I am having a problem with that on my [docker-compose](https://github.com/artefactual-labs/am/pull/90) branch for that, so if you figure it out, I wouldn't mind hearing how you did it. Approving as this all seems like stuff you can tick off without any need for re-review but let me know if you need a second pair of eyes before merging. 

Edit: Thank you once again for going the extra mile for diacritics/accented characters. :pray:

storage_service/administration/validators.py
tw4l: > It's just me asking "wait, did black format this like that?"

Sharp eye - `black` in fact, did not format this for me! I'll keep that in mind about letting the formatter do the work moving forward. If we're using it, may as well rely on it!

> This reminds me though - you'll need to make translations now you're adding this nice message!

Good point! I'll do that as the last step before merging. Thanks for the reminder! :smile: 

> (The message looks good to me, even though it's verbose - and I tried a password with emoji and it worked!!)

That's great to hear! I briefly flirted with whether it would make sense to normalize as well, but figure that since the password is being hashed anyway, as long as the user is consistent about how they enter the password, it should work out in the end. I suspect we won't have too many users trying to put emoji in their passwords anyway, but then again who knows!

storage_service/administration/validators.py
tw4l: Oh, nice! Love a good readable shortcut :sunglasses:

tw4l:
Thanks again @ross-spencer! FYI, here's what worked to make the message files in Docker-Compose:

```
docker-compose run --workdir /src/storage_service \
    --rm --user=root \
    --entrypoint /src/storage_service/manage.py archivematica-storage-service makemessages

```

The only difference from the [PR you linked to](https://github.com/artefactual-labs/am/pull/90/files) is I changed the `--workdir` from `/src` to `/src/storage_service` to keep the `cwd` for the relative paths in the message files.

PR #552, opened 2020-12-11
Add package status filtering to create_aip_replicas command

tw4l:
QA of the new `create_aip_replicas` command by support identified an issue where the command did not work as expected when deleted AIPs were present in the Storage Service database. This PR ensures that packages are properly filtered to only those with status `Package.UPLOADED` to avoid this issue moving forward. I have tested the fix locally with a Storage Service database containing deleted AIPs (including deleted AIP replicas) and verified it is now working as originally intended.

This PR also includes a commit which removes code that explicitly sets a replicated AIP's status to `Package.DELETED` after deletion, since `Package.delete_from_storage()` already does the same.

Connected to https://github.com/archivematica/Issues/issues/1209

PR #554, opened 2021-01-01
Replication work for NHA

ross-spencer:
This PR implements three key components of replication and AIP deletion:

1. We now clean up AIP store quad-directory structures and do so safely.
2. When reingest is called in Archivematica replicas associated with that package are deleted. 
3. When reingest is called in Archivematica new replicas are created.

_**NB.** Contains the two commits from https://github.com/artefactual/archivematica-storage-service/pull/544 as they were used for reference as the work began and I anticipate merging on top of a `qa` branch with those in it anyway._

Testing has been added to back up these assertions trying to cover a wide-number of tests, including a handful of tests which ask digital-preservation like questions around file-properties. 

Tests have been implemented in `pytest` where possible, but the package suite is still in Python's unit test framework. They arei verbose. 

Small refactors have been done in functions we are working with, and those are documented in code, and in separate commits. Documentation of code has also been done in select cases and will largely be in separate commits. 

Connected to https://github.com/archivematica/Issues/issues/985
Connected to https://github.com/archivematica/Issues/issues/359

**Docs needed (note to self)**

* As well as: https://github.com/archivematica/Issues/issues/685
* Reingest now becomes a strategy for creating new replicas in new locations,
* As well as removing old replicas from old locations.

### DIFF
--- a/storage_service/common/utils.py
+++ b/storage_service/common/utils.py
@@ -584,7 +584,13 @@ def removedirs(relative_path, base=None):
     """ Removes leaf directory of relative_path and all empty directories in
     relative_path, but nothing from base.
 
-    Cribbed from the implementation of os.removedirs. """
+    Cribbed from the implementation of os.removedirs.
+
+    :param relative_path: quad-dir structure,
+        e.g. aa9a/f3b1/32ae/4f3e/8841/0539/568c/43f2 (path-string)
+    :param base: root location of quad-dir structure,
+        e.g. /var/archivematica/storage_service (path-string, (optional))
+    """
     if not base:
         return os.removedirs(relative_path)
     try:
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -1457,21 +1457,16 @@ class Package(models.Model):
     def create_replicas(self):
         """Create replicas of this AIP in any replicator locations.
 
-        Note: in a future iteration, this should be done asynchronously by put
-        requests to replicate this package on the "Package Replication Queue";
-        workers will read from this queue and replicate the package to
-        different locations asynchronously, e.g.,::
+        Existing replicas must be discovered and then deleted so that
+        they can then be updated, e.g. through Archivematica's reingest
+        capability.
 
-            >>> job = {'package': self.uuid, 'location': replicator_loc.uuid}
-            >>> RMQ_CHANNEL.basic_publish(
-                exchange='',
-                routing_key=PACKAGE_REPLICATION_QUEUE,
-                body=json.dumps(job).encode('utf8'),
-                properties=pika.BasicProperties(
-                    delivery_mode = 2,  # make message persistent
-                )
-            )
+        If this is the first iteration of an AIP then delete replicas
+        will be called and no replicas will be found to be deleted.
+
+        Once the clean-up work is done, new replicas are then minted.
         """
+        self._delete_replicas()
         replicator_locs = self.current_location.replicators.all()
         for replicator_loc in replicator_locs:
             self.replicate(replicator_loc)
@@ -1998,55 +1993,108 @@ class Package(models.Model):
             )
         return report, response
 
+    @staticmethod
+    def _if_lockss_notify(space, uuid, attributes):
+        """Notify lOM that files will be deleted."""
+        if not space.access_protocol == Space.LOM:
+            return
+        NUM_FILES = "num_files"
+        if NUM_FILES not in attributes:
+            return
+        lom = space.get_child_space()
+        lom.update_service_document()
+        delete_lom_ids = [
+            lom._download_url(uuid, idx + 1) for idx in range(attributes[NUM_FILES])
+        ]
+        return lom._delete_update_lom(delete_lom_ids)
+
+    @staticmethod
+    def _delete_pointer_file(
+        uuid, pointer_path, pointer_file_path, pointer_file_location
+    ):
+        """Delete pointer file and UUID quad directories."""
+        if not pointer_path:
+            return
+        try:
+            os.remove(pointer_path)
+        except OSError:
+            LOGGER.info(
+                "Error deleting pointer file %s for package %s",
+                pointer_path,
+                uuid,
+                exc_info=True,
+            )
+        utils.removedirs(
+            os.path.dirname(pointer_file_path), base=pointer_file_location.full_path
+        )
+
+    @staticmethod
+    def _update_storage_size(space, location, size):
+        """Remove size from location and space used values. Like the
+        initial checks and updates, this is not considering race
+        conditions.
+        """
+        space.used -= size
+        space.save()
+        location.used -= size
+        location.save()
+
+    def _find_replicas(self, status=UPLOADED):
+        """Find replicas associated with a given package.
+
+        :param status: Storage status within the storage service to aid
+            with filtering. (const string)
+        :returns: Replicas QuerySet object limited to only the replicas
+            which have the given status.
+        """
+        return Package.objects.filter(replicated_package=self.uuid, status=status).all()
+
+    def _delete_replicas(self):
+        """Finalize the deletion of a package by deleting its replicas
+        as well.
+        """
+        for replica in self._find_replicas():
+            _, replica_error = replica.delete_from_storage()
+            if replica_error:
+                LOGGER.error(
+                    "Could not delete replica: '%s' with error: '%s' please delete it manually",
+                    replica.uuid,
+                    replica_error,
+                )
+
     def delete_from_storage(self):
         """ Deletes the package from filesystem and updates metadata.
         Returns (True, None) on success, and (False, error_msg) on failure.
         """
+        LOGGER.debug(
+            "Deleting Package: '%s' from the storage service, replica: '%s'",
+            self.uuid,
+            True if self.replicated_package else False,
+        )
+
         error = None
         location = self.current_location
         space = location.space
 
-        # LOCKSS must notify LOM before deleting
-        if space.access_protocol == Space.LOM:
-            # Notify LOM that files will be deleted
-            if "num_files" in self.misc_attributes:
-                lom = space.get_child_space()
-                lom.update_service_document()
-                delete_lom_ids = [
-                    lom._download_url(self.uuid, idx + 1)
-                    for idx in range(self.misc_attributes["num_files"])
-                ]
-                error = lom._delete_update_lom(self, delete_lom_ids)
+        error = self._if_lockss_notify(space, self.uuid, self.misc_attributes)
+
         try:
             space.delete_path(self.full_path)
         except (StorageException, NotImplementedError, ValueError) as err:
             return False, err
-        # If deleted correctly, remove pointer file, and the UUID quad
-        # directories if they're empty.
-        pointer_path = self.full_pointer_file_path
-        if pointer_path:
-            try:
-                os.remove(pointer_path)
-            except OSError:
-                LOGGER.info(
-                    "Error deleting pointer file %s for package %s",
-                    pointer_path,
-                    self.uuid,
-                    exc_info=True,
-                )
-            utils.removedirs(
-                os.path.dirname(self.pointer_file_path),
-                base=self.pointer_file_location.full_path,
-            )
+
+        self._delete_pointer_file(
+            self.uuid,
+            self.full_pointer_file_path,
+            self.pointer_file_path,
+            self.pointer_file_location,
+        )
+
         self.status = self.DELETED
         self.save()
 
-        # Remove size from location and space used values. Like the initial
-        # checks and updates, this is not considering race conditions.
-        space.used -= self.size
-        space.save()
-        location.used -= self.size
-        location.save()
+        self._update_storage_size(space, location, self.size)
+        self._delete_replicas()
 
         return True, error
 
@@ -2253,18 +2301,19 @@ class Package(models.Model):
         call by making it more like the re-ingested AIP at
         ``reingest_location/path``.
 
-        1. Fetch the reingested AIP from the origin_location.
-        2. Replace this AIP's METS file with the reingested AIP's mets file.
-        3. Copy the new (reingested) metadata directory over the old
-           (current) one.
-        4. Copies preservation derivatives from reingested AIP to current AIP.
-           New files will be added, updated files will be overwritten.
-        5. Recreate the bagit manifest.
-        6. Compress the AIP according to what was selected during reingest in
-           Archivematica.
-        7. Create a pointer file if AM has not done so.
-        8. Store the AIP in the reingest_location.
-        9. Update the pointer file.
+        1.  Fetch the reingested AIP from the origin_location.
+        2.  Replace this AIP's METS file with the reingested AIP's mets file.
+        3.  Copy the new (reingested) metadata directory over the old
+            (current) one.
+        4.  Copies preservation derivatives from reingested AIP to current AIP.
+            New files will be added, updated files will be overwritten.
+        5.  Recreate the bagit manifest.
+        6.  Compress the AIP according to what was selected during reingest in
+            Archivematica.
+        7.  Create a pointer file if AM has not done so.
+        8.  Store the AIP in the reingest_location.
+        9.  Create or update replicas if they need to be made.
+        10. Update the pointer file.
 
         :param Location origin_location: Location the newly re-ingested AIP was
             procesed on.
@@ -2467,7 +2516,10 @@ class Package(models.Model):
                 )
                 write_pointer_file(revised_pointer_file, self.full_pointer_file_path)
 
-        # 9. Update the pointer file.
+        # 9. Create or update replicas if they need to be made.
+        self.create_replicas()
+
+        # 10. Update the pointer file.
         self._process_pointer_file_for_reingest(
             to_be_compressed, was_compressed, compression, updated_aip_path
         )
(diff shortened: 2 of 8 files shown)


storage_service/locations/models/package.py
```diff
@@ -1996,25 +1996,28 @@ def get_fixity_check_report_send_signals(
...
-            if "num_files" in self.misc_attributes:
-                lom = space.get_child_space()
-                lom.update_service_document()
-                delete_lom_ids = [
-                    lom._download_url(self.uuid, idx + 1)
-                    for idx in range(self.misc_attributes["num_files"])
-                ]
-                error = lom._delete_update_lom(self, delete_lom_ids)
+            # LOCKSS must notify LOCKSS-o-matic before deleting.
+            error = self._notify_lockss_matic(space, self.uuid, self.misc_attributes)
```
tw4l: This is nice clean-up work. Glad to see it!

storage_service/locations/models/package.py
```diff
@@ -2046,9 +2049,9 @@ def delete_from_storage(self):
         error = None
         location = self.current_location
         space = location.space
-        if space.access_protocol == Space.LOM:
-            # LOCKSS must notify LOCKSS-o-matic before deleting.
-            error = self._notify_lockss_matic(space, self.uuid, self.misc_attributes)
+
+        error = self._if_lockss_notify(space, self.uuid, self.misc_attributes)
```
tw4l, line 2079: Simple and effective. Nice work!

storage_service/locations/tests/test_space.py
```diff
@@ -130,3 +136,330 @@ def test_path2browse_dict_object_counting_ignores_read_protected_directories(
             "tree_a.txt": {"size": 6},
         },
     }
+
+
+# AIP store directory structure with components of the quad structure
+# we're generating created specifically to share branches which pushes
+# the limit of "possibilities' but is not impossible.
```
tw4l: Very minor editing issue: mixed (double, then single) quotations used here

storage_service/locations/tests/test_space.py
```diff
@@ -130,3 +136,330 @@ def test_path2browse_dict_object_counting_ignores_read_protected_directories(
...
+# Compressed or "packaged" AIP names to combine with the AIPstore dirs.
+C1 = "0a-11112222-3333-4444-5555-666677778888.7z"
+C2 = "0b-11112222-3333-4444-5555-666688889999.7z"
+C3 = "0c-11112222-3333-4444-5555-88889999aaaa.7z"
+C4 = "0d-11112222-3333-4444-8888-9999aaaabbbb.7z"
+C5 = "0e-11112222-3333-8888-9999-aaaabbbbcccc.7z"
+C6 = "0f-11112222-8888-9999-aaaa-bbbbccccdddd.tar.gz"
+C7 = "1a-11118888-9999-aaaa-bbbb-cccceeeeeeee.zip"
+
+magic = "\x37\x7A\xBC\xAF\x27\x1C"
```
tw4l: Should this constant be capitalized like the others?

storage_service/locations/tests/test_space.py
```diff
@@ -130,3 +136,330 @@ def test_path2browse_dict_object_counting_ignores_read_protected_directories(
...
+
+AIPSTORE = "AIPstore"
+
+
+@pytest.fixture
+def aipstore_compressed(tmpdir):
+    """Create a structure simulating an AIPstore with an appropriate
```
tw4l, line 193: I really appreciate the level of detail in this docstring. The directory structure diagram and explanation made it much easier to jump in and get my head around the test set-up :raised_hands:

storage_service/locations/tests/test_space.py
```diff
@@ -130,3 +136,330 @@ def test_path2browse_dict_object_counting_ignores_read_protected_directories(
...
+
+    :returns: (path to AIPstore (py.path, list of all paths, package
+        file (tuple)} (tuple)
+    """
+    aipstore = tmpdir.mkdir(AIPSTORE)
+    aipstore_path = str(aipstore)
+
+    # Pair paths with AIP names. The P{no} prefix is used to simplify
+    # the quad-path and the A{no} prefix is used to simplify the
+    # uncompressed package directory.
```
tw4l: Should this be "C{no}" and "compressed package"?

storage_service/locations/tests/test_space.py
```diff
@@ -130,3 +136,330 @@ def test_path2browse_dict_object_counting_ignores_read_protected_directories(
...
+P6 = os.path.join("1111", "2222", "8888", "9999", "aaaa", "bbbb", "cccc", "dddd")
+P7 = os.path.join("1111", "8888", "9999", "aaaa", "bbbb", "cccc", "eeee", "eeee")
+
+# Compressed or "packaged" AIP names to combine with the AIPstore dirs.
+C1 = "0a-11112222-3333-4444-5555-666677778888.7z"
+C2 = "0b-11112222-3333-4444-5555-666688889999.7z"
+C3 = "0c-11112222-3333-4444-5555-88889999aaaa.7z"
+C4 = "0d-11112222-3333-4444-8888-9999aaaabbbb.7z"
+C5 = "0e-11112222-3333-8888-9999-aaaabbbbcccc.7z"
+C6 = "0f-11112222-8888-9999-aaaa-bbbbccccdddd.tar.gz"
```
tw4l, line 178: Glad to see a `.tar.gz` in the test data. They seem to have been causing some issues elsewhere in the Storage Service.

storage_service/locations/tests/test_space.py
```diff
@@ -130,3 +136,330 @@ def test_path2browse_dict_object_counting_ignores_read_protected_directories(
...
+    :returns: (path to AIPstore (py.path, list of all paths, package
+        file (tuple)} (tuple)
+    """
+    aipstore = tmpdir.mkdir(AIPSTORE)
+    aipstore_path = str(aipstore)
+
+    # Pair paths with AIP names. The P{no} prefix is used to simplify
+    # the quad-path and the A{no} prefix is used to simplify the
+    # uncompressed package directory.
+    aips = [(P1, C1), (P2, C2), (P3, C3), (P4, C4), (P5, C5), (P6, C6), (P7, C7)]
```
tw4l: Since this is used elsewhere as well (e.g. as the parametrized inputs to `test_delete_compressed_path_local`), maybe it's worth pulling out into its own constant? That could also help simplify the setup of the latter a bit.

storage_service/locations/tests/test_space.py
```diff
@@ -130,3 +136,330 @@ def test_path2browse_dict_object_counting_ignores_read_protected_directories(
...
+
+    # Initialize our space and create a path to delete.
+    sp = Space()
+    path_to_delete = os.path.join(
+        str(aipstore_uncompressed[AIPSTORE_LOC]), path, aip_name
+    )
+
+    # rmtree is used to delete directories, we want to make sure we
+    # do call it from the delete function.
+    mocker.patch("shutil.rmtree", side_effect=shutil.rmtree)
```
tw4l, line 466: Clever!

storage_service/locations/tests/test_space.py
tw4l: Same for the uncompressed `aips` list below. Maybe could be named `COMPRESSED_AIPS` and `UNCOMPRESSED_AIPS` and defined as constants up top?

storage_service/common/utils.py
```diff
@@ -584,7 +584,13 @@ def removedirs(relative_path, base=None):
     """ Removes leaf directory of relative_path and all empty directories in
     relative_path, but nothing from base.
 
-    Cribbed from the implementation of os.removedirs. """
+    Cribbed from the implementation of os.removedirs.
+
+    :param relative_path: quad-dir structure,
+        e.g. aa9a/f3b1/32ae/4f3e/8841/0539/568c/43f2 (path-string)
+    :param base: root location of quad-dir structure,
+        e.g. /var/archivematica/storage_service (path-string, (optional))
```
tw4l, line 592: Really appreciate you fleshing out these docstrings! :raised_hands:

storage_service/locations/models/space.py
```diff
@@ -795,6 +800,45 @@ def _delete_path_local(self, delete_path):
...
+        utils.removedirs(quad_dir, base=base_dir)
+
+    def _delete_path_local(self, delete_path):
+        """
+        Deletes `delete_path` in this space, assuming it is locally accessible.
+        """
+        if not os.path.exists(delete_path):
+            return
+        self._del_package(delete_path)
+        self._delete_quad_dir_structure(delete_path)
```
tw4l: This is a really nice and readable implementation. Nice work!

I'm wondering if we might want to log instances where the path doesn't exist? I'm less familiar with this part of the codebase, but would that imply that something has gone wrong/is unexpected?

storage_service/locations/tests/test_space.py
```diff
@@ -449,8 +499,13 @@ def test_delete_uncompressed_path_local(aipstore_uncompressed, path, aip_name, m
...
 
+    # Ensure that the correct parts of the quad-directory structure
+    # remain.
+    assert os.path.exists(os.path.join(aipstore, remaining_directory))
+    assert not os.path.exists(os.path.join(aipstore, deleted_directory))
 
-def test_delete_non_existant_path_local(tmpdir):
+
+def test_delete_non_existant_path_local(tmpdir, aipstore_uncompressed):
     """Test that the behavior when deleting non-existent paths."""
```
tw4l: It looks like this docstring could use a bit of editing

storage_service/locations/models/package.py
```diff
@@ -2042,10 +2042,39 @@ def _update_storage_size(space, location, size):
         location.used -= size
         location.save()
 
+    def _find_replicas(self, status=UPLOADED):
```
tw4l, line 2042: Since `UPLOADED` is a class constant, would it be better to write this as `status=Package.UPLOADED` (as is seemingly used in other methods) or `status=self.UPLOADED`?

storage_service/locations/tests/test_package.py
```diff
@@ -50,14 +65,110 @@ def tearDown(self):
...
+        assert package.current_location.space.used == -package.size
+
+        # Ensure that the replicas and the original package have an
+        # updated status of DELETED.
+        replicas = package._find_replicas(status=models.Package.DELETED)
+        assert package.status == models.Package.DELETED
+        assert replicas[0].status == models.Package.DELETED
+        assert replicas[1].status == models.Package.DELETED
+
+    def test_model_delete_failure_with_replicas(self):
```
tw4l, line 147: I really appreciate the thoroughness and attention to detail of the testing in this PR. Really really solid work.

storage_service/locations/models/package.py
```diff
@@ -1455,21 +1455,16 @@ def create_pointer_file(
...
+        Existing replicas must be discovered and then deleted so that
+        they can then be updated, e.g. through Archivematica's reingest
+        capability.
+
+        If this is the first iteration of an AIP then delete replicas
+        will be called and no replicas will be found to be deleted.
+
+        Once the clean-up work is done, new replicas are then minted.
         """
+        self._delete_replicas()
```
tw4l, line 1469: Simple and effective, once again! This is great work, and a really clear interface :ok_hand:

storage_service/locations/models/package.py
```diff
@@ -2518,7 +2513,10 @@ def finish_reingest(
                 )
                 write_pointer_file(revised_pointer_file, self.full_pointer_file_path)
 
-        # 9. Update the pointer file.
+        # 9. Create or update replicas if they need to be made.
+        self.create_replicas()
+
+        # 10. Update the pointer file.
```
tw4l, line 2522: Excellent! It would be great to update the steps in the method docstring to match.

storage_service/locations/tests/test_package.py
```diff
@@ -739,6 +741,125 @@ def test_deletion_and_creation_of_replicas_compressed(self):
...
+        for f in files:
+            with open(os.path.join(aip_dir, f), "w") as test_file:
+                test_file.write(DATA_TO_WRITE)
+        return tmp_dir
+
+    def test_deletion_and_creation_of_replicas_uncompressed(self):
+        """Ensure that an uncompressed package is created properly.
+        Because replication seeks to also update the package we try
+        adding new files, e.g. it could be through enabling
+        normalization during partial-reingest. Through these tests We
```
tw4l: Editing nit: the w in "We" should be lower-case

tw4l (approved):
Hi @ross-spencer! This is really excellent work, and organized in a way that made code review extremely easy on me, so thank you for that! I especially appreciate the level of detail and care that's evident here in making sure that the test coverage is excellent for this feature (which is nice to see in terms of the project but especially for something like this where deletion is involved) as well as improving the readability and maintainability of the existing code, tests included.

In addition to looking through the code and running the tests, I also verified that everything is working as expected in my development environment by:
1. Setting up a replicator location associated with my local filesystem AIP store
2. Ingesting some AIPs of different types (uncompressed, different flavours of compressed)
3. Reingesting AIPs and verifying that the old replica and associated UUID quad were deleted, and new ones created
4. Deleting AIPs and verifying that both the original AIP and the replicas, along with their associated UUID quads, were deleted as expected

In all cases I saw exactly the behaviour that I would expect, and nothing unexpected.

Question: Have you tested this code with Storage Service location types other than local filesystem? Do you have a sense if this might work with any of the others, or if there might be any unintended side effects in those locations? What would happen e.g. if the AIP store and replicator locations were in S3? I understand that's out of scope so I'm not looking for more work here to make it work for other storage types, but rather a sense of what if any the implications outside of that scope might be, understanding that the answer may well be none.

Other than that, I've left a few questions/comments and a few nits where I saw e.g. docstring typos, but they're all quite minor. Given that, I'm approving this PR so that you don't need to go through another round of CR after making the changes you feel are appropriate. That said, feel free to just ping me if you want to discuss or get a second pair of eyes on anything.

I do see that you have the two commits from https://github.com/artefactual/archivematica-storage-service/pull/552 in this branch as well - I'm going to merge that in today, which should make removing those from this branch as easy as just rebasing on `qa/1.x`. (Your work here also means we can likely simplify that command greatly and have more confidence in its efficacy, which is fantastic!)

ross-spencer:
@tw4l thanks for this review and your kind comments. I'll look to fixup these points and prepare to milestone and merge. 

I wanted to make sure I followed up with your larger question:

> Question: Have you tested this code with Storage Service location types other than local filesystem? Do you have a sense if this might work with any of the others, or if there might be any unintended side effects in those locations? What would happen e.g. if the AIP store and replicator locations were in S3? I understand that's out of scope so I'm not looking for more work here to make it work for other storage types, but rather a sense of what if any the implications outside of that scope might be, understanding that the answer may well be none.

I have been meditating on that throughout the PR and S3 at least should be in-scope for the client as a replicator. Have a look at the thread here and follow it down throughout the slight existential crisis: https://github.com/archivematica/Issues/issues/359#issuecomment-754401482 

I thought I had S3 completely tested by hand while working on this. But I gave it a whirl, and there's an issue with pointer files when using S3 first and compressed packages. I have clarified is out of scope for the client right now: https://github.com/archivematica/Issues/issues/1342 

So thank you for that prompt. 

That aside, I double checked a S3 first then replica setup for uncompressed and that all works nicely. Which is what I did anticipate. 

Ultimately, the responsibility for things like deletion falls on the responsibility of the storage adapters themselves. So the S3 implementation of the Space interface as its own delete_path for example, and that works nicely as referenced in 359 above. 

I think it's also fair to ask, and obviously aspirational there's a place we need to get to where it is going to be easier to understand the implications throughout the storage service for each of these different permutations. The pointer file issue above alluded me this time, but I'm glad it's recorded now.

storage_service/locations/models/package.py
ross-spencer: That's an interesting point Tessa and I think I thought the same. I don't think the function signature knows its own context and so `self` and `Package` cannot be referenced from there. I don't know the technical limitation here, but worth a try if you get curious.

storage_service/locations/tests/test_space.py
ross-spencer: Welp (my consistency with this language!)! :slightly_smiling_face:

storage_service/locations/models/space.py
ross-spencer: I am happy to put a `LOGGER.debug` in here. That only shows up in the most verbose of logs and I cannot imagine this will create a lot of noise. Good suggestion.

storage_service/common/utils.py
ross-spencer: Your influence! I have some issues with support and consistency of docstrings in Python in general but I am increasingly seeing the advantage. And I think I am settling into a consistency with their formatting from our AIPscan work. And also for for a function like this which despite its seeming simplicity, just really needs to be as correct as it can be :ok_hand: seemed extra needed.

storage_service/locations/tests/test_space.py
ross-spencer: That's a good idea. In hindsight, returning the tuple here from the fixture and how they needed to be accessed seems like a bit of a smell, and the two consts clean this up nicely. :+1:

storage_service/locations/tests/test_space.py
ross-spencer: Good catch!

storage_service/locations/tests/test_space.py
ross-spencer: Done! and good catch! :slightly_smiling_face:

storage_service/locations/tests/test_space.py
ross-spencer: Nice, also I think I meant probability!

storage_service/locations/models/package.py
tw4l: It looks like you're right about the signature, Ross! 

The More You Learn GIF

PR #560, opened 2021-01-25
S3 increased timeouts and config debugging

ross-spencer:
Introduces a configuration block for S3 which will be useful for more fine-grained control in future. In this instance it allows us to increase the S3 timeout to enable big packages to be sent over the wire where we were seeing some system limits breached before. The default 60 second timeout was too short and so we have opted for 900 which is referenced in a clients S3 implementation from Hitachi. 

Users can now control Boto3 logging more easily using the standard mechanisms available in the Storage Service through Python's logging implementation. The logging README has been brought up to date and is hopefully a little more user-friendly. We would like more than just developers to make use of the logging abilities of the storage service for debugging. 

This PR is complemented with a [storage service docs](https://github.com/artefactual/archivematica-storage-service-docs/pull/45) PR that will help users of the S3 service. 

Connected to archivematica/issues#1346

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -125,6 +125,11 @@ of these settings or provide values to mandatory fields.
     - **Type:** `boolean`
     - **Default:** `false`
 
+- **`SS_S3_TIMEOUTS`**:
+    - **Description:** read and connect timeouts for S3 matching your implementation's recommended defaults.
+    - **Type:** `integer`
+    - **Default:** `900`
+
 The configuration of the database is also declared via environment variables. Storage Service looks up the `SS_DB_URL` environment string. If defined, its value is expected to follow the form described in the [dj-database-url docs](https://github.com/kennethreitz/dj-database-url#url-schema), e.g.: `mysql://username:password@192.168.1.20:3306/storage_service`. If undefined, Storage Service defaults to the `django.db.backends.sqlite3` [engine](https://docs.djangoproject.com/en/1.8/ref/settings/#engine) and expects the following environment variables to be defined:
 
 - **`SS_DB_NAME`**:
@@ -260,17 +265,17 @@ These variables specify the behaviour of LDAP authentication. If `SS_LDAP_AUTHEN
     - **Description:** LDAP "bind DN"; the object to authenticate against the LDAP server with, in order
     to lookup users, e.g. "cn=admin,dc=example,dc=com".  Empty string for anonymous.
     - **Type:** `string`
-    - **Default:** ``
+    - **Default:** `''`
 
 - **`AUTH_LDAP_BIND_PASSWORD`**:
     - **Description:** Password for the LDAP bind DN.
     - **Type:** `string`
-    - **Default:** ``
+    - **Default:** `''`
 
 - **`AUTH_LDAP_USER_SEARCH_BASE_DN`**:
     - **Description:** Base LDAP DN for user search, e.g. "ou=users,dc=example,dc=com".
     - **Type:** `string`
-    - **Default:** ``
+    - **Default:** `''`
 
 - **`AUTH_LDAP_USER_SEARCH_BASE_FILTERSTR`**:
     - **Description:** Filter for identifying LDAP user objects, e.g. "(uid=%(user)s)". The `%(user)s`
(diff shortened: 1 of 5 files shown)


install/README.md
```diff
@@ -470,31 +475,88 @@ These variables can be set to allow AWS authentication for S3 storage spaces as
...
+This is largely the same with external libraries, however, increasing their
+logging level can make available more information that isn't output by choice
+in the storage service's modules. Take for example the S3 Boto3 adapter.
+Logging can be changed from INFO to DEBUG to reveal more detailed information
+about a file transfer:
+
+```json
+    "boto3.*": {"level": "INFO"}, becomes "boto3.*": {"level": "DEBUG"},
+    "botocore.*": {"level": "INFO"}, becomes "botocore.*": {"level": "DEBUG"},
+```
```
sevein, line 539: I propose two alternative ways to show this snippet.

1. Using `diff`:

```diff
diff --git a/install/storageService.logging.json b/install/storageService.logging.json
index 5680b2a..2f85fd1 100644
--- a/install/storageService.logging.json
+++ b/install/storageService.logging.json
@@ -73,4 +73,4 @@
     },
-    "boto3": {"level": "INFO"},
-    "botocore": {"level": "INFO"}
+    "boto3": {"level": "DEBUG"},
+    "botocore": {"level": "DEBUG"}
   },
```

2. Using `js` instead of `json` so you can use comments:

```js
"boto3": {"level": "INFO"}, // becomes "boto3.*": {"level": "DEBUG"},
"botocore": {"level": "INFO"}, // becomes "botocore.*": {"level": "DEBUG"},
```

install/README.md
```diff
@@ -470,31 +475,88 @@ These variables can be set to allow AWS authentication for S3 storage spaces as
...
+
+Debug logging should never be used on a production server without the full
+implications being understood. The Boto3 developers, for example, ask users
+to [heed their warning][boto3].
+
+```plain
+    Warning: Be aware that when logging anything from 'botocore' the
+    full wire trace will appear in your logs. If your payloads contain
+    sensitive data this should not be used in production.
+```
```
sevein: Not sure if it's intentional, but typically [fenced code blocks](https://spec.commonmark.org/0.29/#fenced-code-blocks) don't need to be indented since they're rendered with `<pre><code>...</code></pre>`.

storage_service/storage_service/settings/components/s3.py
```diff
@@ -0,0 +1,13 @@
...
+"""
+from os import environ
+
+# Read and connect timeouts for S3. Ideally these will match the
+# defaults recommended by your S3 implementation.
+S3_TIMEOUTS = 900
+try:
+    S3_TIMEOUTS = int(environ.get("SS_S3_TIMEOUTS", S3_TIMEOUTS))
+except ValueError:
+    pass
```
sevein: When it comes to config the preferred pattern is fail-fast, i.e. report errors and exit, so the administrator receives feedback right away. We inherit that model from Django but it's pretty common in other systems, e.g. Django raises `django.core.exceptions.ImproperlyConfigured` (we do too from `get_env_variable`).
An alternative is to leverage Django's [system checks framework](https://docs.djangoproject.com/en/1.8/topics/checks/), but I think that's a solution we haven't explored yet.

sevein (commented):
This is excellent! I've added a few minor comments.

install/README.md
ross-spencer: It is the style I adopted while editing. What are the impacts? Is there something concrete that you have a problem with? (markdownlint is fine with this for example - and it renders as expected).

ross-spencer: I've changed this to an inline quotation as that was closer to the intention.

storage_service/storage_service/settings/components/s3.py
ross-spencer: When you say the preferred pattern do you have a link to anything in our documentation that describes this? Or do we just work it out? It might seem like hard work @sevein but 3 years later these CRs are still painful-as because we haven't any of this stuff written down when it comes up. If you take anything away nit-picking the undocumented assumptions in our work is just not a good way to create the harmony needed to work on a code-base at scale. Believe it or not, I'm not actually trying to get this wrong the first time around.

The change has been made to raise improperly configured with a helpful message.

sevein (approved):
Thanks.

install/README.md
sevein: > and it renders as expected

image

I only meant the extra spaces, that I've highlighted in the screenshot above. I didn't know it was intentional but I see now you're on top of it, thanks. It's good.

storage_service/storage_service/settings/components/s3.py
sevein: No sorry, we don't have docs for that. I can refer to common software design patterns I see out there, say for example in Redis there is a `protected-mode` config directive that expects a boolean, and if you pass something else then you see something like the following:

```
*** FATAL CONFIG FILE ERROR (Redis 6.0.10) ***
Reading the configuration file, at line 87
>>> 'protected-mode 1'
argument must be 'yes' or 'no'
```

> If you take anything away nit-picking the undocumented assumptions in our work is just not a good way to create the harmony needed to work on a code-base at scale.

Noted, thanks.

install/README.md
```diff
@@ -484,22 +484,79 @@ These variables can be set to allow AWS authentication for S3 storage spaces as
...
+* WARNING.
+* ERROR.
+* CRITICAL.
+
+The [Python documentation][python-docs] provides greater explanation.
+
+Though best efforts are taken to include the most useful information for
+debugging as possible. Your mileage may vary in debugging Archivematica or the
+Storage Service depending on the way the developer has written any particular
+module.
```
tw4l: I'd make this one sentence - the first sentence reads more like a fragment than a full sentence at present.

storage_service/storage_service/settings/components/s3.py
```diff
@@ -4,10 +4,13 @@
...
+from django.core.exceptions import ImproperlyConfigured
+
 # Read and connect timeouts for S3. Ideally these will match the
 # defaults recommended by your S3 implementation.
 S3_TIMEOUTS = 900
 try:
     S3_TIMEOUTS = int(environ.get("SS_S3_TIMEOUTS", S3_TIMEOUTS))
 except ValueError:
-    pass
+    err_msg = "S3 timeout value configured incorrectly in the environment please check the 'S3_TIMEOUTS' variable"
```
tw4l: Feel free to ignore but adding a hyphen or other separator between "environment" and "please" here might help with the readability of the error message

tw4l (approved):
I left a few minor suggestions but LGTM :muscle: Nice work! The README changes are especially nice.

PR #561, opened 2021-01-25
Make password validation UX tweaks

tw4l:
This commit resolves two issues identified during client testing of the password validation feature:

- Ensure that user attribute similarity validation works for all user attributes, not just username, by doing password validation in `UserCreationForm` after `self.instance` is updated with validated form data.
- Ensure user attributes do not disappear on edit user screen after an invalid password change attempt. This is a long-standing UX issue but was made more visible during QA for the password validation feature.

I also cleaned up the docstrings for the `UserCreationForm` and `UserChangeForm`.

Connected to archivematica/issues#1332

### DIFF
--- a/storage_service/administration/forms.py
+++ b/storage_service/administration/forms.py
@@ -2,6 +2,8 @@ from __future__ import absolute_import
 
 from django import forms
 from django.contrib import auth
+from django.contrib.auth.password_validation import validate_password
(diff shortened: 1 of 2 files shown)


ross-spencer (approved):
LGTM :+1:

tw4l:
Thanks @ross-spencer !

PR #562, opened 2021-01-26
Add audit log middleware

tw4l:
This commit adds middleware which adds an `X-Username` header with the name of the authenticated user to each response sent from the Storage Service. This provides a mechanism for audit logging of user behaviour via the nginx access logs.

Connected to https://github.com/archivematica/Issues/issues/1341

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -120,6 +120,11 @@ of these settings or provide values to mandatory fields.
     - **Type:** `boolean`
     - **Default:** `false`
 
+- **`SS_AUDIT_LOG_MIDDLEWARE`**:
+    - **Description:** enable X-Username header with authenticated HTTP responses.
+    - **Type:** `boolean`
+    - **Default:** `false`
+
 The configuration of the database is also declared via environment variables. Storage Service looks up the `SS_DB_URL` environment string. If defined, its value is expected to follow the form described in the [dj-database-url docs](https://github.com/kennethreitz/dj-database-url#url-schema), e.g.: `mysql://username:password@192.168.1.20:3306/storage_service`. If undefined, Storage Service defaults to the `django.db.backends.sqlite3` [engine](https://docs.djangoproject.com/en/1.8/ref/settings/#engine) and expects the following environment variables to be defined:
 
 - **`SS_DB_NAME`**:
--- a/storage_service/common/middleware.py
+++ b/storage_service/common/middleware.py
@@ -16,6 +16,19 @@ if hasattr(settings, "LOGIN_EXEMPT_URLS"):
     EXEMPT_URLS += [compile(expr) for expr in settings.LOGIN_EXEMPT_URLS]
 
 
+class AuditLogMiddleware(object):
+    """Add X-Username header with authenticated user to responses."""
+
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+        if request.user.is_authenticated:
+            response["X-Username"] = request.user.get_username()
+        return response
+
+
 class LoginRequiredMiddleware(MiddlewareMixin):
     """
     Middleware that requires a user to be authenticated to view any page other
--- a/storage_service/common/tests/test_middleware.py
+++ b/storage_service/common/tests/test_middleware.py
@@ -1,0 +1,38 @@
+from __future__ import absolute_import
+
+from django.conf import settings
+from django.contrib.auth import get_user_model
+from django.test import TestCase
+from django.test.client import Client
+
+
+class AuditLogMiddlewareTestCase(TestCase):
+    def setUp(self):
+        self.client = Client()
+        User = get_user_model()
+        self.user = User.objects.create_user(username="testclient", password="test")
+        self.client.force_login(self.user)
+
+    def test_audit_log_middleware_adds_username(self):
+        """Test that X-Username is added for authenticated users."""
+        with self.modify_settings(
+            MIDDLEWARE={"append": "common.middleware.AuditLogMiddleware"}
+        ):
+            response = self.client.get("/")
+            self.assertTrue(response.has_header("X-Username"))
+            self.assertEqual(response["X-Username"], self.user.username)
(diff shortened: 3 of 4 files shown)


storage_service/common/middleware.py
```diff
@@ -16,6 +16,36 @@
...
+    logs for auditing purposes.
+
+    To make use of this header in the nginx conf, add the following to
+    `log_format`:
+
+    user=$upstream_http_x_username
+
+    For more information on this middleware's use as part of an
+    Archivematica auditing solution, see the README for auditmatica:
+    https://github.com/artefactual-labs/auditmatica.
```
ross-spencer: I understand why this is here as it represents the current context and work effort but I'd like to suggest we tighten this up to just focus on just what the function does, i.e. adds the username header. And then we can rely on the docs proper to explain `auditmatica` and how it can potentially be used. Primarily because the header has broader application outside of nginx (and of course we might not always need to use nginx)

storage_service/common/middleware.py
```diff
@@ -16,6 +16,36 @@
...
+    def __init__(self, get_response):
+        self.get_response = get_response
+
+    def __call__(self, request):
+        response = self.get_response(request)
+
+        username = None
+        if request.user.is_authenticated:
+            username = request.user.get_username()
+        response["X-Username"] = username
```
ross-spencer:
```suggestion
        if request.user.is_authenticated:
            response["X-Username"] = username        
```
I guess this is both a suggestion and a question. If the user isn't authenticated, and the user isn't returned from get_username do we need to modify the header and send anything? You'll see if you do a `HEAD` request you'll get `None` in this field. Depending on your perspective this might be seen as unnecessary, but it might also be confusing. 

What do you think? 

(Although I see in your tests you're testing `None` explicitly on logout.)

storage_service/common/tests/test_middleware.py
```diff
@@ -0,0 +1,31 @@
...
+            self.assertEqual(response["X-Username"], self.user.username)
+
+    def test_audit_log_middleware_unauthenticated(self):
+        with self.modify_settings(
+            MIDDLEWARE={"append": "common.middleware.AuditLogMiddleware"}
+        ):
+            self.client.logout()
+            response = self.client.get(settings.LOGIN_URL)
+            self.assertTrue(response.has_header("X-Username"))
+            self.assertEqual(response["X-Username"], "None")
```
ross-spencer: Is the person logging out really `None`? Some test `docstrings` here for each of the functions would be nice in terms of understanding the test's intentions if that's possible? I feel like my question is a stupid one, but it doesn't seem unreasonable that it could come up again. Someone was doing the logging out after all. I am sure it is covered in the project use-cases.

storage_service/common/tests/test_middleware.py
```diff
@@ -0,0 +1,31 @@
+from __future__ import absolute_import
```
ross-spencer, line 1: `pylint --py3k` likes this, thank you!

ross-spencer (commented):
@tw4l fascinating PR in the context of what you're trying to achieve. It was nice to learn more about nginx's functionality. It's looking good, I've largely just asked for a clean-up of the doc string work here and addition of more context around the tests.

install/README.md
```diff
@@ -120,6 +120,11 @@ of these settings or provide values to mandatory fields.
     - **Type:** `boolean`
     - **Default:** `false`
 
+- **`SS_AUDIT_LOG_MIDDLEWARE`**:
+    - **Description:** enable logging of authenticated users with responses.
```
ross-spencer:
```suggestion
    - **Description:** enable X-Username header with authenticated HTTP responses.
```
Suggested text or similar? I feel like the context could be a little clearer. 

_(I don't know how this works as a second PR? - Just noticed it with the AM PR)_

storage_service/common/middleware.py
tw4l: Sounds good!

storage_service/common/middleware.py
tw4l: I want to do a bit of testing to make sure I understand the impact in the nginx log, but I think you're right that only adding the header if the user is authenticated is the better approach here.

storage_service/common/tests/test_middleware.py
tw4l: @ross-spencer Good question, and not unreasonable at all! You're right that the person logging out is not `None`, and that the username would be recorded in that case.

The `self.client.logout()` on line 28 here is basically test setup - ensuring that there is not an authenticated user for the next request on line 29, `response = self.client.get(settings.LOGIN_URL)`, which is the one we're actually testing. I'll see if I can re-structure the test and/or clarify in the docstring as you suggest.

install/README.md
tw4l: Nice suggestion! I'll make that change.

I'm not sure what you mean by "how this works as a second PR" - would you mind clarifying for me?

ross-spencer: > I'm not sure what you mean by "how this works as a second PR

Sorry. I meant CR. An afterthought submitted after the first CR as a second CR.

storage_service/common/tests/test_middleware.py
ross-spencer: > I'll see if I can re-structure the test and/or clarify in the docstring as you suggest.

A docstring should suffice.

install/README.md
tw4l: Ah, gotcha! Thanks for clarifying

tw4l:
Hi @ross-spencer, thanks for the suggestions! They were solid all around and I have made the changes. This should be ready for re-review.

install/README.md
ross-spencer: I suspect I make this mistake a lot in conversation too. Too similar and used too often. Anyhoo!

ross-spencer (approved):
LGTM! :+1:

install/README.md
tw4l: I'm sure I do too!

PR #564, opened 2021-02-01
WIP: S3 increased timeouts and config debugging (0.16)

ross-spencer:
Connected to archivematica/Issues#1346

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -99,6 +99,11 @@ of these settings or provide values to mandatory fields.
     - **Type:** `boolean`
     - **Default:** `false`
 
+- **`SS_S3_TIMEOUTS`**:
+    - **Description:** read and connect timeouts for S3 matching your implementation's recommended defaults.
+    - **Type:** `integer`
+    - **Default:** `900`
+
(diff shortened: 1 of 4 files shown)


sevein:
@ross-spencer, when it comes to logging, have you considered leveraging `LOGGING_CONFIG_FILE` instead of extending the configuration surface? We could document how to extend the [logging configuration file](https://github.com/artefactual/archivematica-storage-service/blob/stable/0.17.x/install/storageService.logging.json) and describe the different loggers that are available.

ross-spencer:
@sevein I haven't and I would need some convincing to go anywhere near that file. This logging is meant to be easy to turn on and off for a sys admin. And it does need to be off more than on as the code notes, there should be no leakage if at all possible. If it can be done easily with that file, then great, but some more context to your suggestion and reflection on how it can best be used would also be helpful.

sevein:
Ok, I can try to do that. All we have at the moment is [this section in the install README](https://github.com/artefactual/archivematica-storage-service/tree/stable/0.17.x/install#logging-configuration). It does not mention that it is based on Python's standard [configuration dictionary schema](https://docs.python.org/3/library/logging.config.html#configuration-dictionary-schema), so there is some room for improvement.

In the config document ([storageService.logging.json](https://github.com/artefactual/archivematica-storage-service/blob/stable/0.17.x/install/storageService.logging.json)), there is a section named `loggers`. We're already setting up some config for a number of known loggers, like `sword2`. Users can add config for additional loggers, say `requests` or `boto3`, `boto3.resources` and so on... It follows a hierarchy model. I think that this is similar to what you may find in other platforms, like Java's log4j.

Django adopts this model too, i.e. [`LOGGING_CONFIG` is just a dictConfig](https://docs.djangoproject.com/en/3.1/ref/settings/#std:setting-LOGGING_CONFIG).

I think it's a pretty good solution for sysadmins, this lets them own the logging configuration globally from a single place with a well-established config schema that is shared across many Python projects. So it could be that we don't need to write new code at all.

PR #565, opened 2021-02-02
Update test setup

replaceafill:
This:

- Sets a file based SQLite test database which mimics production setups better and prevents [issues with multiple connections](https://sqlite.org/inmemorydb.html).
- Uses absolute paths in the `checkformigrations` script so it can be called without `cd`ing into the `storage_service` subdirectory.
- Removes a redundant `pytest.ini` file. The settings in both files have been unified. Also, the `.coveragerc` file is now placed in the root of the repo.
- Sets default command line options for Pytest and allows to parametrize them when calling `tox` and the integration tests through [the `PYTEST_ADDOPTS` environment variable](https://docs.pytest.org/en/stable/example/simple.html#how-to-change-command-line-options-defaults). This is now used in the GitHub test workflows to set Pytest verbosity. 

Connected to https://github.com/archivematica/Issues/issues/1140


sevein (approved):
Thanks!

PR #566, opened 2021-02-02
Update python-gnupg and relax constraint

sevein:
We were some versions behind, in particular v0.4.4 addresses [CVE-2019-6690](https://bitbucket.org/vinay.sajip/python-gnupg/src/master/README.rst).

It needs to be tested.

Connects to https://github.com/archivematica/Issues/issues/1317.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,12 +1,12 @@
 repos:
 - repo: https://github.com/ambv/black
-  rev: 18.9b0
+  rev: 20.8b1
   hooks:
   - id: black
     args: [--safe, --quiet]
     language_version: python3
 - repo: https://gitlab.com/pycqa/flake8
-  rev: 3.7.7
+  rev: 3.8.4
   hooks:
   - id: flake8
     language_version: python3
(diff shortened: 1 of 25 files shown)


replaceafill:
I tested this and found an issue when I run the `aip-encrypt` and `aip-encrypt-mirror` tags of the acceptance tests where the GPG space is never created. I also noticed the `/administration/keys/` view raises `ValueError: gnupghome should be a directory (it isn't): /var/archivematica/storage_service/.gnupg` which doesn't seem familiar:

Screenshot_2021-02-16 ValueError at administration keys

I couldn't find anything in the AM or SS code that creates that directory, and then I looked at the [`python-gnupg` changelog](https://bitbucket.org/vinay.sajip/python-gnupg/src/master/#rst-header-change-log) which lead me to a related change in 0.4.5 where [the `.gnupg` directory is not created anymore](https://bitbucket.org/vinay.sajip/python-gnupg/issues/112/silent-fail-if-path-for-gnupg-home-isnt). [Someone reported a similar problem](https://bitbucket.org/vinay.sajip/python-gnupg/issues/122/gnupghome-folder-not-being-created-anymore) and the author answered with:

> With this change, the caller is responsible for ensuring that the directory exists beforehand.

replaceafill (approved):
Thanks!

PR #569, opened 2021-02-12
Allow replication in Duracloud spaces

replaceafill:
Connected to https://github.com/archivematica/Issues/issues/1350


mamedin (approved):
Thanks, it fixes the issue :)

PR #570, opened 2021-02-16
Fix disabled Purpose field choices

replaceafill:
This removes the deprecated `DisableableSelectWidget.render_option`
method that rendered `<option>` elements directly for the `Purpose`
field of the Location forms, and replaces it with a method that relies
on the new template-based widget rendering.

Connected to https://github.com/archivematica/Issues/issues/1424


sevein (approved):
:clap:

PR #573, opened 2021-03-01
Log traceback when the watchdog loop catches an exception

sevein:
For users relying on the Async API, the watchdog loop was only reporting warning messages when exceptions were detected. f89404a424b1eba3565a67d742b45e60d238e139 updates the logging entry so it includes the traceback (fixing https://github.com/archivematica/Issues/issues/742).

Additionally, this pull request (1) changes the logging of deferred tasks so it includes the traceback of the caller's frame, and (2) updates the watchdog loop so it reports currently running tasks using a gauge metric, removing unnecessary logging filling up our log streams.

### DIFF
--- a/storage_service/locations/metrics.py
+++ b/storage_service/locations/metrics.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+from __future__ import absolute_import, unicode_literals
+
+import time
+from contextlib import contextmanager
+
+from prometheus_client import Counter, Gauge
+
+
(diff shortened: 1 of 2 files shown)


replaceafill (approved):
Tested it and the added traceback and metrics look good :+1: Glad to see the multiple

```
archivematica-storage-service_1  | DEBUG     2021-03-02 08:20:10  locations.models.async_manager:async_manager:_watchdog_loop:123:  Watchdog sees 0 tasks running
```

messages gone!

PR #575, opened 2021-03-11
Use server-side processing for fixity checks table

replaceafill:
This extends the [DataTable](http://legacy.datatables.net/) server-side processing functionality added in https://github.com/artefactual/archivematica-storage-service/pull/470 to the Packages tables and applies it to the Fixity Checks view to sort entries by date.

I also enabled template debugging in the local settings to ease this kind of fixes.

Connected to https://github.com/archivematica/Issues/issues/1196

### DIFF
--- a/storage_service/locations/views.py
+++ b/storage_service/locations/views.py
@@ -80,10 +80,10 @@ def package_list(request):
 
 
 def package_list_ajax(request):
-    datatable = datatable_utils.DataTable(request.GET)
+    datatable = datatable_utils.PackageDataTable(request.GET)
     data = []
     csrf_token = get_token(request)
-    for package in datatable.packages:
+    for package in datatable.records:
         data.append(
             get_template("snippets/package_row.html")
             .render(
--- a/storage_service/locations/views.py
+++ b/storage_service/locations/views.py
@@ -112,8 +112,38 @@ def package_fixity(request, package_uuid):
     log_entries = FixityLog.objects.filter(package__uuid=package_uuid).order_by(
         "-datetime_reported"
     )
-    context = {"log_entries": log_entries}
+    context = {
+        "log_entries": log_entries,
+        "uri": request.build_absolute_uri("/"),
+        "package_uuid": package_uuid,
+    }
     return render(request, "locations/fixity_results.html", context)
+
+
+def fixity_logs_ajax(request):
+    datatable = datatable_utils.FixityLogDataTable(request.GET)
+    data = []
+    for fixity_log in datatable.records:
+        data.append(
+            get_template("snippets/fixity_log_row.html")
+            .render(
+                {
+                    "entry": fixity_log,
+                }
+            )
+            .strip()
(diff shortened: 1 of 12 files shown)


storage_service/static/js/project.js
```diff
@@ -1,7 +1,54 @@
-$(document).ready(function () {
+var getAjaxDataTableOptions = function (options, ajaxSource, headers, filter) {
+  var result = {};
+  for (var k in options) {
+    result[k] = options[k];
+  }
```
tw4l: Bit of a nit but I generally try to avoid single-letter variables. Maybe `key` would work here?

storage_service/static/js/project.js
```diff
@@ -1,7 +1,54 @@
...
+  // enable server-side processing
+  result["bServerSide"] = true;
+  result["bProcessing"] = true;
+  result["sAjaxSource"] = ajaxSource;
+  if (undefined !== filter) {
```
tw4l, line 10: I think this is the first Yoda condition I've seen in JS! I did a double-take but I think it's fine to keep as-is.

storage_service/locations/views.py
```diff
@@ -112,10 +112,40 @@ def package_fixity(request, package_uuid):
...
+            get_template("snippets/fixity_log_row.html")
+            .render(
+                {
+                    "entry": fixity_log,
+                }
+            )
+            .strip()
+        )
+    # these are the values that DataTables expects from the server
+    # see "Reply from the server" in http://legacy.datatables.net/usage/server-side
```
tw4l, line 137: Thanks, this comment/link is appreciated!

tw4l (approved):
I left a few very minor comments, but LGTM!

This might be out of scope (I'll defer to your judgment on that @replaceafill ) but while we're touching the Fixity tables page, the template CSS could use some tweaking to visually separate "Show x entries" and "Search" above the table in Firefox. At present they're bleeding into each other:

image

In Chrome/Chromium it looks better:

image

I'm happy to make an issue if we want to handle/note that separately.

Edit: Updated Firefox screenshot to show that the entire table isn't expanding to the page width.

replaceafill:
Thank you @tw4l :bow: 

> This might be out of scope (I'll defer to your judgment on that @replaceafill ) but while we're touching the Fixity tables page, the template CSS could use some tweaking to visually separate "Show x entries" and "Search" above the table in Firefox.

I just evaluated a few alternatives: extend the `fluid.html` template instead of `base.html`, add a `div.span12` container in the `fixity_results.html` template, add custom css just to that table.

As you suggest I think we should address it in a separate issue and revise the rest of the tables in UI if possible. Most of them have the same problem. See Administration -> Users for example:

Screenshot_2021-03-16 Archivematica Storage Service

In others like Pipelines I think the issue is not noticeable because there's a big paragraph above the table that makes the parent container to expand. The Locations table could probably use the fluid container too.

PR #576, opened 2021-03-12
Add eslint precommit hook

replaceafill:
This adds [ESLint](https://eslint.org/) as a pre-commit hook to automatically format (with [Prettier](https://prettier.io/)) and lint our JS files in `storage_service/static/js`.

Connected to https://github.com/archivematica/Issues/issues/1196


sevein (approved):
Nice!

PR #579, opened 2021-03-17
Revert tastypie permissions change for reading resource lists and details through API

tw4l:
This fixes a regression in the Storage Service API introduced by a change in `django-tastypie`.

Connects to https://github.com/archivematica/Issues/issues/1336.


sevein (approved):
Thank you!

Issue #258, opened 2017-11-06
Upgrade bagit, remove bagit workaround


Hwesta:
Bagit-python 1.6 has been released, which includes fix LibraryOfCongress/bagit-python#63  This means that https://github.com/artefactual/archivematica-storage-service/blob/1afb22aa7c05d1ba43dc531536b8e49dfe4c2da8/storage_service/locations/models/package.py#L1201  can be removed, since it was a workaround for that issue.

PR #582, opened 2021-03-22
Delete workaround for bagit-python#63

sevein:
Connects to https://github.com/artefactual/archivematica-storage-service/issues/258.


replaceafill (approved):
:exploding_head:

PR #585, opened 2021-03-23
Prevent modifications to pointer file fixture

tw4l:
This PR modifies the test `test_replicate_aic` to use a temporary pointer file to prevent changes to the test fixture.

Connected to https://github.com/archivematica/Issues/issues/1264

### DIFF
--- a/storage_service/locations/tests/test_package.py
+++ b/storage_service/locations/tests/test_package.py
@@ -612,13 +612,19 @@ class TestPackage(TestCase):
             purpose=models.Location.REPLICATOR,
         )
         assert aic.replicas.count() == 0
-        with mock.patch.object(
-            models.Package,
-            "full_pointer_file_path",
+
+        def create_temporary_pointer(path):
+            """Create temporary copy of pointer file at path."""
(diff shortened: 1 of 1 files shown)


storage_service/locations/tests/test_package.py
```diff
@@ -612,13 +612,20 @@ def test_replicate_aic(self):
             purpose=models.Location.REPLICATOR,
         )
         assert aic.replicas.count() == 0
-        with mock.patch.object(
-            models.Package,
-            "full_pointer_file_path",
+
+        def create_temporary_pointer(path):
+            """Create temporary copy of pointer file at path."""
+            temp_dir = tempfile.gettempdir()
```
replaceafill: I notice this class already provides `self.tmp_dir` which is deleted on `tearDown`. Could that be used here?

replaceafill (approved):
LGTM :+1:

storage_service/locations/tests/test_package.py
tw4l: Yes, thanks!

PR #590, opened 2021-03-30
Track and delete temporary directories for remote packages in Storage Service internal location

tw4l:
Connected to https://github.com/archivematica/Issues/issues/431

This PR modifies the `Package` model to keep track of temporary directories created in the Storage Service's internal location for packages which are fetched remotely via `Package.fetch_local_path()`. It then adds a method to clear these local directories, and calls that method at the end of the package download, fixity check, and reingest API methods, just prior to serving the response.

Functionally this means that temporary directories containing a copy of remote packages are no longer left behind in the Storage Service after downloading, reingesting, or checking fixity for packages stored in remote locations such as S3. I initially tried an approach of deleting tempdirs closer to their time of creation but that quickly became a web of side effects, and settled on the approach in this PR.

Incidentally, this change also helps with https://github.com/archivematica/Issues/issues/1112 by clearing temporary directories previously left behind when downloading, reingesting, or fixity checking uncompressed AIPs in GPG-encrypted AIP storage locations. Compressed encrypted AIPs are still leaving behind copies of the compressed package (seemingly 1 copy when downloading, 2 when checking fixity). I think these are being created and not cleaned up by the GPG Space's [`move_to_storage_service`](https://github.com/artefactual/archivematica-storage-service/blob/63f74d3314e29bd34d2844567047ecd248a5471c/storage_service/locations/models/gpg.py#L90-L133) method but think that might be better addressed in a separate PR.

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -189,6 +189,7 @@ class Package(models.Model):
         self.local_path = None
         self.local_path_location = None
         self.origin_location = None
+        self.local_tempdirs = []
 
     def __str__(self):
         return u"{uuid}: {path}".format(uuid=self.uuid, path=self.full_path)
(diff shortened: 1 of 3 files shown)


storage_service/locations/models/package.py
```diff
@@ -352,8 +353,27 @@ def fetch_local_path(self):
...
+                try:
+                    shutil.rmtree(tempdir)
+                    self.local_tempdirs.remove(tempdir)
+                    LOGGER.debug(
+                        "Deleted Storage Service internal tempdir: {}".format(tempdir)
+                    )
+                except OSError as e:
+                    LOGGER.debug(
+                        "Error deleting Storage Service internal tempdir: {}".format(
+                            str(e)
```
sevein: I believe str() isn't needed.

storage_service/locations/models/package.py
```diff
@@ -189,6 +189,7 @@ def __init__(self, *args, **kwargs):
         self.local_path = None
         self.local_path_location = None
         self.origin_location = None
+        self.local_tempdirs = []
```
sevein, line 192: This could be a set if you wanted to ensure unique values? In practice I don't think it'll make a difference.

sevein (approved):
We've discussed internally and decided that this would be a good initial solution until we have time to explore other ideas like having the caller provide the temporary resource or implement a context manager. Thanks Tessa!

PR #592, opened 2021-04-21
Avoid config reset in pipeline edit form

sevein:
When a pipeline is being edited, "create_default_locations" defaults to
True. This is often misleading because it forces the pipeline to
re-create the default locations.

This commit changes the pipeline view so the checkbox appears enabled by
default when creating a new pipeline but not when editing an existing
one.

Connects to https://github.com/archivematica/Issues/issues/1107.

### DIFF
--- a/storage_service/locations/views.py
+++ b/storage_service/locations/views.py
@@ -538,7 +538,10 @@ def pipeline_edit(request, uuid=None):
     else:
         action = _("Create Pipeline")
         pipeline = None
-        initial = {"enabled": not utils.get_setting("pipelines_disabled")}
+        initial = {
+            "create_default_locations": True,
+            "enabled": not utils.get_setting("pipelines_disabled"),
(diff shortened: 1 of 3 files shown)


storage_service/locations/views.py
```diff
@@ -549,7 +552,15 @@ def pipeline_edit(request, uuid=None):
             return redirect("locations:pipeline_list")
     else:
         form = forms.PipelineForm(instance=pipeline, initial=initial)
-    return render(request, "locations/pipeline_form.html", locals())
```
replaceafill, line 552: Thanks for making this `locals()` pattern explicit :bow:

replaceafill (approved):

storage_service/locations/views.py
sevein: 26 left :laughing:

PR #593, opened 2021-04-26
Add OfflineReplicaStaging Space

tw4l:

This PR adds a new `OfflineReplicaStaging` Space, which is a write-only Space used solely for replicas. Packages are stored to a directory on the local filesystem to facilitate pickup by offline storage devices such as tape robots. Uncompressed packages are packaged as a tarball prior to storage.

Because this is a write-only space intended for staging to external storage systems, it is not possible by design to download, delete, or check fixity for replicas stored in this Space.

### DIFF
--- a/storage_service/common/utils.py
+++ b/storage_service/common/utils.py
@@ -11,6 +11,7 @@ import mimetypes
 import os
 import shutil
 import subprocess
+import tarfile
 import uuid
 
 import scandir
@@ -76,7 +77,9 @@ COMPRESS_EXTENSIONS = (
     COMPRESS_EXTENSION_GZIP,
 )
 
-PACKAGE_EXTENSIONS = (".tar",) + COMPRESS_EXTENSIONS
+TAR_EXTENSION = ".tar"
+
+PACKAGE_EXTENSIONS = (TAR_EXTENSION,) + COMPRESS_EXTENSIONS
 
 COMPRESS_PROGRAM_7Z = "7-Zip"
 COMPRESS_PROGRAM_TAR = "tar"
@@ -348,7 +351,7 @@ def get_compress_command(compression, extract_path, basename, full_path):
         `compressed_filename` is the full path to the compressed file
     """
     if compression in (COMPRESSION_TAR, COMPRESSION_TAR_BZIP2, COMPRESSION_TAR_GZIP):
-        compressed_filename = os.path.join(extract_path, basename + ".tar")
+        compressed_filename = os.path.join(extract_path, basename + TAR_EXTENSION)
         relative_path = os.path.dirname(full_path)
         algo = ""
         if compression == COMPRESSION_TAR_BZIP2:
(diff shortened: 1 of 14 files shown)


storage_service/locations/models/replica_staging.py
```diff
@@ -0,0 +1,71 @@
...
+        )
+
+    def move_from_storage_service(self, src_path, dest_path, package=None):
+        """ Moves self.staging_path/src_path to dest_path."""
+        self.space.create_local_directory(dest_path)
+        if not package.is_packaged(src_path):
+            try:
+                return self._store_tar_replica(src_path, dest_path, package)
+            except utils.TARException:
+                raise
```
sevein: We could get rid of this try/except if we're just going to raise the error.

storage_service/locations/tests/test_package.py
```diff
@@ -709,6 +709,74 @@ def test_replicate_aip_gpg_encrypted(self, mock_encrypt):
         ]
         self._test_bagit_structure(replica, replication_dir)
 
+    def test_replicate_aip_offline_staging_uncompressed(self):
```
sevein, line 712: `test_replicate_aip_offline_staging_uncompressed` and `test_replicate_aip_offline_staging_compressed` are very similar.
Have you considered using `pytest.mark.parametrize`?

storage_service/locations/tests/test_replica_staging.py
```diff
@@ -0,0 +1,57 @@
...
+FIXTURES_DIR = os.path.abspath(os.path.join(THIS_DIR, "..", "fixtures"))
+
+
+class TestOfflineReplicaStaging(TestCase):
+
+    fixtures = ["base.json", "replica_staging.json"]
+
+    def setUp(self):
+        self.replica = models.Package.objects.get(id=1)
+        space_dir = tempfile.mkdtemp()
```
sevein: TempDirMixin can do this for you, it takes care of the clean-up. See `TestArkivum` for an example.

storage_service/common/utils.py
```diff
@@ -553,6 +556,91 @@ def set_compression_transforms(aip, compression, transform_order):
...
+
+    try:
+        if extension:
+            assert tarfile.is_tarfile(tarpath)
+            assert not os.path.exists(path)
+        else:
+            assert tarfile.is_tarfile(path)
+            assert not os.path.exists(tarpath)
+    except AssertionError:
+        _abort_create_tar(path, tarpath)
```
sevein: I know this is old code but I was wondering if we want to take this chance to improve it. We should probably avoid `assert`, there is the risk that is stripped away e.g. when using PYTHONOPTIMIZE. Instead, we could just write:

```python
if not tarfile.is_tarfile(tarpath if extension else path):
    raise TARException("...")

if not os.path.exists(path if extension else tarpath):
    raise TARException("...")
```

storage_service/common/utils.py
```diff
@@ -553,6 +556,91 @@ def set_compression_transforms(aip, compression, transform_order):
...
+    pass
+
+
+def _abort_create_tar(path, tarpath):
+    fail_msg = _(
+        "Failed to create a tarfile at %(tarpath)s for dir at %(path)s"
+        % {"tarpath": tarpath, "path": path}
+    )
+    LOGGER.error(fail_msg)
+    raise TARException(fail_msg)
```
sevein: I suggest to delete this function, the level of indirection is undesirable. We could stop logging errors as long as we're conveniently raising the exceptions since errors are also logged in the upper layers.

storage_service/common/utils.py
```diff
@@ -553,6 +556,91 @@ def set_compression_transforms(aip, compression, transform_order):
...
+        _abort_create_tar(path, tarpath)
+
+
+def _abort_extract_tar(tarpath, newtarpath, err):
+    fail_msg = _(
+        "Failed to extract %(tarpath)s: %(error)s" % {"tarpath": tarpath, "error": err}
+    )
+    LOGGER.error(fail_msg)
+    os.rename(newtarpath, tarpath)
+    raise TARException(fail_msg)
```
sevein: My comments on _abort_create_tar apply to this function too.

sevein (commented):
Hi @tw4l. I haven't had a chance to test yet but I'm sharing some initial feedback. Some thoughts that I've shared have to do with old code that it's just being moved to a new file, but perhaps we can take the chance to fix it.

storage_service/locations/models/replica_staging.py
```diff
@@ -0,0 +1,71 @@
...
+        self.space.move_rsync(src_path, dest_path)
+
+    def _store_tar_replica(self, src_path, dest_path, package):
```
sevein, line 56: Should this be a @staticmethod?

sevein (approved):
Looks good! I've tested and I was able to create replicators using the new space. I made some minor comments.

I've found that the browser in the location form isn't telling the user that the browsing functionality isn't available. That can be confusing, but that was already true for at least in dspace.py.

You may want to reword your attribution using [Co-authored-by](https://docs.github.com/es/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors).

storage_service/locations/models/replica_staging.py
tw4l: I was thinking having access to `self` is useful for `self.space.move_rsync(tar_src_path, tar_dest_path)` but otherwise it could be a `@staticmethod`

storage_service/locations/tests/test_package.py
tw4l: I wanted to use `pytest.mark.parametrize` but it's [not supported within Django's `TestCase` class](https://docs.pytest.org/en/latest/how-to/unittest.html#pytest-features-in-unittest-testcase-subclasses). I preferred some duplication to having to recreate the `setUp` fixture loading independently for this test.

tw4l: I'm fine with changing this if preferred or if I'm missing something though!

tw4l:
Thanks @sevein ! I've made some of the changes and left a few comments if you want to take a last look. Good call on cleaning up the TAR utilities while moving that code and thanks for the pointers.

storage_service/locations/tests/test_package.py
sevein: :man_facepalming: you're right!

sevein (approved):
Thank you!

ablwr:
@tw4l For this feature, since users cannot download AIPs from the write-only replica, can we hide the "Download" option in the Storage Service GUI? Otherwise, seems to be working fine across replicas.

PR #595, opened 2021-05-06
Remove download/delete links for offline replicas

tw4l:
Connected to https://github.com/archivematica/Issues/issues/1440

This PR removes action (download and delete) links from the Packages table for replicas stored in an `OfflineReplicaStaging` Space, based on feedback from internal QA.

Example row for a replica with this commit:

image

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -274,6 +274,43 @@ class Package(models.Model):
             raise StorageException(message)
 
     @property
+    def is_downloadable(self):
+        """Determines whether or not package is downloadable."""
+        space_is_offline_replica = (
+            self.current_location.space.access_protocol == Space.OFFLINE_REPLICA_STAGING
+        )
+        return self.status != self.DELETED and not space_is_offline_replica
+
+    @property
+    def is_deletable(self):
+        """Determines whether or not package is deletable."""
+        space_is_offline_replica = (
+            self.current_location.space.access_protocol == Space.OFFLINE_REPLICA_STAGING
+        )
+        return (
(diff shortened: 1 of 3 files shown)


ablwr (approved):
I think this looks fine, although it'd be good for @sevein to weigh in. Like, is it the case that no replica should ever be able to be downloaded from the AMSS? Or just ones using the write-only space?

sevein:
I don't really know the answer to your question, @sromkey probably knows!
What I'd like to suggest if you still have time @tw4l is to create a model property getter (downlodeable?) so the logic sits just in one place.

sromkey:
I would say it's better to have the download link available but to have an explanatory error message, OR grey-out the download link if it's a write-only space. This is really making me wish we weren't treating replicas as their own package entry at all, but that's an issue for another time...

tw4l:
@sevein @sromkey  I've moved the logic checks into `Package` model properties, but based on Sarah's comment above maybe it's best to just close this PR? The existing behaviour is that clicking on the Download or Delete links for offline staged replicas returns a Not Implemented Error in the browser.

sromkey:
"Not implemented" error sounds fine to be. I'm just thinking to someone having a mix of replicated and not replicated AIPs, it's going to be confusing to have the download link on some and not others. Does that make sense? Sorry for all the to-ing and fro-ing on this one haha! @ablwr let me know if you think I'm out to lunch!!

ablwr:
I defer to you @sromkey -- you are the project/product/program owner/manager! :laughing: 

OK sorry for the trouble @tw4l -- this feature is sufficient as-is!

PR #597, opened 2021-05-18
Add new arguments to create_aip_replicas

tw4l:

This commit adds `--aip-uuid` and `--replicator-location` arguments to the `create_aip_replicas` management command. These allow administrators to specify the UUID of the AIP to replicate, or the UUID of the Replicator location in which replicas should be created, respectively.

To help disambiguate between the script's arguments, the existing `--location` argument has also been renamed to `--aip-store-location`.


sevein (approved):
:+1:

PR #601, opened 2021-06-01
Cleanup tar replica left in staging directory

tw4l:

The initial implementation of the `OfflineReplicaStaging` Space left an extra copy of the tar replica in the Space's staging directory, which was causing disk space issues. This commit fixes the issue and adds an assertion to the test for replicating uncompressed packages in the new Space to ensure that extra copies of the new tar replica aren't left behind in staging.


scollazo (approved):
Looks good to me, LGTM!

sevein (approved):

PR #603, opened 2021-06-02
Cleanup quad dirs left in staging directory

tw4l:

The initial implementation of the `OfflineReplicaStaging` Space left behind empty quad dirs in the Space's staging directory. This commit fixes the issue and adds an assertion to the test for replicating uncompressed packages in the new Space to ensure that the empty quad dirs are cleaned up properly.


sevein (approved):
:+1:

PR #608, opened 2021-06-14
Enable flat output for Write-Only Replica Staging

tw4l:

This commit modifies the Write-Only Replica Staging space to write replicas into a flat directory, rather than Archivematica's typical UUID quad directories. This enables more efficient pickup of replicas stored in this space by offline storage systems.


sevein (approved):
:+1:

PR #610, opened 2021-06-15
Minimize leftovers on disk for Write-Only Replica Staging

tw4l:

These small tweaks will help not to leave leftover copies of the AIP around after replication.

### DIFF
--- a/storage_service/locations/models/replica_staging.py
+++ b/storage_service/locations/models/replica_staging.py
@@ -59,8 +59,13 @@ class OfflineReplicaStaging(models.Model):
         self.space.create_local_directory(dest_path)
         if not package.is_packaged(src_path):
             return self._store_tar_replica(src_path, dest_path, package)
-        self.space.move_rsync(src_path, dest_path)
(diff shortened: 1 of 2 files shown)


tw4l:
Apologies for the additional commit after submitting for PR - I noticed another leftover and wanted to make sure I was being thorough and covering all bases here. This is now properly ready for review.

storage_service/locations/models/replica_staging.py
```diff
@@ -59,9 +59,14 @@ def move_from_storage_service(
         self.space.create_local_directory(dest_path)
         if not package.is_packaged(src_path):
             return self._store_tar_replica(src_path, dest_path, package)
-        self.space.move_rsync(src_path, dest_path)
+        self.space.move_rsync(src_path, dest_path, try_mv_local=True)
```
replaceafill, line 62: TIL about `try_mv_local` and now I understand where those log messages come from :sweat_smile:

replaceafill (approved):

PR #612, opened 2021-06-24
Lock default locations creation on startup

replaceafill:
This adds a MySQL exclusive lock to the startup process which prevents
multiple gunicorn workers from creating different default locations at
once.

Co-authored-by: Jesús García Crespo <jesus@sevein.com>

Connected to https://github.com/archivematica/Issues/issues/1463

### DIFF
--- a/storage_service/administration/models.py
+++ b/storage_service/administration/models.py
@@ -1,10 +1,12 @@
 from __future__ import absolute_import
 from django.db import models
+import six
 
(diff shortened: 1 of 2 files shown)


sevein (approved):
Looks good. I've tested locally using `docker-compose up -d --scale archivematica-storage-service=3` and I couldn't reproduce the error again.

PR #615, opened 2021-06-30
Fix coerce_str helper in Python 3

replaceafill:
Connected to https://github.com/archivematica/Issues/issues/1472


sevein (approved):
Thank you! :pray: We ran some tests and it went very well, successfully delivered AIPs to DuraCloud. There was a tiny change needed, that I've included - see 00f87d1.

replaceafill:
Thanks @sevein! This looks OK after testing Duracloud chunking :+1:

PR #618, opened 2021-08-05
Don't stage encrypted AIPs twice, refs #1112

djjuhasz:
Connects to https://github.com/archivematica/Issues/issues/1112

Update `self.current_location` to the temporary staging directory after
copying and decrypting the package.  This is necessary so subsequent
calls to `fetch_local_path()` check if the current (staging) Space is
encrypted (via `is_encrypted()`), not whether the AIPsStore Space
is encrypted.  If `fetch_local_path()` thinks the package is still
encrypted it will copy it to a new temp dir and decrypt it again.

Also:
- Copy encrypted AIPs directly from storage location to temporary
  staging directory and decrypt there
- Add a debug log message to `get_download_path()`

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -309,6 +309,7 @@ class Package(models.Model):
         else:  # LOCKSS AU number specified, but not a LOCKSS package
             LOGGER.warning("Trying to download LOCKSS chunk for a non-LOCKSS package.")
             path = full_path
+        LOGGER.debug("Got download path {} for package {}".format(path, self.uuid))
         return path
 
     def get_local_path(self):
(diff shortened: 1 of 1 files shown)


sevein (commented):
> Update self.current_location to the temporary staging directory after
copying and decrypting the package. This is necessary so subsequent
calls to fetch_local_path() check if the current (staging) Space is
encrypted (via is_encrypted()), not whether the AIPsStore Space
is encrypted. If fetch_local_path() thinks the package is still
encrypted it will copy it to a new temp dir and decrypt it again.

David, in regards to this specific comment, I'm trying to reproduce the scenario that you describe but I couldn't figure out the context where subsequent calls to fetch_local_path() are happening within the same web request. How could I reproduce?

djjuhasz:
> David, in regards to this specific comment, I'm trying to reproduce the scenario that you describe but I couldn't figure out the context where subsequent calls to fetch_local_path() are happening within the same web request. How could I reproduce?

@sevein I don't remember the exact details, as I haven't looked at the code for almost two months, but I followed the "Steps to reproduce" from Ashley's bug report https://github.com/archivematica/Issues/issues/1112 to reproduce the bug and test the fix.

sevein (approved):
LGTM.

sevein:
31729ca1d852d394b4f5168dba47bd9ca257670b and 056f71231c30766dcc253820da6bea6a0929a929 in stable/0.18.x

PR #619, opened 2021-08-18
Add .zip as an acceptable compress extension

fozboz:
This remediates https://github.com/archivematica/Issues/issues/1481 but I'm unsure if this is the right approach.

The method that causes the issue was added in [this commit](https://github.com/artefactual/archivematica-storage-service/commit/f35d3620d0edc156520762ba06882aa7d573fb67) which relates specifically to AIP storage, not Transfer Source. But currently both the Transfer Source and AIP storage functions are using the same `s3.py` script, which is beholden to this list of acceptable compression extensions.

If `.zip` cannot be added as an acceptable extension for an AIP, then perhaps we need a separate script (or separate list) for using S3 as a Transfer Source.


sevein (approved):
Thanks, I've tested and it works. Glad that you figured it out.

> If .zip cannot be added as an acceptable extension for an AIP, then perhaps we need a separate script (or separate list) for using S3 as a Transfer Source.

I think it's fine. Archivematica doesn't give users that option but it's not a problem if it's added to the space (S3) as far as I know.

sevein:
c3924e0 in stable/0.18.x

PR #623, opened 2021-10-18
Use os.makedirs to build tree in extract_file

sevein:
Since baaede6b3, `is_compressed` returns `True` when expected. This forces
`extract_file_request` to call `extract_file` to retrieve the file. This method
was failing because `os.mkdir` is not able to create directories recursively.
This commit uses `os.makedirs` instead.

I've tested with encrypted and non-encrypted AIPs (compressed and uncompressed).

Connects to https://github.com/archivematica/Issues/issues/1112.


replaceafill (approved):

sevein:
0fb9d0a in stable/0.18.x

PR #625, opened 2021-11-01
Add permission checks

sevein:
This pull request introduces the concept of user roles:
* **Administrators** - with the `is_superuser` flag enabled,
* **Managers** - a Django group with all application permissions given,
* **Reviewers** - a Django group with the `approve_package_deletion` permission),
* **Readers** - any user that is authenticated; read-only access.

The new `roles` module provides some functions to associate these roles to application users. A new database migration ensures that the new groups are created and have the permissions linked, as well as migrating existing users so they are grante dthe Managers role automatically.

Permission checks have been added where needed, including the API. In particular, the package deletion approve/reject workflow is available for Reviewers, Managers and Administrators. Certain user interface elements are hidden from the user when they're not actionable due to lack of permissions, hence the updates in multiple templates.

Our supported authentication backends have been revisited to ensure that the migration to this new permission-role model is possible. To ensure backward compatibility, a new setting (`SS_AUTH_DEFAULT_USER_ROLE="reader"`) has been added to automatically promote users to managers that would become read-only users otherwise, in the absence of proper configuration to match other roles. Additionally:
* **LDAP**: the application will attempt to discover group membership in the directory, where the group names are "Administrators", "Managers" and "Reviewers" (see [example](https://gist.github.com/sevein/7cbd262c1141de3ad220a326f734032f)) but they're also configurable,
* **Shibboleth**: in addition to the existing `preservation-admin` entitlement, we've added `preservation-manager` and `preservation-reviewer`,
* **CAS**: in addition to the existing `AUTH_CAS_ADMIN_ATTRIBUTE{,_VALUE}` vars, new settings `CAS_MANAGER_ATTRIBUTE{,_VALUE}` and `CAS_REVIEWER_ATTRIBUTE{,_VALUE}` have been added,
* **OIDC**: all authenticated users will become readers - use `SS_AUTH_DEFAULT_USER_ROLE` if you want your users to a role with more privileges. Group membership could be identified via claims in the future.

Connects to https://github.com/archivematica/Issues/issues/1486.
Build status: [see last commit in Archivematica branch](https://github.com/artefactual/archivematica/commits/dev/issue-1485-permissions-take-two).


install/README.md
```diff
@@ -93,6 +93,11 @@ of these settings or provide values to mandatory fields.
     - **Type:** `boolean`
     - **Default:** `false`
 
+- **`SS_AUTH_DEFAULT_USER_ROLE`**:
+    - **Description:** user role that must be assigned to authenticated users that would be readers otherwise based on the results given by the authentication backend. The default is "manager" to match the old behavior in Storage Service where authenticated user was promoted as a manager. Other options are "reviewer" and "reader" (or empty string).
+    - **Type:** `string`
+    - **Default:** `"reader"`
```
replaceafill, line 99: The **Description** still says the default is `manager`.

storage_service/administration/forms.py
```diff
@@ -257,40 +267,78 @@ def _post_clean(self):
             except ValidationError as error:
                 self.add_error("password1", error)
 
+    def save(self, commit=True):
+        user = super().save(commit)
+        user.set_role(self.cleaned_data["role"])
+
+
+class RoleChoiceField(forms.ChoiceField):
+    """This is a ... """
```
replaceafill: Suspense! :joy:

storage_service/administration/forms.py
```diff
@@ -257,40 +267,78 @@ def _post_clean(self):
...
+    widget = RadioSelectWidget
+
 
 class UserChangeForm(auth.forms.UserChangeForm):
     """Modifies an existing user. Inherits from Django's UserChangeForm."""
 
+    role = RoleChoiceField(
+        required=True,
+        choices=roles.USER_ROLES,
+        help_text=_(
```
replaceafill: Could the help text string be extracted to a module variable and used in both fields (create and edit)?

storage_service/administration/forms.py
```diff
@@ -257,40 +267,78 @@ def _post_clean(self):
...
     def clean(self):
         """Validate the form to protect against potential user errors."""
         if self.superusers.count() > 1:
             return self.cleaned_data
         try:
-            # Protect field from being reverted if only one superuser.
-            if self.user_being_edited.is_superuser:
-                self.cleaned_data["is_superuser"] = True
+            # An administration can only change its role with the help of
+            # another administrator.
```
replaceafill: I have trouble matching this comment with the code below since the `if` seems to be checking the context of the form and not who's performing the edit :thinking:

storage_service/administration/templates/widgets/role_choice.html
```diff
@@ -0,0 +1,11 @@
+{% load i18n %}
+
+{% include "django/forms/widgets/multiple_input.html" %}
+
+{% if widget.attrs.disabled %}
+  {% if user.is_superuser %}
```
replaceafill: It seems `user` is not being passed in the widget context so this condition is always false. I'm not sure of your intention with the block but does it make sense for a "superuser" to see both conditional messages?

storage_service/administration/tests/test_users.py
```diff
@@ -0,0 +1,126 @@
...
+from django.contrib.auth.models import User
+from django.test import TestCase
+from django.urls import reverse
+
+from administration import roles
+
+from unittest import skipIf
+
+
+class TestUserManagement(TestCase):
```
replaceafill, line 11: Just out of curiosity, why did you go with `TestCase` subclassing here instead of pytest based functions like in `test_roles.py`?

storage_service/common/backends.py
```diff
@@ -21,31 +23,44 @@ def configure_user(self, user):
...
 class CustomOIDCBackend(OIDCAuthenticationBackend):
-    """
-    Provide OpenID Connect authentication
-    """
+    """Provide OpenID Connect authentication."""
 
     def get_userinfo(self, access_token, id_token, verified_id):
+        """Extract user details from JSON web tokens.
+
+        It returns two values. The first value is a dict of user details that
```
replaceafill: Is the `returns two values` part still accurate? I only see one dict in the `return` statement below.

storage_service/locations/api/resources.py
```diff
@@ -616,7 +638,7 @@ def _handle_location_file_move(self, move_files_fn, request, *args, **kwargs):
 
         return move_files_fn(data["files"], origin_location, destination_location)
 
-    @_custom_endpoint(expected_methods=["post"])
+    @_custom_endpoint(expected_methods=["post"], required_permission="change_detail")
```
replaceafill: Should this be `required_permission="update_detail"`?

storage_service/locations/tests/test_api.py
```diff
@@ -107,12 +108,21 @@ class TestLocationAPI(TestCase):
...
+        self.user.set_password("test")
         self.client.defaults["HTTP_AUTHORIZATION"] = "Basic " + base64.b64encode(
             b"test:test"
         ).decode("utf8")
 
+    def as_reader(self):
+        self.user.set_role(roles.USER_ROLE_READER)
+
+    def as_reviewer(self):
+        self.user.set_role(roles.USER_ROLE_MANAGER)
```
replaceafill:
```suggestion
        self.user.set_role(roles.USER_ROLE_REVIEWER)
```

storage_service/locations/tests/test_api.py
```diff
@@ -704,24 +794,30 @@ class TestPipelineAPI(TestCase):
...
```
replaceafill:
```suggestion
        self.user.set_role(roles.USER_ROLE_REVIEWER)
```

storage_service/locations/views.py
```diff
@@ -91,6 +92,7 @@ def package_list_ajax(request):
                     "package": package,
                     "redirect_path": request.META.get("HTTP_REFERER", request.path),
                     "csrf_token": csrf_token,
+                    "perms": PermWrapper(request.user),
```
replaceafill, line 95: :exploding_head:

storage_service/administration/views.py
```diff
@@ -27,7 +28,14 @@
 # ######################## ADMIN ##########################
 
 
-def settings_edit(request):
+def index(request):
+    if request.user.is_superuser:
```
replaceafill: While testing this I noticed `Managers` can access the `Configuration` view but this redirects them to `Users` instead.

storage_service/templates/snippets/locations_table.html
```diff
@@ -51,7 +53,9 @@
         <td>{{ loc.used }}B / {{ loc.quota|default:_("unlimited") }}</td>
         <td>{{ loc.enabled|yesno:_("Enabled,Disabled") }}</td>
         <td>{{ loc.default|yesno:_("Yes,No") }}</td>
-        <td><a href="{% url 'locations:location_edit' loc.space.uuid loc.uuid %}">{% trans "Edit" %}</a> | <a href="{% url 'locations:location_switch_enabled' loc.uuid %}?next={{ request.path }}">{{ loc.enabled|yesno:_("Disable,Enable") }}</a> | <a href="{% url 'locations:location_delete' loc.uuid %}?next={{ request.path }}">{% trans "Delete" %}</a></td>
+        {% if perms.locations.change_location %}
```
replaceafill: Should the `Delete` link contained here be protected with the `locations.delete_location` permission?

storage_service/templates/snippets/pipelines_table.html
```diff
@@ -14,7 +16,9 @@
         <td><a href="{% url 'locations:pipeline_detail' pipeline.uuid %}">{{ pipeline.uuid }}</a></td>
         <td>{{ pipeline.description }}</td>
         <td>{{ pipeline.enabled|yesno:_("Enabled,Disabled") }}</td>
-        <td><a href="{% url 'locations:pipeline_edit' pipeline.uuid %}">{% trans "Edit" %}</a> | <a href="{% url 'locations:pipeline_switch_enabled' pipeline.uuid %}?next={{ request.path }}">{{ pipeline.enabled|yesno:_("Disable,Enable") }}</a> | <a href="{% url 'locations:pipeline_delete' pipeline.uuid %}?next={{ request.path }}">{% trans "Delete" %}</a></td>
+        {% if perms.locations.change_pipeline %}
```
replaceafill: Should the `Delete` link contained here be protected with the `locations.delete_pipeline` permission?

replaceafill (approved):
This looks great to me @sevein! I just left some minor observations/questions. Feel free to address them as convenient.

storage_service/administration/tests/test_users.py
sevein: Ouch, sorry about that. I didn't even realize, :facepalm:

storage_service/administration/views.py
sevein: Good catch, I'll use this instead:

    request.user.has_perm("administration.change_settings")

storage_service/locations/api/resources.py
sevein: Thank you! :smile:

PR #626, opened 2021-11-12
Add openssh-client to Dockerfile

scollazo:
This allows the use of Pipeline Local Filesystem spaces

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,7 @@ RUN set -ex \
 		libsasl2-dev \
 		locales \
 		locales-all \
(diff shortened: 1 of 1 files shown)


mamedin (approved):

replaceafill (commented):
Would it be worth to add the dependency to `osdeps` too?

scollazo:
I don't think adding it to osdeps will bring anything, ssh is almost mandatory on vm's, and adding more packages to osdeps means waiting a bit more for them to be checked/installed , so in my opinion, I prefer to skip it.

PR #633, opened 2022-04-20
Send mimetype when uploading object to s3 and DuraCloud

fozboz:
And sends the locally calculated checksum when uploading to DuraCloud.

Creates a new util method for returning mimetype, so we don't repeat code.

I noticed that `dspace.py` also uses `mimetypes`. It could re-use the `get_mimetype` method introduced here, but I wanted to confirm something with the maintainers first.

In `dspace.py` there is this line:

```
"Content-Type": str(mimetypes.guess_type(upload_path)),
```

which effectively means the Content-Type header being sent to dspace is `('video/mp4', None)`. Is that intentional? I would have thought only `video/mp4` is necessary.

Connected to https://github.com/archivematica/Issues/issues/1550

### DIFF
--- a/storage_service/common/tests/test_utils.py
+++ b/storage_service/common/tests/test_utils.py
@@ -468,3 +468,9 @@ def test_create_tar(
 )
 def test_strip_quad_dirs_from_path(input_path, expected_path):
     assert utils.strip_quad_dirs_from_path(input_path) == expected_path
+
+def test_get_mimetype():
+    assert utils.get_mimetype('video.mp4') == 'video/mp4'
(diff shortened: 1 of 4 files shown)


replaceafill:
Thank you @fozboz! I rebased your commit on top of the current `qa/0.x` branch. It's been merged in https://github.com/artefactual/archivematica-storage-service/commit/7a18357e51497105e158bac2594ed5ba7b044121.

Regarding your question about `dspace.py`, it also makes sense to me to just send the type and not the encoding guessed by `mimetypes.guess_type`. But I haven't been able to confirm that since I don't have a `DSpace` instance to test with. Feel free to send a follow up PR.

PR #632, opened 2022-03-23
Write pointer file in binary mode

fozboz:
The `etree.tostring()` function generates bytes, not a string. Therefore, the file to be written must be opened in binary mode, otherwise this error is thrown.

`TypeError: write() argument must be str, not bytes`

I believe this would have worked with python2. python3 enforces this distinction.

Resolves https://github.com/archivematica/Issues/issues/1545

PR #634, opened 2022-07-08
delete source path after moving

fitnycdigitalinitiatives:
Piggybacking off of #632, even if the pointer file is successfully recreated, I don't think 'move' is working fully as intended because it doesn't delete the original package in the source after the move. Though, I'm not sure if that was the intent to leave the original there. In the SS api, it uses the term 'relocate' so I'm assuming that the intent is to remove the original files in the source. This uses 'delete_path' to do so.

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -489,6 +489,8 @@ class Package(models.Model):
         source_path = os.path.join(
             self.current_location.relative_path, self.current_path
         )
+        #get full path of origin for delete_path
+        source_delete_path = self.full_path
 
         destination_path = os.path.join(to_location.relative_path, self.current_path)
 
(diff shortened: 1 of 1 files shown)


eviau-artefactual:
I will try to start the tests by closing/re-opening the branch...

Dhwaniartefact:
@fitnycdigitalinitiatives :Thank you for the contribution!

Based on the results of the tests, it appears that the linting test failed. Would it be possible for you to fix the linting issue ? There is a [pre-commit](https://pre-commit.com/) setup you can create. It should be done pre-commit install inside the git repo.

Thanks!

fitnycdigitalinitiatives:
Ok, let me see when I can get to this, it's been a while since I worked on this.

PR #635, opened 2022-08-09
Add rclone Space to Storage Service

tw4l:
Connected to https://github.com/archivematica/Issues/issues/1567

This PR introduces a new `RClone` space to the Storage Service, which uses [rclone](https://rclone.org) to provide support for use of virtually any cloud storage provider for AIP store, DIP store, Replicator, and Transfer Source locations. This includes but is not limited to:

- Amazon S3 and other S3 API-compatible object stores
- Microsoft Azure Blob Storage
- Openstack Swift (e.g. OVH object storage)
- Backblaze B2
- Google Drive
- Box
- Dropbox
- Microsoft OneDrive

Configuration of access keys and other details other than the remote name and container/bucket name is done entirely through [environment variables](https://rclone.org/docs/#environment-variables).

There is overlap in functionality with the existing `S3` and `Swift` models, which could potentially live alongside or be deprecated in favor of the `RClone` Space in time.

Note: Use of `lsjson` for the `browse` method requires rclone v1.42 or newer, which means it may be better to install the latest binary (as in the Dockerfile example in this PR) rather than relying on the default version in package managers. As of writing this, `apt install rclone` in Ubuntu 18.04 installs version v1.36.


storage_service/locations/models/space.py
```diff
@@ -145,12 +145,13 @@ class Space(models.Model):
...
     OFFLINE_REPLICA_STAGING = "REPLICA"
     PIPELINE_LOCAL_FS = "PIPE_FS"
+    RCLONE = "RCLONE"
     SWIFT = "SWIFT"
     GPG = "GPG"
     S3 = "S3"
     # These will not be displayed in the Space Create GUI (see locations/forms.py)
     BETA_PROTOCOLS = {}
-    OBJECT_STORAGE = {DATAVERSE, DSPACE, DSPACE_REST, DURACLOUD, SWIFT, S3}
+    OBJECT_STORAGE = {DATAVERSE, DSPACE, DSPACE_REST, DURACLOUD, SWIFT, S3, RCLONE}
```
scollazo: Should rclone be before Swift, to keep alphabetical order here?

tw4l: yes, good call!

storage_service/locations/models/rclone.py
```diff
@@ -0,0 +1,239 @@
...
+        """
+        LOGGER.debug("Test that container '%s' exists", self.container)
+        prefixed_container_name = "{}{}".format(self.remote_prefix, self.container)
+        cmd = ["ls", prefixed_container_name]
+        _, stderr = self._execute_subprocess(cmd)
+        if stderr:
+            LOGGER.info("Creating container '%s'", self.container)
+            create_container_cmd = ["mkdir", prefixed_container_name]
+            _, stderr = self._execute_subprocess(create_container_cmd)
+            if stderr:
```
scollazo: Should we raise the exception based on the [return code](https://rclone.org/docs/#list-of-exit-codes) of rclone, instead of stderr?

When the config file doesn't exist , seems like a NOTICE is logged trough stderr, thus breaking the process.

tw4l: Yeah that makes sense to me! I'll make that change today.

tw4l: Change made in commit https://github.com/artefactual/archivematica-storage-service/pull/635/commits/5a17501b25d10845c6e0f655b706756d0713198c

storage_service/locations/models/rclone.py
```diff
@@ -56,17 +56,21 @@ def _execute_subprocess(self, subcommand):
...
-                LOGGER.debug("rclone stdout: %s", stdout)
-                if stderr:
-                    LOGGER.warning(stderr)
+            LOGGER.debug("rclone cmd: %s", cmd)
+            LOGGER.debug("rclone stdout: %s", stdout)
+            if stderr:
+                LOGGER.warning("rclone stderr: %s", stderr)
 
-                return stdout, stderr
+            if proc.returncode != 0:
```
scollazo: Should we configure a retry system for retryable errorrs (Return code 5)?

tw4l: Nice suggestion! Commit https://github.com/artefactual/archivematica-storage-service/pull/635/commits/f75110b8e8cbac92b271d50f1fe2dc4b306137c8 introduces up to 5 retries, with two seconds in between each

[edit: fixed commit reference]

scollazo (approved):
It looks really good, congrats @tw4l !

scollazo:
@replaceafill what do you think?

tw4l:
@replaceafill quick note to say that I created an issue to link this PR to. I'm also planning to squash the Django migrations, but left it as a series of small commits to help with code review :)

Dockerfile
```diff
@@ -24,8 +24,17 @@ RUN set -ex \
 		p7zip-full \
 		rsync \
 		unar \
+		unzip \
 	&& rm -rf /var/lib/apt/lists/*
 
+# Install rclone
+RUN set -ex \
+	cd $(mktemp -d) && \
+ 	curl -OfsS "https://downloads.rclone.org/rclone-current-linux-amd64.zip" \
```
replaceafill: Would it be worth to add an `ARG` instruction that allows you to set the Rclone version? Similar to [`PYTHON_VERSION` in `a3m`](https://github.com/artefactual-labs/a3m/blob/3056896549885eb9b9c4fdbce796deb57c7eab22/Dockerfile#L88). It could default to `1.59.1`.

Dockerfile
```diff
@@ -24,8 +24,17 @@ RUN set -ex \
 		p7zip-full \
 		rsync \
 		unar \
+		unzip \
 	&& rm -rf /var/lib/apt/lists/*
 
+# Install rclone
+RUN set -ex \
+	cd $(mktemp -d) && \
```
replaceafill: Starting each command with `&&` in the `RUN` instruction might help edits/diffs on future modifications.

storage_service/locations/models/rclone.py
```diff
@@ -0,0 +1,265 @@
...
+    MAX_RETRIES = 5
+
+    ALLOWED_LOCATION_PURPOSE = [
+        Location.AIP_STORAGE,
+        Location.DIP_STORAGE,
+        Location.REPLICATOR,
+        Location.TRANSFER_SOURCE,
+    ]
+
+    def _execute_subprocess(self, subcommand):
```
replaceafill:
```python
         cmd = ["rclone"] + subcommand
```

Based on its name and the way the `subcommand` argument is used I wonder if this function (and docstring) should be modified to indicate it executes `Rclone` **subcommands** (e.g. `_execute_rclone_subcommand`). The subprocess part in the name seems like an implementation detail to me.

storage_service/locations/models/rclone.py
```diff
@@ -0,0 +1,265 @@
...
+                )
+                (stdout, stderr) = proc.communicate()
+
+                LOGGER.debug("rclone cmd: %s", cmd)
+                LOGGER.debug("rclone stdout: %s", stdout)
+                if stderr:
+                    LOGGER.warning("rclone stderr: %s", stderr)
+
+                # Return code of 5 from rclone indicates retriable error.
+                if proc.returncode == 5:
```
replaceafill: Is it worth to encapsulate the retriable return codes into a tuple in case we want to handle more that one in future versions of `Rclone`? If not maybe turn return code 5 into a constant like `MAX_RETRIES`?

storage_service/locations/models/rclone.py
```diff
@@ -0,0 +1,265 @@
...
+            5 attempts at retriable errors.
+        """
+        cmd = ["rclone"] + subcommand
+        attempt = 0
+        while attempt < self.MAX_RETRIES:
+            try:
+                proc = subprocess.Popen(
+                    cmd, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+                )
+                (stdout, stderr) = proc.communicate()
```
replaceafill, line 75: If `stdout` and `stderr` are decoded into strings as early as possible here, the callers of the function don't need to decode each time and you won't get `b'...'` strings in logger and error messages. Alternatively you could pass `Popen(..., universal_newlines=True)` to [get the streams result in text mode](https://docs.python.org/3.6/library/subprocess.html#frequently-used-arguments).

storage_service/locations/models/rclone.py
```diff
@@ -0,0 +1,265 @@
...
+
+    @property
+    def remote_prefix(self):
+        """Return remote prefix from env vars matching case-insensitive RClone.remote_name."""
+        remotes = self._execute_subprocess(["listremotes"])
+        LOGGER.debug("rclone listremotes output: %s", remotes)
+        remotes = six.ensure_str(remotes).split("\n")
+
+        for remote in remotes:
+            if remote.startswith(self.remote_name.lower()):
```
replaceafill: What's the reason for the `startswith` approach? If my space `remote_name` is set to `aips-zipped` I think this could return a wrong remote like for example `aips-replicas` depending on the sorting returned by `listremotes`?

storage_service/locations/models/rclone.py
```diff
@@ -0,0 +1,265 @@
...
+            "delete",
+            "{}{}{}".format(self.remote_prefix, container, delete_path.lstrip("/")),
+        ]
+        self._execute_subprocess(cmd)
+
+    def move_to_storage_service(self, src_path, dest_path, dest_space):
+        """ Moves src_path to dest_space.staging_path/dest_path. """
+        # strip leading slash on src_path
+        src_path = src_path.rstrip(".")
+        src_path = src_path.lstrip("/")
```
replaceafill:
```suggestion
        src_path = src_path.rstrip(".").lstrip("/")
```

storage_service/locations/models/rclone.py
```diff
@@ -0,0 +1,265 @@
...
+        dest_path = dest_path.rstrip(".")
+
+        subcommand = "copyto"
+        if utils.package_is_file(src_path) is False:
```
replaceafill: Since `utils.package_is_file` always returns a boolean its calls might look more "idiomatic" written like:

```python
if not utils.package_is_file(src_path):
```

storage_service/locations/tests/test_rclone.py
```diff
@@ -0,0 +1,464 @@
...
+        assert sorted(return_value["entries"]) == [
+            "dir1",
+            "dir2",
+            "obj1.txt",
+            "obj2.mp4",
+        ]
+        assert return_value["properties"] == expected_properties
+    else:
+        with pytest.raises(models.StorageException):
+            _ = rclone_space.browse("/")
```
replaceafill: Why is the `_` necessary? Same below.

storage_service/locations/tests/test_rclone.py
```diff
@@ -0,0 +1,464 @@
...
+    return aip
+
+
+def test_rclone_delete(mocker, rclone_aip):
+    """Mock method call and assert correctness of rclone command."""
+    delete_path = mocker.patch("locations.models.rclone.RClone._execute_subprocess")
+    mocker.patch(
+        "locations.models.rclone.RClone.remote_prefix",
+        return_value="testremote:",
+        new_callable=mocker.PropertyMock,
```
replaceafill, line 104: TIL `PropertyMock` :open_mouth:

storage_service/locations/tests/test_rclone.py
```diff
@@ -0,0 +1,464 @@
...
+        # Package is file, expect "copyto" subcommand
+        (
+            True,
+            [
+                "copyto",
+                SRC_PATH,
+                "testremote:testcontainer/{}".format(DEST_PATH.lstrip("/")),
+            ],
+        ),
+        # Package is directory, expect "copy" subcommand
```
replaceafill, line 332: Although I appreciate the comments in each case, the use of a single `DEST_PATH` for `Package is file/directory` is a bit hard to follow. For example, `DEST_PATH` seems to be used the same in this test but one is a package and the other a directory?

Maybe you could use different constants for each case like the integration test does:

https://github.com/artefactual/archivematica-storage-service/blob/f75110b8e8cbac92b271d50f1fe2dc4b306137c8/integration/test_integration.py#L29-L34

storage_service/locations/tests/test_rclone.py
```diff
@@ -0,0 +1,464 @@
+from __future__ import absolute_import
```
replaceafill, line 1: I really appreciate the test coverage of the change :bow:

replaceafill (changes requested):
Really nice work @tw4l! I think my only concern is the use of `startswith` when getting the remote prefix. I'm not sure of its use case but I think it might give you the wrong remote under certain conditions? Let me know if I'm missing anything.

I've also found an intermittent failure with the `test_rclone_move_from_storage_service_no_container` test similar to the one shown in this [Integration job run](https://github.com/artefactual/archivematica-storage-service/runs/7844712515?check_suite_focus=true). I think in that case the base Space class tries to call `os.makedirs` which raises a permission error. If switching the tests to use temporary dirs/paths is too difficult, maybe just mocking `makedirs` would do it?

I'm not sure if you were planning to add integration tests, but since the integration tests Compose environment already has a `minio` container available, I created https://github.com/artefactual/archivematica-storage-service/commit/724acd1366da9b0e3a5f6bfd69a81cff2e20aecb with some of them just to get familiar with the environment variables needed for the feature. I hope it's useful to you.

storage_service/locations/tests/test_rclone.py
tw4l: It's not! Thanks for pointing it out

storage_service/locations/tests/test_rclone.py
tw4l: This one took me a while to figure out! :laughing:

storage_service/locations/models/rclone.py
tw4l: The idea was to make it so that it'll match regardless of whether someone puts e.g. `azure` or `azure:` in the Space config, but you're right that there's a better way to do that!

storage_service/locations/models/rclone.py
tw4l: I agree!

Dockerfile
tw4l: I think I'd rather keep this to `latest` since the only version dependency we have is >1.42, which will never be the latest.

tw4l:
Thank you for the review, @replaceafill! I've made the suggested changes, including improving the matching on remote name to be more exact. I've also squashed the migrations into one. Before merging, I'll do the same for the commits.

I've adopted your integration tests here (thanks for those!), and patched `Space.create_local_directory` in the unit tests to avoid the `OSError`. I've run the integration tests ~10-15 times now and haven't seen the intermittent failure you reported there. Have you seen it more than once, or just in the run you linked to?

tw4l:
It looks like in the [integration test run you linked to](https://github.com/artefactual/archivematica-storage-service/runs/7844712515?check_suite_focus=true), the unit tests were being run as well and that was what was failing. Is that expected?

integration/docker-compose.yml
```diff
@@ -8,13 +8,19 @@ services:
       dockerfile: "Dockerfile"
     entrypoint: ""
     working_dir: "/src"
-    command: ["./integration/wait-for-it.sh", "mysql:3306", "--timeout=30", "--", "pytest", "storage_service", "integration"]
+    command: ["./integration/wait-for-it.sh", "mysql:3306", "--timeout=30", "--", "pytest", "integration"]
```
replaceafill: > I've run the integration tests ~10-15 times now and haven't seen the intermittent failure you reported there. Have you seen it more than once, or just in the run you linked to?

Sorry for the confusion @tw4l. I removed `storage_service` here to avoid the issue on purpose. We need it back since this is the only way we have to run the SS tests under MySQL at the moment.

tw4l: ah okay, good to know!

replaceafill (approved):

PR #636, opened 2022-08-19
Add stored datetime to packages

tw4l:
Connected to https://github.com/archivematica/issues/issues/1529


tw4l:
@replaceafill another PR related to preservation storage that could use code review if you have some time :) thank you!

tw4l:
Rebased on latest `qa/0.x`

storage_service/locations/models/package.py
```diff
@@ -973,6 +981,7 @@ def _store_aip_to_uploaded(self, v, related_package_uuid):
                 related_package = Package.objects.get(uuid=related_package_uuid)
                 self.related_packages.add(related_package)
             self.status = Package.UPLOADED
+            self.stored = datetime.now()
```
replaceafill: Do you think the stored datetime should also be set in the `except` block below? I'm wondering specifically about `# 6. set the status to "uploaded" (if applicable),` where the package seems to be marked as `UPLOADED` too.

storage_service/locations/tests/test_package.py
```diff
@@ -287,6 +288,51 @@ def test_files_are_added_to_database(self):
             == "742f10b0-768a-4158-b255-94847a97c465"
         )
 
+    def test_stored_datetime(self):
+        package = models.Package.objects.get(
+            uuid="a59033c2-7fa7-41e2-9209-136f07174692"
+        )
+        assert package.stored is None
+
+        def mock_v():
```
replaceafill: This mock could use a docstring describing what `V` is. Even though it might not be very helpful I've noticed a couple of `Package` methods describe it as `object with attributes needed for processing.` :shrug:

storage_service/templates/snippets/packages_table.html
```diff
@@ -9,6 +9,7 @@
         <th class="sortable">{% trans "Type" %}</th>
         <th class="sortable">{% trans "Replica Of" %}</th>
         <th class="sortable">{% trans "Status" %}</th>
+        <th class="sortable">{% trans "Stored" %}</th>
```
replaceafill, line 12: This makes the DataTable column sortable in the frontend, but the backend class that handles the sorting needs to be updated too. The implementation is a bit clunky and not the most efficient (cough since I wrote it cough...), but the change should be:

1. Add a new column index to the `SORT_KEY_HELPERS_MAPPING` dictionary which value should be the name of a method that receives a `Package` instance and returns a sorting key for it. The new column index needs to be inserted before the fixity ones.
2. Add that named method to the class. As you can see in the `sort_by_fixity_date_key` method for example, for datetimes, since the attribute is nullable we need to consider that it's not possible to compare datetimes and `None`. In that particular case I decided to use `datetime.now`.

I've created https://github.com/artefactual/archivematica-storage-service/commit/4e5c7f42eea40543807dddfde090380ec0af3966 as an example of how that might look. I hope it's useful.

replaceafill (changes requested):
This looks great to me @tw4l and I think only the DataTable sorting needs to be addressed.

storage_service/locations/api/resources.py
```diff
@@ -772,6 +772,7 @@ class Meta:
             "package_type",
             "size",
             "status",
+            "stored",
```
scollazo: Can we use a different name here? "stored" is a reserved keyword since mysql 5.7.6 ( https://dev.mysql.com/doc/refman/5.7/en/keywords.html )

scollazo (changes requested):

storage_service/templates/snippets/packages_table.html
tw4l: that's so helpful, thank you! and I wouldn't call the implementation clunky :)

storage_service/locations/api/resources.py
replaceafill: Maybe `stored_at`?

tw4l: Good catch! Renaming it to "stored_date".

tw4l:
@replaceafill @scollazo this is now ready for re-review - thank you both!

replaceafill (approved):
Thanks for extending the test coverage! :bow:

scollazo (approved):
LGTM!

PR #637, opened 2022-08-26
Use TIME_ZONE from settings for AIP Package.stored_date values

tw4l:
Related to https://github.com/archivematica/Issues/issues/1529

This commit is a small follow-up to https://github.com/artefactual/archivematica-storage-service/pull/636, which makes the values saved to`Package.stored_date` for AIPs (including replicas) timezone-aware according to the Storage Service's `TIME_ZONE` setting.

For more on this, see: https://docs.djangoproject.com/en/1.11/topics/i18n/timezones/#time-zones

### DIFF
--- a/storage_service/locations/models/package.py
+++ b/storage_service/locations/models/package.py
@@ -5,7 +5,6 @@ from __future__ import absolute_import
 from collections import namedtuple
 import codecs
 import copy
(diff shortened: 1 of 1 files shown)


replaceafill (approved):
Thanks for taking care of it!

PR #638, opened 2022-08-26
Add management command to populate Package.stored_date for local filesystem AIPs

tw4l:
Related to https://github.com/archivematica/Issues/issues/1529

This PR introduces a new Storage Service management command, which will populate `Package.stored_date` for AIPs and AIP replicas in local filesystem AIPstore and Relicator locations based on the last modified dates on the filesystem.

This may be useful for clients upgrading to the eventual 0.20 release of the Storage Service that want to populate the `Package.stored_date` field to help with use cases such as reporting.

Note that because of filesystem permissions, this management command must be run as root or with sudo.

### DIFF
--- a/storage_service/common/management/commands/populate_aip_stored_dates.py
+++ b/storage_service/common/management/commands/populate_aip_stored_dates.py
@@ -1,0 +1,94 @@
+# -*- coding: utf-8 -*-
+"""Populate the Package.stored_date field for AIPs and AIP replicas.
+
+Currently this works only for AIPs and replicas that are stored in
+a local filesystem Space.
+
+AIPs that already have timestamps are ignored and not updated.
+
+Execution example:
+./manage.py populate_aip_stored_dates
+"""
+from __future__ import absolute_import, print_function
+from datetime import datetime
+import logging
+import os
+import pytz
+
+from django.conf import settings
+from django.core.management.base import CommandError
+
+from common.management.commands import StorageServiceCommand
+from locations.models.package import Package, Space
+
+# Suppress the logging from models/package.py.
+logging.config.dictConfig({"version": 1, "disable_existing_loggers": True})
+
+TIMEZONE = pytz.timezone(settings.TIME_ZONE)
+
+
+class Command(StorageServiceCommand):
+
+    help = __doc__
+
+    def add_arguments(self, parser):
+        """Entry point to add custom arguments"""
+        parser.add_argument(
+            "--location-uuid",
+            help="UUID for specific AIP Store or Replicator location",
+            default=None,
+        )
+
+    def handle(self, *args, **options):
+        aips = Package.objects.filter(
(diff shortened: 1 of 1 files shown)


storage_service/common/management/commands/populate_aip_stored_dates.py
```diff
@@ -0,0 +1,105 @@
...
+            status=Package.UPLOADED,
+        ).all()
+        if not aips:
+            raise CommandError("No AIPs with status UPLOADED found")
+
+        local_fs_aips = self.filter_aips_by_local_filesystem(aips)
+
+        location_uuid = options["location_uuid"]
+        if location_uuid:
+            local_fs_aips = self.filter_aips_by_location(aips, location_uuid)
```
sevein: Should we try to build a queryset with lookups that span relationships to do most if not all of the filtering in the database? If the queryset is built in advance then location_uuid could be added conditionally.

storage_service/common/management/commands/populate_aip_stored_dates.py
```diff
@@ -0,0 +1,105 @@
...
+        skipped_count = 0
+
+        for aip in local_fs_aips:
+            # Skip AIPs that already have datestamps.
+            if aip.stored_date is not None:
+                skipped_count += 1
+                continue
+            try:
+                modified_unix = os.path.getmtime(aip.full_path)
+                timezone = pytz.timezone(settings.TIME_ZONE)
```
sevein: This looks like it could be executed before we enter the loop.

storage_service/common/management/commands/populate_aip_stored_dates.py
```diff
@@ -0,0 +1,105 @@
...
+                aip.stored_date = datetime.fromtimestamp(
+                    int(modified_unix), tz=timezone
+                )
+                aip.save()
+                success_count += 1
+            except OSError as err:
```
sevein, line 69: It isn't clear where OSError might raise - is it fromtimestamp and getmtime?
Maybe for clarity I'd had the try block focused on the function call that raises that error.
save() would raise a django.db.DatabaseError I believe?

sevein (approved):
I haven't tested, but this looks fine! :+1: 
I've made a few small suggestions - small polishes if you have some extra time.

storage_service/common/management/commands/populate_aip_stored_dates.py
tw4l: yes, good call!

storage_service/common/management/commands/populate_aip_stored_dates.py
tw4l: It was from getmtime - this is mostly intended to catch permissions issues, e.g. if the command isn't run as root. I'll make that more explicit in the code, thanks!

tw4l:
thanks @sevein !!

PR #639, opened 2022-08-31
Store package checksum to verify we have the expected package during fixity checks

tw4l:
Related to https://github.com/archivematica/Issues/issues/1570

This commit adds checksum and checksum_algorithm fields to the Package model, which are populated during AIP storage. These are then used during fixity checks to ensure that not only does the Bag validate, but we are validating the expected package.

For compressed AIPs, we store the checksum of the package. For uncompressed AIPs, we store the checksum of the Bag's `tagmanifest` file.

This also enables faster fixity checks for compressed AIPs, by making it unnecessary to extract the compressed package in order to perform fixity.

As a side benefit, this should also solve https://github.com/archivematica/Issues/issues/920 for packages that have a `Package.checksum` value.

The PR also includes a new `populate_aip_checksums` management command, which can be used to populate these fields for compressed and uncompressed AIPs created using Storage Service version 0.19 or earlier. For compressed AIPs, the checksum for the compressed AIP is pulled from the pointer file. For uncompressed AIPs, a checksum of the Bag's `tagmanifest` file is computed. By default, this will only work for AIPs stored in a local filesystem. The command includes a `--download` flag which can be used to download each uncompressed AIP stored remotely (e.g. in S3) in order to calculate the `tagmanifest` checksum.


storage_service/common/utils.py
```diff
@@ -637,15 +661,33 @@ def generate_checksum(file_path, checksum_type="md5"):
...
     return checksum
 
 
+def find_tag_manifest(file_path):
+    """Return the path to a Bag's tagmanifest file."""
+    if not os.path.isdir(file_path):
+        return
+
+    for bag_file in os.listdir(file_path):
+        if bag_file.startswith("tagmanifest"):
```
scollazo: The spec says that [a bag may contain one or more tag manifest](https://www.rfc-editor.org/rfc/rfc8493.html#section-2.2.1), should we look for them following an order? (sha512, sha256, md5)?

scollazo (changes requested):

storage_service/common/utils.py
scollazo: Also, they should always end in ```.txt```

tw4l: great catch, thank you!

tw4l:
Running the black box Automated Tests reveals an issue with reingest, which currently fails on this branch. Some log output:

```
archivematica-storage-service_1  | INFO      2022-09-12 16:42:38  locations.models.package:package:extract_file:1650:  Extracting file with: ['7z', 'x', '-bd', '-y', '-o/var/archivematica/storage_service/tmp53xl3qya', '/var/archivematica/sharedDirectory/www/AIPsStore/f0ab/2c9e/16f2/454d/b07b/7a97/91ae/5930/amauat-transfer_1663013929-f0ab2c9e-16f2-454d-b07b-7a9791ae5930.7z'] to /var/archivematica/storage_service/tmp53xl3qya/amauat-transfer_1663013929-f0ab2c9e-16f2-454d-b07b-7a9791ae5930
archivematica-storage-service_1  | DEBUG     2022-09-12 16:42:38  locations.models.package:package:start_reingest:2253:  Reingest: extracted to /var/archivematica/storage_service/tmp53xl3qya/amauat-transfer_1663013929-f0ab2c9e-16f2-454d-b07b-7a9791ae5930
archivematica-storage-service_1  | ERROR     2022-09-12 16:42:38  locations.models.package:package:check_fixity:2002:  Package checksum 1ccce54804393a5014b42f372d9233a4ce5f83edd556fb17b7260d5338c41a34 for package f0ab2c9e-16f2-454d-b07b-7a9791ae5930 did not validate against expected value 4d09bc1d38efb7e00c1f5a580bd4503818cb10e399dadf5ee4145b2ae29016a5	
```

Calling the `/check_fixity/` endpoint for the same package succeeds, so I think it might be about the order of operations during reingest (unpacking before checking fixity). I will look into this closer tomorrow.,

tw4l:
Commit https://github.com/artefactual/archivematica-storage-service/pull/639/commits/c8442818d5d66c801f2f082d941768f9eec79e37 modifies `Package.start_reingest` to check a package's fixity prior to extracting it, which fixes reingest and causes the black box tests to once again pass:

```
7 features passed, 0 failed, 20 skipped
19 scenarios passed, 0 failed, 59 skipped
116 steps passed, 0 failed, 613 skipped, 0 undefined
Took 11m41.462s
```

I've checked elsewhere in the Storage Service where `Package.check_fixity()` is called and it seems that this was the only place that packages were extracted prior to running fixity, so I wouldn't expect to see this same side effect elsewhere.

To explain what was happening: Because the compressed AIP was extracted prior to running `Package.check_fixity()`, `check_fixity` was treating the package as uncompressed and computing a hash for the `tagmanifest` file rather than the package itself. When this didn't match the `Package.checksum` value, an error was thrown and reingest failed.

scollazo (approved):

tw4l:
Rebased on latest `qa/0.x`

PR #649, opened 2023-06-08
Development environment updates

replaceafill:
This makes the following changes to the Docker environment:

* Installs Python 3.6 using [`pyenv`](https://github.com/pyenv/pyenv).
* Migrates the old [Compose V1 configuration](https://docs.docker.com/compose/migrate/).
* Upgrades the dependencies and removes the `wait-for-it` script in the integration test.

Connected to https://github.com/archivematica/Issues/issues/1612


sevein (approved):
:+1:

PR #659, opened 2023-08-25
Adds Archipelago as a new space option in the storage service.

KidsSeeGhosts:
Closed, meant to open as draft, not ready for review. I still need to add documentation, instructions and images to the PR.

This PR adds Archipelago as a new space option in the storage service. It allows the user to enter the url, user and password for archipelago to create the space, and AIP Storage location can then be made for the space. AIPs are stored in Archipelago by the storage service creating JSON API requests, using the AIP title and UUID as the keys. The fields of the AIP are populated in archipelago by the storage service extracting the Dublin Core fields from the mets file and sending these as optional extra fields in the request.


DiegoPino:
@KidsSeeGhosts interesting!

PR #664, opened 2023-08-31
Remove Django warnings

Dhwaniartefact:
Remove Django 3.2 warnings created during "make bootstrap" in development environment.

Reference: https://docs.djangoproject.com/en/3.2/ref/settings/#std-setting-DEFAULT_AUTO_FIELD

### DIFF
--- a/storage_service/locations/apps.py
+++ b/storage_service/locations/apps.py
@@ -1,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class LocationsAppConfig(AppConfig):
+    default_auto_field = "django.db.models.AutoField"
+    name = "locations"
--- a/storage_service/locations/apps.py
+++ b/storage_service/locations/apps.py
@@ -4,3 +4,6 @@ from django.apps import AppConfig
 class LocationsAppConfig(AppConfig):
     default_auto_field = "django.db.models.AutoField"
     name = "locations"
+
+    def ready(self):
+        import locations.signals  # noqa: F401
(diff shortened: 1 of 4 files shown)


storage_service/locations/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
+
+
+class LocationsAppConfig(AppConfig):
+    default_auto_field = "django.db.models.AutoField"
+    name = "locations"
```
replaceafill, line 6: Now that this exists I think we can move [this import](https://github.com/artefactual/archivematica-storage-service/blob/6196e34ad20b66544e2426cd9be7551103efae99/storage_service/locations/models/__init__.py#L2-L6) here.

What do you think of creating a `ready` method that imports the signals like the `CommonAppConfig.ready` method does?

replaceafill (approved):
Looks fine @Dhwaniartefact, and I think we could improve the `LocationsAppConfig` now that it exists.

storage_service/locations/apps.py
Dhwaniartefact: Thanks for the suggestion. Can I ask you what is the reason behind this change? Is it because of testypie warnings?

replaceafill: From https://docs.djangoproject.com/en/3.2/ref/applications/#django.apps.AppConfig.ready

> Subclasses can override this method to perform initialization tasks such as registering signals.

That's exactly what the `import` in that `__init__` file is trying to do (see the comment block above the `import` line). But the `AppConfig` subclass is the appropriate place to do that. See all our other `AppConfig` subclasses in AM and SS. Some of them do exactly that.

replaceafill (approved):

storage_service/locations/apps.py
Dhwaniartefact: Thanks for your suggestion.

PR #667, opened 2023-09-13
Upgrade Django to 4.2

Dhwaniartefact:
This sets django-upgrade in pre-commit to upgrade Django to 4.2. 

A few changes are worth mentioning:

The USE_L10N setting is removed as localized formatting of data will always be enabled. Ref: https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-USE_L10N

STATICFILES_STORAGE: Static files storage engine can be configured with the [STORAGES] setting. Ref: https://docs.djangoproject.com/en/4.2/ref/settings/#std-setting-STORAGES

Connected to https://github.com/archivematica/Issues/issues/1624


storage_service/storage_service/settings/base.py
```diff
@@ -86,7 +86,6 @@ def _get_settings_from_file(path):
 USE_I18N = True
 
 # See: https://docs.djangoproject.com/en/dev/ref/settings/#use-l10n
```
replaceafill: Should we delete this line?

replaceafill (commented):
Looks like you still need to make new migrations? And could you please add Python 3.11 to the CI workflow?

storage_service/storage_service/settings/base.py
Dhwaniartefact: yes we can as per documentation this feature is added by default in Django 4.2

replaceafill:
This looks great @Dhwaniartefact, thanks! Let's leave this PR here until we milestone its issue.

replaceafill:
@Dhwaniartefact It seems you still need to clean some comments in the `base` settings module left from the settings you removed.

Dhwaniartefact:
Sure let me do it.

.github/workflows/test.yml
```diff
@@ -13,14 +13,10 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: [
-          "3.8",
-          "3.9",
-          "3.10",
-        ]
+        python-version: ["3.8", "3.9", "3.10","3.11"]
```
replaceafill: Could you turn these lists into multi-line again? It simplifies code reviews because you see one value being added or removed instead of the entire single-line list changing. Same with `ubuntu-version` and `python-version` below.

.github/workflows/test.yml
```diff
@@ -13,14 +13,10 @@ jobs:
...
     steps:
       - name: "Check out repository"
-        uses: "actions/checkout@v4"
+        uses: "actions/checkout@v3"
```
replaceafill: You're downgrading all the `checkout` actions. Could you revert them back?

replaceafill (approved):
Looks good @Dhwaniartefact, but I left a couple of small requests.

.github/workflows/test.yml
Dhwaniartefact: Thank you for suggestions. Let me work on it.

.github/workflows/test.yml
Dhwaniartefact: Done.

PR #669, opened 2023-09-27
Fix Ldap Group Type and roles

mamedin:
After SS roles commit, it was needing the LDAP_GROUP_SEARCH settings (getting 500 error). This commit allows to use thed default role when LDAP_GROUP_SEARCH variables are not set.

Added the AUTH_LDAP_GROUP_TYPE variable, that was fixed to ActiveDirectoryGroupType (it is the default value now).

Connects to https://github.com/archivematica/Issues/issues/1629

### DIFF
--- a/storage_service/common/signals.py
+++ b/storage_service/common/signals.py
@@ -111,13 +111,25 @@ def ldap_populate_user_profile(sender, user=None, ldap_user=None, **kwargs):
     if user.is_superuser:
         return
 
-    role = roles.USER_ROLE_READER
-    if settings.AUTH_LDAP_ADMIN_GROUP in ldap_user.group_names:
-        role = roles.USER_ROLE_ADMIN
-    elif settings.AUTH_LDAP_MANAGER_GROUP in ldap_user.group_names:
-        role = roles.USER_ROLE_MANAGER
-    elif settings.AUTH_LDAP_REVIEWER_GROUP in ldap_user.group_names:
-        role = roles.USER_ROLE_REVIEWER
+    role = roles.settings.DEFAULT_USER_ROLE
+
+    if hasattr(settings, "AUTH_LDAP_GROUP_SEARCH"):
+        LOGGER.debug(
+            "Using LDAP groups for user %s. LDAP Groups: %s",
+            user.username,
+            ldap_user.group_names,
+        )
+        if settings.AUTH_LDAP_ADMIN_GROUP in ldap_user.group_names:
+            role = roles.USER_ROLE_ADMIN
+        elif settings.AUTH_LDAP_MANAGER_GROUP in ldap_user.group_names:
+            role = roles.USER_ROLE_MANAGER
+        elif settings.AUTH_LDAP_REVIEWER_GROUP in ldap_user.group_names:
+            role = roles.USER_ROLE_REVIEWER
+    else:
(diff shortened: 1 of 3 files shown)


storage_service/common/signals.py
```diff
@@ -111,13 +111,25 @@ def ldap_populate_user_profile(sender, user=None, ldap_user=None, **kwargs):
...
         return
 
-    role = roles.USER_ROLE_READER
-    if settings.AUTH_LDAP_ADMIN_GROUP in ldap_user.group_names:
-        role = roles.USER_ROLE_ADMIN
-    elif settings.AUTH_LDAP_MANAGER_GROUP in ldap_user.group_names:
-        role = roles.USER_ROLE_MANAGER
-    elif settings.AUTH_LDAP_REVIEWER_GROUP in ldap_user.group_names:
-        role = roles.USER_ROLE_REVIEWER
+    role = roles.settings.DEFAULT_USER_ROLE
```
mamedin, line 114: Use the default role defined in environment variables instead of `reader`

storage_service/common/signals.py
```diff
@@ -111,13 +111,25 @@ def ldap_populate_user_profile(sender, user=None, ldap_user=None, **kwargs):
...
+
+    if hasattr(settings, "AUTH_LDAP_GROUP_SEARCH"):
+        LOGGER.debug(
+            "Using LDAP groups for user %s. LDAP Groups: %s",
```
mamedin, line 118: Not sure if it is safe to print all the ldap groups of the user in SS log, but it is useful when configuring ldap

storage_service/common/signals.py
```diff
@@ -111,13 +111,25 @@ def ldap_populate_user_profile(sender, user=None, ldap_user=None, **kwargs):
...
-    role = roles.USER_ROLE_READER
-    if settings.AUTH_LDAP_ADMIN_GROUP in ldap_user.group_names:
-        role = roles.USER_ROLE_ADMIN
-    elif settings.AUTH_LDAP_MANAGER_GROUP in ldap_user.group_names:
-        role = roles.USER_ROLE_MANAGER
-    elif settings.AUTH_LDAP_REVIEWER_GROUP in ldap_user.group_names:
-        role = roles.USER_ROLE_REVIEWER
+    role = roles.settings.DEFAULT_USER_ROLE
+
+    if hasattr(settings, "AUTH_LDAP_GROUP_SEARCH"):
```
mamedin, line 116: Adding a conditional to skip the following when not configuring the GROUP_SEARCH fixes the issue and makes optional the group settings.

storage_service/storage_service/settings/base.py
```diff
@@ -386,7 +386,15 @@ def _get_settings_from_file(path):
         )
 
     # https://pythonhosted.org/django-auth-ldap/groups.html#types-of-groups
-    AUTH_LDAP_GROUP_TYPE = ldap_config.ActiveDirectoryGroupType()
+    if "AUTH_LDAP_GROUP_TYPE" in environ:
```
mamedin, line 389: This is not part of the issue, but when testing the GROUP settings it is needed to use a different ldap schema than the Active Directory. Probably we need the same in the Dashboard LDAP settings.

sevein (approved):
Nicely done!

PR #681, opened 2024-01-30
Use tempfile library for temp file creation in package.py

camlyall:
Fixes use of hardcoded temp directory.
This is part of the fix for issue https://github.com/archivematica/Issues/issues/1425

Reviewed in SS 0.21.1 Am 1.15.1
- Transfer with compression
- Full Reingest

Custom temporary directory can be set in the storages service environment file. Example:
- `TMPDIR=/path/to/dir`


replaceafill (approved):
Thank you @camlyall.

PR #692, opened 2024-02-27
Update common to use pathlib instead of os.path

klavman:
Connects to https://github.com/archivematica/Issues/issues/1622

Remarks:
- There is no direct equivalent function like [`os.path.commonprefix`](https://github.com/artefactual/archivematica-storage-service/blob/a8aefbc84b605fad340e383c3a9f353caadc1e8b/storage_service/common/management/commands/import_aip.py#L209).
- [`os.path.join(source, "")`](https://github.com/artefactual/archivematica-storage-service/blob/a8aefbc84b605fad340e383c3a9f353caadc1e8b/storage_service/common/management/commands/import_aip.py#L315) has the behaviour an empty last part will result in a path that ends with a separator.  I have opted for [this equivalence](https://stackoverflow.com/questions/47572165/whats-the-best-way-to-add-a-trailing-slash-to-a-pathlib-directory).
- flake8-use-pathlib does not find [some changes](https://github.com/artefactual/archivematica-storage-service/blob/a8aefbc84b605fad340e383c3a9f353caadc1e8b/storage_service/common/management/commands/import_aip.py#L184). They have also changed.

### DIFF
--- a/storage_service/common/management/commands/import_aip.py
+++ b/storage_service/common/management/commands/import_aip.py
@@ -45,13 +45,13 @@ via a pipeline? Should this be fixed by the import command?::
 """
 import logging
 import os
+import pathlib
 import shlex
 import shutil
 import subprocess
 import tarfile
 import tempfile
 import uuid
-from pathlib import Path
 from pwd import getpwnam
 
 import bagit
@@ -174,14 +174,14 @@ def fail(string):
 
 
 def is_compressed(aip_path):
-    return Path(aip_path).is_file()
+    return pathlib.Path(aip_path).is_file()
 
 
 def tree(path):
     for root, _, files in os.walk(path):
         level = root.replace(path, "").count(os.sep)
         indent = " " * 4 * (level)
-        print(header(f"{indent}{Path(root).name}/"))
+        print(header(f"{indent}{pathlib.Path(root).name}/"))
         subindent = " " * 4 * (level + 1)
         for f in files:
             print(okgreen(f"{subindent}{f}"))
@@ -208,19 +208,19 @@ def _decompress_tar_gz(aip_path, temp_dir):
     with tarfile.open(aip_path) as tar:
         aip_root_dir = os.path.commonprefix(tar.getnames())
         tar.extractall(path=temp_dir)
-    return Path(temp_dir) / aip_root_dir
+    return pathlib.Path(temp_dir) / aip_root_dir
 
 
 def _decompress_7z(aip_path, temp_dir):
     cmd = shlex.split(f"7z x {aip_path} -o{temp_dir}")
     subprocess.check_output(cmd)
-    temp_dir_path = Path(temp_dir)
+    temp_dir_path = pathlib.Path(temp_dir)
     # Convert the Path object to a string to ensure compatibility with bagit.py
     return str(temp_dir_path / next(temp_dir_path.iterdir()))
 
 
 def confirm_aip_exists(aip_path):
-    if not Path(aip_path).exists():
+    if not pathlib.Path(aip_path).exists():
         raise ImportAIPException(f"There is nothing at {aip_path}")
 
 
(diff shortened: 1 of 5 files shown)


replaceafill:
Hhmm I don't see the button to approve the CI workflow run. I'll try closing and reopening the PR.

storage_service/common/startup.py
```diff
@@ -149,10 +140,10 @@ def populate_default_locations():
         if created and loc_info.get("create_dirs"):
             LOGGER.info("Creating %s Location %s", loc_info["purpose"], new_loc)
             try:
-                os.mkdir(new_loc.full_path)
+                pathlib.Path(new_loc.full_path).mkdir()
                 # Hack for extra recovery dir
                 if loc_info["purpose"] == locations_models.Location.AIP_RECOVERY:
-                    os.mkdir(os.path.join(new_loc.full_path, "backup"))
+                    pathlib.Path(new_loc.full_path).joinpath("backup").mkdir()
```
replaceafill: Could these two `pathlib.Path(new_loc.full_path)` be extracted to a variable above?

I think that would also allow you to do `(NEW_VAR / "backup").mkdir()`.

storage_service/common/management/commands/import_aip.py
```diff
@@ -43,9 +43,9 @@
     -rwxrwxr-x  1 joeldunham  wheel    51M 30 Jul 22:58 /tmp/am-pipeline-data/www/AIPsStore/237e/12b4/03c9/451a/9be7/4915/b0bd/9012/FakeCVA2-237e12b4-03c9-451a-9be7-4915b0bd9012.7z
 
 """
-import glob
 import logging
 import os
+import pathlib
```
replaceafill, line 48: I think the logic of this command (and probably others) is very well abstracted from the rest of the application which gives us more freedom.

The change you propose here of converting several strings into paths (granted, it's the approach we've been following so far) might be unnecessary if we start with paths from the beginning. Specifically from the point where the command parses its arguments:

```python
        parser.add_argument("aip_path", help="Full path to the AIP to be imported", type=pathlib.Path)
```

That would give you a path from the beginning which you only convert to strings when it's necessary.

It reminds me of [Ned Batchelder's "Unicode Sandwich"](https://nedbatchelder.com/text/unipain.html). The concept of handling native types inside the application which you decode as soon as possible and encode as late as possible was really useful for porting strings from Python 2 to 3 and more recently when we migrated strings to UUIDs.

storage_service/common/startup.py
```diff
@@ -58,15 +59,14 @@ def release(self):
 
 def populate_default_locations():
     """Create default local filesystem space and its locations."""
+
+    BASE_PATH = pathlib.Path("var") / "archivematica"
```
replaceafill, line 63: :partying_face:

replaceafill (changes requested):
This is looking great @klavman. Please see my comment about the `import_aip` command and let me know what you think.

storage_service/common/startup.py
klavman: Done, the requested change has been implemented and pushed to the branch.

storage_service/common/management/commands/import_aip.py
klavman: Thank you for the insightful suggestion. I appreciate the parallel drawn to the "Unicode Sandwich" concept and agree with it. I've implement this approach to minimize unnecessary string conversions. Please have another look when you get a chance.

In [this part](https://github.com/artefactual/archivematica-storage-service/blob/7559a8ee151736e082a4901a354e3c32cd9b30c2/storage_service/common/management/commands/import_aip.py#L388-L393) I have more doubts due to its link with `models/package`, `models/location` and `utils`. If I focus on them now or on the next PRs.

replaceafill: This looks good. We can address cleanups in future PRs :+1:

replaceafill (approved):
This looks great @klavman. Thank you so much!

PR #693, opened 2024-03-06
Update common.utils and common.tests to use pathlib instead of os.path

klavman:
Connects to https://github.com/archivematica/Issues/issues/1622

- **Refactoring:** In some cases, the use of `os.path` functions like `os.path.split` or  `tests`, necessitated slight refactoring for a seamless transition to `pathlib`. These changes were made to ensure compatibility and maintain the readability.

- **Feedback and reversions:** I understand the importance of keeping changes focused and minimally invasive. If there are any changes that seem to extend beyond the intended scope of simply switching to `pathlib`, or if any refactoring is viewed as unnecessary or overly complex, please feel free to highlight these concerns. I'm open to reverting specific changes or discussing alternative approaches to ensure that the PR aligns well with our coding practices and objectives.


.pre-commit-config.yaml
```diff
@@ -56,7 +56,10 @@ repos:
         storage_service/common/gpgutils.py |
         storage_service/common/startup.py |
         storage_service/common/management/commands/populate_aip_checksums.py |
-        storage_service/common/management/commands/import_aip.py
+        storage_service/common/management/commands/import_aip.py |
+        storage_service/common/utils.py |
+        tests/common/test_utils.py |
+        tests/common/test_command_import_aip.py
```
replaceafill: I think the `flake8-use-pathlib` hook now passes completely on the `storage_service.common` and `tests.common` packages. What do you think of simplifying the hook config like this?

```diff
-        storage_service/common/gpgutils.py |
-        storage_service/common/startup.py |
-        storage_service/common/management/commands/populate_aip_checksums.py |
-        storage_service/common/management/commands/import_aip.py |
-        storage_service/common/utils.py |
-        tests/common/test_utils.py |
-        tests/common/test_command_import_aip.py
+        storage_service/common/.*\.py |
+        tests/common/.*\.py
```

storage_service/common/utils.py
```diff
@@ -347,28 +351,31 @@ def get_compress_command(compression, extract_path, basename, full_path):
         `command` is the compression command (as a list of strings)
         `compressed_filename` is the full path to the compressed file
     """
+    extract_path = pathlib.Path(extract_path) / basename
+    full_path = pathlib.Path(full_path)
+
     if compression in (COMPRESSION_TAR, COMPRESSION_TAR_BZIP2, COMPRESSION_TAR_GZIP):
-        compressed_filename = os.path.join(extract_path, basename + TAR_EXTENSION)
-        relative_path = os.path.dirname(full_path)
+        compressed_filename = extract_path.with_suffix(TAR_EXTENSION)
```
replaceafill, line 358: I think `compressed_filename` needs to be converted into a string when returning from this function to not break the code that calls this utility function:

```python
    return (command, str(compressed_filename))
```

storage_service/common/utils.py
```diff
@@ -690,7 +698,7 @@ def find_tagmanifest(file_path):
 
     for tagmanifest in tagmanifest_files:
         if tagmanifest in bag_files:
-            return os.path.join(file_path, tagmanifest)
+            return file_path / tagmanifest
```
replaceafill, line 701: I think this also needs to be returned as a string.

storage_service/common/utils.py
```diff
@@ -699,7 +707,7 @@ def uuid_to_path(uuid):
     Every 4 alphanumeric characters of the UUID become a folder name."""
     uuid = uuid.hex
     path = [uuid[i : i + 4] for i in range(0, len(uuid), 4)]
-    path = os.path.join(*path)
+    path = pathlib.Path(*path)
     LOGGER.debug("path %s", path)
     return path
```
replaceafill: I think this also needs to be returned as a string.

replaceafill (changes requested):
Thanks @klavman, this is looking good.

I ran the [AMAUATs workflow](https://github.com/artefactual/archivematica/blob/3a3deeffcc1a849d9ca51d32e4557099279f1f92/.github/workflows/acceptance-test.yml) on a copy of your branch to detect integration problems not caught by the unit tests.

You can see in the output of [the workflow run](https://github.com/artefactual/archivematica/actions/runs/8208652702) a few tags failed. At the bottom of the workflow run page in the `Artifacts` section you can download the logs of each failed tag run. They contain a zip file with the logs of the four services.

For example, I extracted the [`docker-logs-aip-encrypt-Chrome` artifact](https://github.com/artefactual/archivematica/actions/runs/8208652702/artifacts/1310561870) and noticed in the `storage-service.log` file the following traceback:

```console
ERROR     2024-03-08 12:42:24  django.request.tastypie:resources:_handle_500:309:  Internal Server Error: /api/v2/file/1ff70e22-05df-4151-b796-6b6df8e3f1a0/
  File "/pyenv/data/versions/3.9.18/lib/python3.9/site-packages/tastypie/resources.py", line 221, in wrapper
    response = callback(request, *args, **kwargs)
  File "/pyenv/data/versions/3.9.18/lib/python3.9/site-packages/tastypie/resources.py", line 465, in dispatch_detail
    return self.dispatch('detail', request, **kwargs)
  File "/pyenv/data/versions/3.9.18/lib/python3.9/site-packages/tastypie/resources.py", line 488, in dispatch
    response = method(request, **kwargs)
  File "/pyenv/data/versions/3.9.18/lib/python3.9/site-packages/tastypie/resources.py", line 1491, in put_detail
    updated_bundle = self.obj_update(bundle=bundle, **self.remove_api_resource_names(kwargs))
  File "/src/storage_service/locations/api/resources.py", line 1167, in obj_update
    bundle = self.obj_update_hook(bundle, **kwargs)
  File "/src/storage_service/locations/api/resources.py", line 1208, in obj_update_hook
    bundle.obj.finish_reingest(
  File "/src/storage_service/locations/models/package.py", line 2623, in finish_reingest
    storage_effects = self._move_rein_updated_to_final_dest(
  File "/src/storage_service/locations/models/package.py", line 2812, in _move_rein_updated_to_final_dest
    updated_aip_src_path = updated_aip_path.replace(
TypeError: replace() takes 2 positional arguments but 4 were given
ERROR     2024-03-08 12:42:24  django.request:log:log_response:241:  Internal Server Error: /api/v2/file/1ff70e22-05df-4151-b796-6b6df8e3f1a0/
```

I think that `updated_aip_path.replace` error is produced at the end of the reingest process because the code in `package.py` that calls the `utils` module is expecting a string but is getting a `pathlib.Path` object instead. I've raised a few occurrences of this problem in my review.

I think in this iteration we need to guarantee the utility functions still return strings and not `pathlib.Path` objects. In a future iteration we can start tracking all the callers and updating them to pass and expect `pathlib.Paths`.

In case you're interested in running the AMAUATs tags to reproduce the problem in your development environment we have a [sample command](https://github.com/artefactual/archivematica/tree/3a3deeffcc1a849d9ca51d32e4557099279f1f92/hack#amauats) in the README.

Let me know if this you have any questions.

.pre-commit-config.yaml
klavman: Absolutely agree, let's proceed.

klavman:
I'll take a look and start addressing it. Thanks for the detailed feedback!

klavman:
@replaceafill  as you mentioned and after reviewing it, with just this change

`return (command, str(compressed_filename))`

Both the unit tests and the AMAUATs jobs pass successfully.

Here's the process I followed:

- To avoid running the entire job, I used the `Makefile` commands to add a specific scenario where there was an error using `-n`.
- Once scenarios were no longer failing, I left all the jobs to run sequentially, one by one. Between each job, I scheduled a `make flush`.

Although it takes time, if it’s not being done correctly or can be simplified, please let me know.

If everything looks good after your review and you think I should also make changes in `find_tagmanifest` and `uuid_to_path`, please let me know as well.

replaceafill:
> If everything looks good after your review and you think I should also make changes in `find_tagmanifest` and `uuid_to_path`, please let me know as well.

Thanks @klavman, I just ran the AMAUATs workflow again and it [looks good](https://github.com/artefactual/archivematica/actions/runs/8286627226). Please go ahead and make the other changes.

> Although it takes time, if it’s not being done correctly or can be simplified, please let me know.

Yeah, the development environment is not very ergonomic for running AMAUATs, so tweaking things like that is fine. An alternative approach we came up with is to use `behave` [tags](https://behave.readthedocs.io/en/stable/gherkin.html#tags). We edit the feature file in the `archivematica-acceptance-tests` submodule and add a custom tag (e.g. `@wip`) to the scenario or scenarios we want to run. Then we run:

```
make test-at-behave TAGS=wip BROWSER=Firefox
```

to run only that or those scenarios.

A better long term approach would be to add appropriate tags to the feature files and their scenarios so you can target them better. There are [a couple of examples](https://github.com/artefactual/archivematica/blob/d42c61206b0e82af9f6a39386b801c2d39a0e827/.github/workflows/acceptance-test.yml#L19-L20) in the AMAUATs workflow where we do this combining tags attached to the [whole feature](https://github.com/artefactual-labs/archivematica-acceptance-tests/blob/398a07817c9592a273dd570c3c84328872fd95d8/features/core/ingest-policy-check.feature#L1) file and its scenarios. Hope this helps.

storage_service/common/utils.py
klavman: Done.

storage_service/common/utils.py
klavman: Could you please take a look at this? The `find_tagmanifest` function is exclusively utilized by the `generate_checksum` function within [`common.utils`](https://github.com/klavman/archivematica-storage-service/blob/478630dd9ebca1c1d8527f6ece229ec53aa50925/storage_service/common/utils.py#L675). Although `generate_checksum` is used in `package.py` or within the `locations` module, its sole purpose is to calculate and return the checksum. 

If I'm missing something, please let me know, and I will make the corresponding change and adjust the involved tests accordingly.

replaceafill: Ah I see. Thanks for tracking the call chain. In that case I think returning a `pathlib.Path` is safe :+1: (we're **inside** the "sandwich" :drum: :blush:)

replaceafill (approved):
Thank you so much @klavman! Great work.

PR #700, opened 2024-04-02
Remove re_path URL dispatcher from urls

Dhwaniartefact:
Convert regular expressions `re_path` URL dispatcher to `path` in storage service APIs

Ref : https://docs.djangoproject.com/en/3.2/topics/http/urls/#path-converters

### DIFF
--- a/storage_service/administration/urls.py
+++ b/storage_service/administration/urls.py
@@ -1,6 +1,5 @@
 from administration import views
 from django.urls import path
-from django.urls import re_path
 
 app_name = "administration"
 urlpatterns = [
@@ -10,17 +9,15 @@ urlpatterns = [
     path("version/", views.version_view, name="version"),
     path("users/", views.user_list, name="user_list"),
     path("users/create/", views.user_create, name="user_create"),
-    re_path(r"^users/(?P<id>[-\w]+)/edit/$", views.user_edit, name="user_edit"),
-    re_path(r"^users/(?P<id>[-\w]+)/$", views.user_detail, name="user_detail"),
+    path("users/<int:id>/edit/", views.user_edit, name="user_edit"),
+    path("users/<int:id>/", views.user_detail, name="user_detail"),
     path("language/", views.change_language, name="change_language"),
     path("keys/", views.key_list, name="key_list"),
-    re_path(
-        r"^keys/(?P<key_fingerprint>[\w]+)/detail$", views.key_detail, name="key_detail"
-    ),
+    path("keys/<str:key_fingerprint>/detail/", views.key_detail, name="key_detail"),
     path("keys/create/", views.key_create, name="key_create"),
     path("keys/import/", views.key_import, name="key_import"),
-    re_path(
-        r"^keys/(?P<key_fingerprint>[\w]+)/delete/$",
+    path(
+        "keys/<str:key_fingerprint>/delete/",
         views.key_delete,
         name="key_delete",
     ),
--- a/storage_service/administration/urls.py
+++ b/storage_service/administration/urls.py
@@ -13,7 +13,7 @@ urlpatterns = [
     path("users/<int:id>/", views.user_detail, name="user_detail"),
     path("language/", views.change_language, name="change_language"),
     path("keys/", views.key_list, name="key_list"),
-    path("keys/<str:key_fingerprint>/detail/", views.key_detail, name="key_detail"),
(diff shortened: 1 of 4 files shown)


storage_service/administration/urls.py
```diff
@@ -10,17 +9,15 @@
     path("version/", views.version_view, name="version"),
     path("users/", views.user_list, name="user_list"),
     path("users/create/", views.user_create, name="user_create"),
-    re_path(r"^users/(?P<id>[-\w]+)/edit/$", views.user_edit, name="user_edit"),
-    re_path(r"^users/(?P<id>[-\w]+)/$", views.user_detail, name="user_detail"),
+    path("users/<int:id>/edit/", views.user_edit, name="user_edit"),
```
replaceafill, line 12: I think converting `id` to an `int` will break this comparison in the `user_edit` view:

https://github.com/artefactual/archivematica-storage-service/blob/05387d365467a82212fde7db23178410417014f2/storage_service/administration/views.py#L87

storage_service/administration/urls.py
```diff
@@ -10,17 +9,15 @@
...
+    path("users/<int:id>/", views.user_detail, name="user_detail"),
```
replaceafill, line 13: I think converting `id` to an `int` will break this comparison in the `user_detail` view:

https://github.com/artefactual/archivematica-storage-service/blob/05387d365467a82212fde7db23178410417014f2/storage_service/administration/views.py#L140

storage_service/administration/urls.py
```diff
@@ -10,17 +9,15 @@
...
     path("language/", views.change_language, name="change_language"),
     path("keys/", views.key_list, name="key_list"),
-    re_path(
-        r"^keys/(?P<key_fingerprint>[\w]+)/detail$", views.key_detail, name="key_detail"
-    ),
+    path("keys/<str:key_fingerprint>/detail/", views.key_detail, name="key_detail"),
```
replaceafill: Why do we need a new trailing slash in this new route?

storage_service/storage_service/urls.py
```diff
@@ -15,7 +14,7 @@
 urlpatterns = [
     path("", TemplateView.as_view(template_name="index.html")),
     # Uncomment the next line to enable the admin:
```
replaceafill: Do you think we should remove this comment? The admin is enabled by default :shrug:

replaceafill (changes requested):
This looks good @Dhwaniartefact, but I think you need to update a couple of views to handle comparisons correctly. I also left a question about a new trailing slash you have introduced.

Do you plan to remove the `re_path` calls from the `storage_service/locations/api/resources.py` module too? Is that even possible with `tastypie` in the mix?

storage_service/administration/urls.py
Dhwaniartefact: oops!

Dhwaniartefact:
> Do you plan to remove the `re_path` calls from the `storage_service/locations/api/resources.py` module too? Is that even possible with `tastypie` in the mix?

Tastypie resources are using re_path, which I am attempting to remove.

storage_service/administration/views.py
```diff
@@ -84,7 +84,7 @@ def user_list(request):
 
 def user_edit(request, id):
     edit_allowed = settings.ALLOW_USER_EDITS and (
-        request.user.has_perm("auth.change_user") or str(request.user.id) == id
+        request.user.has_perm("auth.change_user") or (request.user.id) == id
```
replaceafill:
```suggestion
        request.user.has_perm("auth.change_user") or request.user.id == id
```

storage_service/administration/views.py
```diff
@@ -137,7 +137,7 @@ def user_create(request):
 
 def user_detail(request, id):
     # Only a superuser or the user themselves can view their full details
-    view_allowed = request.user.is_superuser or str(request.user.id) == id
+    view_allowed = request.user.is_superuser or (request.user.id) == id
```
replaceafill:
```suggestion
    view_allowed = request.user.is_superuser or request.user.id == id
```

replaceafill (approved):

Dhwaniartefact:
> > Do you plan to remove the `re_path` calls from the `storage_service/locations/api/resources.py` module too? Is that even possible with `tastypie` in the mix?
> 
> Tastypie resources are using re_path, which I am attempting to remove.

I am going to address removing of re_path from `storage_service/locations/api/resources.py` in a future PR.

PR #713, opened 2024-05-08
Fix JSON serialization for fixity failure report

Dhwaniartefact:


replaceafill (changes requested):
Could you first add a test that demonstrates the failure?

storage_service/locations/models/package.py
```diff
@@ -2062,25 +2062,25 @@ def get_fixity_check_report_send_signals(
...
             "success": success,
-            "message": message,
+            "message": str(message),
             "failures": {"files": {"missing": [], "changed": [], "untracked": []}},
             "timestamp": timestamp,
         }
         for failure in failures:
             if isinstance(failure, bagit.FileMissing):
-                info = {"path": failure.path, "message": str(failure)}
+                info = {"path": failure.path, "message": failure}
```
replaceafill: Why do we have to remove the `str` here and below?

tests/locations/test_package.py
```diff
@@ -1370,6 +1370,89 @@ def test_move_uncompressed_aip(self):
...
+@pytest.mark.django_db
+def space(tmp_path):
+    space_dir = tmp_path / "space"
+    space_dir.mkdir()
+
+    staging_dir = tmp_path / "staging"
+    staging_dir.mkdir()
+
+    space = models.Space.objects.create(
+        uuid=uuid.uuid4(),
```
replaceafill: Is there a point on initializing the `uuid` attributes in these fixtures? The model fields do it automatically.

tests/locations/test_package.py
```diff
@@ -1370,6 +1370,89 @@ def test_move_uncompressed_aip(self):
...
+def location(space):
+    aipstore = models.Location.objects.create(
+        uuid=uuid.uuid4(),
+        space=space,
+        relative_path="fs-aips",
+        purpose="AS",
+    )
+    models.Location.objects.create(
+        space=space, purpose=models.Location.STORAGE_SERVICE_INTERNAL, relative_path=""
+    )
```
replaceafill: Why don't we move this model instance to its own fixture? It's not related to the AIPStorage location.

tests/locations/test_package.py
```diff
@@ -1370,6 +1370,89 @@ def test_move_uncompressed_aip(self):
...
+    return aipstore
+
+
+@pytest.fixture
+@pytest.mark.django
```
replaceafill: Why does this fixture need database access?

tests/locations/test_package.py
```diff
@@ -1370,6 +1370,89 @@ def test_move_uncompressed_aip(self):
...
+    return shutil.copy(src, dst)
+
+
+@pytest.fixture
+@pytest.mark.django_db
+def package(location, bag_fixture):
+    return models.Package.objects.create(
+        uuid=uuid.uuid4(),
+        current_location=location,
+        current_path=bag_fixture,
```
replaceafill: AFAIK the `current_path` of a package is a relative path inside its location. I think you're setting an absolute path here. The path of the package should depend on its location and space. See the [`full_path` property](https://github.com/artefactual/archivematica-storage-service/blob/f4ff4f3ac4bfeb4ba1a2ed895b76189a7f57c408/storage_service/locations/models/package.py#L209-L216)

tests/locations/test_package.py
```diff
@@ -1370,6 +1370,89 @@ def test_move_uncompressed_aip(self):
...
+        status="Uploaded",
+    )
+
+
+@pytest.mark.django_db
+@mock.patch(
+    "common.utils.generate_checksum",
+    return_value=mock.Mock(
+        **{
+            "hexdigest.return_value": "myhash",
```
replaceafill: Could we use real hashes here and in the `package.checksum` below, maybe MD5 hashes?

tests/locations/test_package.py
```diff
@@ -1370,6 +1370,89 @@ def test_move_uncompressed_aip(self):
...
+        }
+    ),
+)
+def test_get_fixity_check_report_send_signals(generate_checksum, package):
+    """
+    It verifies report of fixity check
+    """
```
replaceafill: Is this really what this test is doing/confirming? It looks like it's testing a failure scenario and its name and docstring don't seem to match that?

tests/locations/test_package.py
```diff
@@ -1370,6 +1370,89 @@ def test_move_uncompressed_aip(self):
...
+
+    report, response = package.get_fixity_check_report_send_signals()
+
+    assert response == {
+        "success": False,
+        "message": "Incorrect package checksum",
+        "failures": {"files": {"missing": [], "changed": [], "untracked": []}},
+        "timestamp": None,
+    }
+    assert response["message"] == "Incorrect package checksum"
```
replaceafill: Why is necessary to assert this again? The full response was checked above.

replaceafill (changes requested):

tests/locations/test_package.py
Dhwaniartefact: There is no necessity of doing it. I just did it for no reason. Let me remove it.

tests/locations/test_package.py
Dhwaniartefact: oh yes, thanks for pointing it.

storage_service/locations/models/package.py
Dhwaniartefact: oh yes that's bagit exceptions.

tests/locations/test_package.py
```diff
@@ -1370,6 +1370,91 @@ def test_move_uncompressed_aip(self):
...
+    dst = os.path.join(tmp_dir, "")
+    return shutil.copy(src, dst)
+
+
+@pytest.fixture
+@pytest.mark.django_db
+def package(location, bag_fixture):
+    return models.Package.objects.create(
+        current_location=location,
+        current_path=os.path.join(location.full_path, bag_fixture),
```
replaceafill: If I insert a `breakpoint()` before this return, the value of the `current_path` field looks like:

```
(Pdb) p os.path.join(location.full_path, bag_fixture)
'/tmp/pytest-of-user/pytest-0/test_get_fixity_check_report_s0/dir/working_bag.zip'
```

That's still an absolute path not relative. The value of `current_path` needs to be a relative path.

tests/locations/test_package.py
```diff
@@ -1370,6 +1370,91 @@ def test_move_uncompressed_aip(self):
...
+            "hexdigest.return_value": "098f6bcd4621d373cade4e832627b4f9",
+        }
+    ),
+)
+def test_get_fixity_check_report_send_signals(
+    generate_checksum, package, internal_location
+):
+    """
+    Verifies scenarios of fixity failure signals
+    """
```
replaceafill: The docstring is too vague: which scenarios is referring to? how many are they? And the name of the test seems to suggest it's testing the "happy path" which I don't think it is.

I'd recommend you to apply the test naming conventions suggested during https://github.com/artefactual/archivematica/pull/1932. I think that if you name the test clearly using that framework you don't even need the docstring.

replaceafill (changes requested):

tests/locations/test_package.py
replaceafill: I just tested this alternative approach:

* Create the location directory in the `location` fixture
* Copy the fixture file in the `package` fixture

See how `current_path` is set to a relative path when creating the `Package` model instance.

```diff
diff --git a/tests/locations/test_package.py b/tests/locations/test_package.py
index 7e235b5..0468e14 100644
--- a/tests/locations/test_package.py
+++ b/tests/locations/test_package.py
@@ -1396,6 +1396,9 @@ def location(space):
         relative_path="fs-aips",
         purpose="AS",
     )
+
+    pathlib.Path(aipstore.full_path).mkdir()
+
     return aipstore
 
 
@@ -1407,25 +1410,21 @@ def internal_location(space):
     )
 
 
-@pytest.fixture
-def bag_fixture(tmp_path):
-    tmp_dir = tmp_path / "dir"
-    tmp_dir.mkdir()
-    src = os.path.join(FIXTURES_DIR, "working_bag.zip")
-    dst = os.path.join(tmp_dir, "")
-    return shutil.copy(src, dst)
-
-
 @pytest.fixture
 @pytest.mark.django_db
-def package(location, bag_fixture):
-    return models.Package.objects.create(
+def package(location):
+    result = models.Package.objects.create(
         current_location=location,
-        current_path=os.path.join(location.full_path, bag_fixture),
+        current_path="working_bag.zip",
         package_type="AIP",
         status="Uploaded",
     )
 
+    src = os.path.join(FIXTURES_DIR, "working_bag.zip")
+    shutil.copy(src, result.full_path)
+
+    return result
+
 
 @pytest.mark.django_db
 @mock.patch(
```

Dhwaniartefact: Ahhh...well understood. Thanks!

tests/locations/test_package.py
replaceafill: @Dhwaniartefact This is not entirely addressed yet. You removed the docstring but didn't rename the test.

replaceafill (approved):

PR #715, opened 2024-05-17
Fixity failure email message changes

Dhwaniartefact:
Email messages about fixity failures are shortened by removing information like `location` and `report`.

### DIFF
--- a/storage_service/locations/signals.py
+++ b/storage_service/locations/signals.py
@@ -8,6 +8,7 @@ from django.db.models import signals
 from django.dispatch import receiver
 from django.dispatch import Signal
 from django.urls import reverse
+from django.utils import timezone
 from django.utils.translation import gettext as _
 from prometheus_client import Counter
 from tastypie.models import create_api_key
@@ -63,7 +64,7 @@ def _log_report(uuid, success, message=None):
     from . import models
 
     package = models.Package.objects.get(uuid=uuid)
-    models.FixityLog.objects.create(
+    return models.FixityLog.objects.create(
         package=package, success=success, error_details=message
     )
 
@@ -71,22 +72,21 @@ def _log_report(uuid, success, message=None):
 @receiver(failed_fixity_check, dispatch_uid="fixity_check")
 def report_failed_fixity_check(sender, **kwargs):
     report_data = json.loads(kwargs["report"])
-    _log_report(kwargs["uuid"], False, report_data["message"])
+    fixity_log = _log_report(kwargs["uuid"], False, report_data["message"])
+    timestamp = timezone.localtime(fixity_log.datetime_reported).strftime(
+        "%Y-%m-%d %H:%M:%S"
+    )
 
     subject = _("Fixity check failed for package %(uuid)s") % {"uuid": kwargs["uuid"]}
     message = (
         _(
             """
-A fixity check failed for the package with UUID %(uuid)s. This package is currently stored at: %(location)s
-
-Full failure report (in JSON format):
-%(report)s
+[%(timestamp)s] A fixity check failed for the package with UUID %(uuid)s.
 """
         )
         % {
+            "timestamp": timestamp,
             "uuid": kwargs["uuid"],
-            "location": kwargs["location"],
-            "report": kwargs["report"],
         }
     )
 
--- a/tests/locations/test_signals.py
+++ b/tests/locations/test_signals.py
@@ -1,0 +1,50 @@
+import datetime
+import json
+from unittest import mock
+
+import pytest
+from django.utils import timezone
+from locations import models
+from locations import signals
+
+
+@pytest.mark.django_db
+@mock.patch("locations.signals._notify_administrators")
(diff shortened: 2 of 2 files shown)


storage_service/locations/signals.py
```diff
@@ -63,30 +63,27 @@ def _log_report(uuid, success, message=None):
...
         package=package, success=success, error_details=message
     )
 
 
 @receiver(failed_fixity_check, dispatch_uid="fixity_check")
 def report_failed_fixity_check(sender, **kwargs):
     report_data = json.loads(kwargs["report"])
-    _log_report(kwargs["uuid"], False, report_data["message"])
+    fixity_log = _log_report(kwargs["uuid"], False, report_data["message"])
+    timestamp = fixity_log.datetime_reported.strftime("%Y-%m-%d %H:%M:%S")
```
replaceafill: Correct me if I'm wrong, but I think this needs to be a [localtime](https://docs.djangoproject.com/en/4.2/ref/utils/#django.utils.timezone.localtime)?

It's 16:47 here right now and I just set up your branch in the development environment which by default uses `America/Los_Angeles` as the `TIME_ZONE` and when I call the check fixity endpoint I get a timestamp that it's in the future:

```console
archivematica-storage-service-1  | [2024-05-17 22:47:32] A fixity check failed for the package with UUID 8c9c347e-7fd4-442e-8a6f-b02338ba2b4e.
```

tests/locations/test_signals.py
```diff
@@ -0,0 +1,52 @@
...
+    with mock.patch.object(
+        timezone,
+        "now",
+        return_value=expected_time,
+    ):
+        signals.report_failed_fixity_check(sender, **kwargs)
+
+    _notify_administrators.assert_called_once_with(
+        f"Fixity check failed for package {package.uuid}",
+        f"\n[2024-05-17 11:00:00] A fixity check failed for the package with UUID {package.uuid}.\n",
```
replaceafill: Could this f-string use the `expected_time` variable like it uses `package`?

tests/locations/test_signals.py
```diff
@@ -0,0 +1,52 @@
...
+    )
+    assert models.FixityLog.objects.filter(
+        package=package.uuid, success=False, datetime_reported=expected_time
+    )
```
replaceafill: I think `filter(...)` evaluates to boolean just fine but I find explicit checks to be clearer. Imagine the `filter(...)` call returns multiple `FixityLog` instances, is that expected? etc.

```suggestion
    assert models.FixityLog.objects.filter(
        package=package.uuid, success=False, datetime_reported=expected_time
    ).count() == 1
```

tests/locations/test_signals.py
```diff
@@ -0,0 +1,52 @@
...
+                "message": "Incorrect package checksum",
+                "failures": {"files": {"missing": [], "changed": [], "untracked": []}},
+                "timestamp": None,
+            }
+        ),
+    }
+    settings.TIME_ZONE = "America/Los_Angeles"
+    expected_time = timezone.make_aware(
+        datetime.datetime(2024, 5, 17, 11, 00),
+        timezone=timezone.activate(settings.TIME_ZONE),
```
replaceafill: I have not tested but I think Django uses the `TIME_ZONE` setting by default in `make_aware`, so you could remove this `timezone` keyword argument. That could also make the `settings.TIME_ZONE` assignment above unnecessary.

tests/locations/test_signals.py
```diff
@@ -0,0 +1,52 @@
...
+def test_report_failed_fixity_check(_notify_administrators: mock.Mock):
+    package = models.Package.objects.create(
+        current_location=models.Location.objects.create(
+            space=models.Space.objects.create()
+        )
+    )
+    sender = None
+    kwargs = {
+        "uuid": str(package.uuid),
+        "location": "/tmp",
```
replaceafill: Is the `location` still necessary in this implementation?

replaceafill (approved):

replaceafill:
@Dhwaniartefact Please remember to follow https://github.com/artefactual/archivematica/blob/4ecae4564e0cd6306e02695d45f90271dd78daf8/CONTRIBUTING.md#commit-messages when you merge this.

tests/locations/test_signals.py
Dhwaniartefact: Nope as long as we are not asserting it.

tests/locations/test_signals.py
Dhwaniartefact: sure

tests/locations/test_signals.py
replaceafill: Is the `settings.TIME_ZONE = "America/Los_Angeles"` line still necessary?

replaceafill (approved):

tests/locations/test_signals.py
Dhwaniartefact: Not really. Thanks, let me remove it.

amayita (approved):
LGTM!

PR #717, opened 2024-05-24
Filter active users to send email notifications

Dhwaniartefact:
Connected to https://github.com/archivematica/Issues/issues/1523

### DIFF
--- a/storage_service/locations/signals.py
+++ b/storage_service/locations/signals.py
@@ -21,9 +21,8 @@ successful_fixity_check = Signal()
 fixity_check_not_run = Signal()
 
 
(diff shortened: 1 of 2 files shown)


replaceafill (commented):
@Dhwaniartefact @amayita the linked issue mentions fixity report emails specifically but this change is going to affect deletion requests emails too.

Is that intended? Should that be filed and implemented as a different issue and PR?

replaceafill (approved):
This looks good @Dhwaniartefact. Please don't forget to adjust the commit message to reflect that you modified the behavior only for email notifications of the fixity check failures.

PR #724, opened 2024-06-24
Fix create_tar utility function for dotted paths

replaceafill:
This fixes the `create_tar` utility function used in the write-only replicator location to support paths with dots.

Connected to https://github.com/archivematica/Issues/issues/1697


Dhwaniartefact (approved):
Verified and tested.

PR #731, opened 2024-08-02
Fix mypy hook in pre-commit

replaceafill:
Similar to https://github.com/artefactual/archivematica/pull/1964 this removes the `types-all` package from the `mypy` hook after `types-pkg-resources` has been deprecated and yanked from PyPI.


Dhwaniartefact (approved):
Awesome :)

PR #739, opened 2024-09-12
Remove setup.py

replaceafill:
This moves the package metadata in `setup.py` to `pyproject.toml` as recommended in [Transitioning from setup.py to declarative config](https://setuptools.pypa.io/en/latest/userguide/quickstart.html#transitioning-from-setup-py-to-declarative-config).

It also introduces `mypy` configuration to allow us to [gradually add type annotations](https://mypy.readthedocs.io/en/stable/existing_code.html#start-small).


Dhwaniartefact (approved):
LGTM!

PR #742, opened 2024-09-20
Enable OIDC tests

replaceafill:
Similar to https://github.com/artefactual/archivematica/pull/1993. It also extend test coverage fort `CustomOIDCBackend.update_user` method.


sbreker (approved):
Thanks @replaceafill! :+1:

PR #743, opened 2024-09-26
Add secondary OIDC provider capability

sbreker:
Add ability to define more than one OIDC provider in AM. Specific providers can be chosen using HTTP query params passed to the server when authenticating.


sbreker (commented):
Looks great @replaceafill! :+1: 

Curious why you chose to hard code `LOGIN_URL = "/oidc/authenticate/"`?

replaceafill:
> Curious why you chose to hard code `LOGIN_URL = "/oidc/authenticate/"`?

@sbreker sorry I forgot to include the explanation in the commit. During manual testing I noticed the `LoginRequiredMiddleware` class prevents Django from reaching the point where it understand the `LOGIN_URL` as a named URL pattern as we intended.

See https://github.com/artefactual/archivematica-storage-service/blob/a38480989789dbd5213b847af3c3221d1fcb33dc/storage_service/common/middleware.py#L53 where the `oidc_authentication_init` pattern is converted into a string and the middleware redirects you to an nonexistent `/oidc_authentication_init` URL instead.

Captura desde 2024-10-03 16-50-56

The Dashboard doesn't have this problem, but I decided to keep the values consistent for now.

replaceafill (approved):
Thank you so much for your hard work @sbreker! :tada:

PR #748, opened 2024-10-10
Fix provider name handling in OIDC integration

replaceafill:
This fixes a problem in the OIDC middleware where secondary providers
with lowercase names were not found in the `OIDC_PROVIDERS`
setting.


Dhwaniartefact (approved):
LGTM !

PR #749, opened 2024-10-16
Fix redirect URL after GPG key deletion

replaceafill:
The redirect query parameter currently points back to the deleted key, resulting in a 404 error.

This updates it to redirect to the keys table instead.


Dhwaniartefact (approved):
LGTM.

PR #751, opened 2024-10-21
Update GPG key deletion test

replaceafill:
This updates the test to work when it is executed from the development environment where most of the Storage Service environment variables are not set and OS tools are not installed.


Dhwaniartefact (approved):
lgtm !

PR #753, opened 2024-10-26
Remove pytest-mock dependency

replaceafill:
`pytest-mock` provided value while we migrated from Python 2 to Python 3, but it's become less relevant now that the `unittest.mock` is included in the standard library. We also don't use any of its extra features like `Spy`s or `Stub`s.

This replaces it with `unittest.mock` and removes the dependency from the requirements files.

### DIFF
--- a/requirements-dev.in
+++ b/requirements-dev.in
@@ -4,7 +4,6 @@
 pip-tools
 pytest-cov
 pytest-django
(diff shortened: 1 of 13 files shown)


Dhwaniartefact (approved):
nice!

PR #760, opened 2025-02-18
Remove temporary bash scripts in compression details

replaceafill:
This is preparatory work for upgrading to Ubuntu 24.04 which contains a newer version of the `7z` command.

A notable difference in the new version of `7z` is that the output doesn't include an explicit `Version` string which is what the Storage Service currently looks for, so this PR updates the `get_7z_version` helper to work across versions.

The current order of commits are:

* 998bd50425f61a516cef1cfd0d9905bc35c84b97 extends test coverage of the `import_aip` management command to ensure the PREMIS `compression` event has been created with the correct details
* 8fd5bf14cade87b23364b09519f66fccbc756b8a updates the test of the `get_format_info` helper to be compatible with newer `7z` versions
* a6f4e33c9ce51a8e516b2704bbd03886f75f82a9 updates the `get_7z_version` helper and its tests to be compatible with newer `7z` versions (check the commit message for output samples)
* fa79dc71f7fa3cb6e56ad68bb8d8e0e5e8e77aa6 replaces the temporary bash scripts created for getting the version of the compression tools with the `get_7z_version` and `get_tar_version` helpers.
* b214b335b347fd810a5ca56cfc09fd181c1e6f75 adds test cases to ensure the `get_7z_version` helper works correctly with the output of the latest `7z` version.

Connected to https://github.com/archivematica/Issues/issues/1729

### DIFF
--- a/storage_service/common/utils.py
+++ b/storage_service/common/utils.py
@@ -467,11 +467,13 @@ def get_tool_info_command(compression):
 
 
 def get_7z_version():
-    return [
-        line
-        for line in subprocess.check_output("7z").splitlines()
-        if b"Version" in line
-    ][0].decode("utf8")
+    lines = subprocess.check_output("7z").decode().splitlines()
+    if lines[2].startswith("p7zip Version"):
+        # 7-Zip 16.02: return only version line.
+        return lines[2]
+    else:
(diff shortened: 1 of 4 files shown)


Dhwaniartefact (approved):
lgtm!

PR #761, opened 2025-02-24
Add management command to process deletion requests

hakamine:
Add a management command to allow approval of deletion requests programmatically.
Related to https://github.com/archivematica/Issues/issues/1696

### DIFF
--- a/src/archivematica/storage_service/common/management/commands/process_deletion_requests.py
+++ b/src/archivematica/storage_service/common/management/commands/process_deletion_requests.py
@@ -1,0 +1,132 @@
+"""Process pending package deletion requests.
+
+Use the ``--approve`` argument to approve the pending deletion request corresponding
+to the specified package UUID (the package must already have a submitted request).
+
+Use the ``--approve-all`` argument to approve all pending deletion requests.
+
+When run without ``--approve`` or ``--approve-all`` arguments, the command lists all
+pending deletion requests without taking any action.
+"""
+
+from argparse import ArgumentParser
+from typing import Any
+from typing import Optional
+from typing import cast
+
+from django.contrib.auth import get_user_model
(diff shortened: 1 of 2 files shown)


replaceafill:
@hakamine I just merged https://github.com/artefactual/archivematica-storage-service/pull/813 which provides a new API endpoint for reviewing AIP deletion requests in the Storage Service. It's documented in https://github.com/artefactual/archivematica-docs/commit/2dbe320a84e2b7fbd3b6afa5679e1c6d5ecdfb50

As part of this work, I refactored all the package request (deletion and recovery) workflow code into a single [module](https://github.com/artefactual/archivematica-storage-service/blob/5ebe3a2bf75f89df29d9cfa921ee6650ac5cfe93/src/archivematica/storage_service/locations/package_request.py) so that the views, API endpoints, and management commands can share the same logic. I’ve also created this [gist](https://gist.github.com/replaceafill/5fc52789cba5e9881ef7bb9f2ef8a039) to show how your management command could be implemented using the new module. Let me know if you have any feedback. Thanks!

hakamine:
Thank you @replaceafill ! I'll take a look.

hakamine:
@replaceafill  I think this may be ready for CR (basically fixed as per your code verbatim; just added an additional test case, following your tests). I ran the command on a test Archivematica instance and it worked as expected. (BTW, I remember we had also discussed adding reject functionality to the command, but I guess we can do it in a future PR).

replaceafill (approved):
This looks good @hakamine, thanks!

And I agree, we could extend it to support rejection in a future PR. I think it'd involve something along the lines of changing these two lines:

```python
            package_request.PackageRequestDecision.APPROVE,
            reason=APPROVAL_REASON,
```

and adding new flags and messages or renaming and rephrasing the existing ones to be more generic.

PR #764, opened 2025-03-14
Add support for PKCE

scollazo:
refs: https://github.com/mozilla/mozilla-django-oidc/pull/447

### DIFF
--- a/src/archivematica/storage_service/storage_service/settings/base.py
+++ b/src/archivematica/storage_service/storage_service/settings/base.py
@@ -652,6 +652,11 @@ if OIDC_AUTHENTICATION:
 
     OIDC_RP_SIGN_ALGO = environ.get("OIDC_RP_SIGN_ALGO", "HS256")
 
+    OIDC_USE_PKCE = is_true(environ.get("OIDC_USE_PKCE", "false"))
+    OIDC_PKCE_CODE_CHALLENGE_METHOD = environ.get(
+        "OIDC_PKCE_CODE_CHALLENGE_METHOD", "S256"
+    )
+
     # Username is email address
     def _get_email(email):
         return email
--- a/install/README.md
+++ b/install/README.md
@@ -584,6 +584,18 @@ If `SS_OIDC_AUTHENTICATION` is false, none of the other ones are used.
   - **Type:** `string`
   - **Default:** `HS256`
 
+- **`OIDC_USE_PKCE`**:
+  - **Description:** Controls whether the authentication backend uses PKCE
+(Proof Key For Code Exchange) during the authorization code flow.
+  - **Type:** `boolean`
+  - **Default:** `false`
+
+- **`OIDC_CODE_CHALLENGE_METHOD`**:
+  - **Description:** Sets the method used to generate the PKCE code challenge.
+This only has an effect if ``OIDC_USE_PKCE`` is ``True``.
+  - **Type:** `string`
+  - **Default:** `S256`
+
 ### AWS-specific environment variables
 
 These variables can be set to allow AWS authentication for S3 storage spaces as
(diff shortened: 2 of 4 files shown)


sbreker (approved):
Looks great @scollazo. I have added a commit which configures the Storage Service integration tests to use PKCE.

sbreker:
It might be good to add notes regarding these vars in the [README](https://github.com/artefactual/archivematica-storage-service/blob/qa/0.x/install/README.md#application-specific-environment-variables) file

src/archivematica/storage_service/storage_service/settings/base.py
```diff
@@ -652,6 +652,11 @@ def _get_settings_from_file(path):
 
     OIDC_RP_SIGN_ALGO = environ.get("OIDC_RP_SIGN_ALGO", "HS256")
 
+    OIDC_USE_PKCE = environ.get("OIDC_USE_PKCE", "false")
```
replaceafill:
```suggestion
    OIDC_USE_PKCE = is_true(environ.get("OIDC_USE_PKCE", "false"))
```

replaceafill (commented):
Looks good, but I wonder if the `OIDC_USE_PKCE` value should be a boolean so `mozilla-django-oidc` doesn't consider it truth-y:

* https://github.com/mozilla/mozilla-django-oidc/blob/2c2334fdc9b2fc72a492b5f0e990b4c30de68363/mozilla_django_oidc/middleware.py#L156
* https://github.com/mozilla/mozilla-django-oidc/blob/2c2334fdc9b2fc72a492b5f0e990b4c30de68363/mozilla_django_oidc/views.py#L199

src/archivematica/storage_service/storage_service/settings/base.py
sbreker: @replaceafill: I struggled with this with my own bool setting on the other PR - seems you can't specify an unquoted bool value in a docker compose file? I could not get it to work. The technique above is what is used in other areas of AM. See the following code to Activate OIDC in AM and SS:

https://github.com/artefactual/archivematica-storage-service/blob/qa/0.x/tests/integration/docker-compose.yml#L35
https://github.com/artefactual/archivematica-storage-service/blob/qa/0.x/src/archivematica/storage_service/storage_service/settings/base.py#L575

I don't see any reason to change this.

scollazo:
Added the PKCE related variables to the README

src/archivematica/storage_service/storage_service/settings/base.py
replaceafill: > I struggled with this with my own bool setting on the other PR - seems you can't specify an unquoted bool value in a docker compose file? I could not get it to work.

The [reference for the `environment` attribute](https://docs.docker.com/reference/compose-file/services/#environment) recommends to quote the values:

> Any boolean values; true, false, yes, no, should be enclosed in quotes to ensure they are not converted to True or False by the YAML parser.

But I think when the SS settings module reads the environment it's all strings and it's your responsibility to validate and convert into the appropriate type the application expects downstream. 

> The technique above is what is used in other areas of AM. See the following code to Activate OIDC in AM and SS:
>
> https://github.com/artefactual/archivematica-storage-service/blob/qa/0.x/tests/integration/docker-compose.yml#L35
> https://github.com/artefactual/archivematica-storage-service/blob/qa/0.x/src/archivematica/storage_service/storage_service/settings/base.py#L575
>
> I don't see any reason to change this.

I'm might be missing your point but I think these examples you refer to are exactly how my suggested change approaches it?

replaceafill: > But I think when the SS settings module reads the environment it's all strings

From https://docs.python.org/3.9/library/os.html#os.environ:

> os.environ
> A [mapping](https://docs.python.org/3.9/glossary.html#term-mapping) object where keys and values are strings that represent the process environment.

sbreker: I realise now I misread your comment @replaceafill! :man_facepalming:  Yes this instance should use is_true(). I will add a commit...

replaceafill (approved):
Thank you!

PR #767, opened 2025-03-26
Fix populate_aip_checksums command for uncompressed AIPs

Dhwaniartefact:
Connected to https://github.com/archivematica/Issues/issues/1734

### DIFF
--- a/tests/common/test_command_populate_aip_checksums.py
+++ b/tests/common/test_command_populate_aip_checksums.py
@@ -8,10 +8,13 @@ from django.core.management.base import CommandError
 
 from archivematica.storage_service.locations import models
 
+TEST_DIR = pathlib.Path(__file__).resolve().parent
+FIXTURES_DIR = TEST_DIR / "fixtures"
+POINTER_FILE_PATH = FIXTURES_DIR / "premis_3_pointer.xml"
+
 
 @pytest.fixture
-@pytest.mark.django_db
-def fs_space(tmp_path: pathlib.Path) -> models.Location:
+def fs_space(tmp_path: pathlib.Path) -> models.Space:
     space_dir = tmp_path / "space"
     space_dir.mkdir()
 
@@ -29,7 +32,6 @@ def fs_space(tmp_path: pathlib.Path) -> models.Location:
 
 
 @pytest.fixture
-@pytest.mark.django_db
 def aip_storage_fs_location(fs_space: models.Space) -> models.Location:
     result = models.Location.objects.create(
         space=fs_space,
@@ -42,7 +44,6 @@ def aip_storage_fs_location(fs_space: models.Space) -> models.Location:
 
 
 @pytest.fixture
-@pytest.mark.django_db
 def compressed_package(aip_storage_fs_location: models.Location) -> models.Package:
     package_uuid = uuid.uuid4()
     package_current_path = f"compressedaip-{package_uuid}.7z"
@@ -61,7 +62,6 @@ def compressed_package(aip_storage_fs_location: models.Location) -> models.Packa
 
 
 @pytest.fixture
-@pytest.mark.django_db
 def uncompressed_package(aip_storage_fs_location: models.Location) -> models.Package:
     package_uuid = uuid.uuid4()
     package_current_path = f"uncompressedaip-{package_uuid}"
@@ -85,8 +85,7 @@ def uncompressed_package(aip_storage_fs_location: models.Location) -> models.Pac
 
 
 @pytest.fixture
-@pytest.mark.django_db
-def aip_deleted_fs_location(aip_storage_fs_location: models.Location) -> models.Package:
+def deleted_package(aip_storage_fs_location: models.Location) -> models.Package:
     return models.Package.objects.create(
         package_type=models.Package.AIP,
         status=models.Package.DELETED,
@@ -97,7 +96,7 @@ def aip_deleted_fs_location(aip_storage_fs_location: models.Location) -> models.
 
 @pytest.mark.django_db
 def test_command_fails_when_there_are_no_uploaded_aips(
-    aip_deleted_fs_location: models.Package,
+    deleted_package: models.Package,
 ) -> None:
     with pytest.raises(CommandError, match="No AIPs with status UPLOADED found"):
         call_command("populate_aip_checksums")
@@ -117,6 +116,7 @@ def test_command_fails_when_checksum_is_missing_for_compressed_aip(
     compressed_package: models.Package,
     aip_storage_fs_location: models.Location,
 ) -> None:
+    # The compressed AIP checksum is set to None when it cannot be retrieved from the pointer file.
     get_compressed_package_checksum.return_value = (None, "sha256")
 
     call_command(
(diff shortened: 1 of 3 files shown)


tests/common/test_command_populate_aip_checksums.py
```diff
@@ -0,0 +1,287 @@
...
+import pytest
+from django.core.management import call_command
+from django.core.management.base import CommandError
+
+from archivematica.storage_service.locations import models
+
+
+@pytest.fixture
+@pytest.mark.django_db
+def fs_space(tmp_path: pathlib.Path) -> models.Location:
```
replaceafill:
```suggestion
def fs_space(tmp_path: pathlib.Path) -> models.Space:
```

tests/common/test_command_populate_aip_checksums.py
```diff
@@ -0,0 +1,287 @@
...
+    )
+
+    assert not result.is_compressed
+
+    return result
+
+
+@pytest.fixture
+@pytest.mark.django_db
+def aip_deleted_fs_location(aip_storage_fs_location: models.Location) -> models.Package:
```
replaceafill:
```suggestion
def deleted_package(aip_storage_fs_location: models.Location) -> models.Package:
```

tests/common/test_command_populate_aip_checksums.py
```diff
@@ -0,0 +1,287 @@
...
+    "archivematica.storage_service.common.utils.get_compressed_package_checksum"
+)
+def test_command_fails_when_checksum_is_missing_for_compressed_aip(
+    get_compressed_package_checksum: mock.Mock,
+    error: mock.Mock,
+    capsys: pytest.CaptureFixture[str],
+    compressed_package: models.Package,
+    aip_storage_fs_location: models.Location,
+) -> None:
+    get_compressed_package_checksum.return_value = (None, "sha256")
```
replaceafill, line 120: A comment explaining why this is needed would be helpful.

tests/common/test_command_populate_aip_checksums.py
```diff
@@ -0,0 +1,287 @@
...
+        captured.out.strip("\n")
+        == "Complete. Checksums added for 0 of 1 identified AIPs. See output for details."
+    )
+
+
+@pytest.mark.django_db
+@mock.patch(
+    "archivematica.storage_service.common.utils.get_compressed_package_checksum"
+)
+def test_command_for_compressed_aip(
```
replaceafill: The names of these next three tests don't describe what they do. Rename them to maintain consistency with your other tests.

tests/common/test_command_populate_aip_checksums.py
```diff
@@ -0,0 +1,287 @@
...
+    aip_storage_fs_location: models.Location,
+) -> None:
+    compressed_package_checksum = (
+        "c2924159fcbbeadf8d7f3962b43ec1bf301e1b4f12dd28a8b89ec819f3714747"
+    )
+    algo = "sha256"
+    get_compressed_package_checksum.return_value = (
+        compressed_package_checksum,
+        algo,
+    )
```
replaceafill: The more I see these `common.utils.get_compressed_package_checksum` mocks the more I think having an actual pointer files would be more appropriate.

`test_package` has a helper to copy existing pointer file fixtures https://github.com/artefactual/archivematica-storage-service/blob/db5b964f29a9a08da9869ce2bd5e3549c1610e20/tests/locations/test_package.py#L76

Or if you don't want to use a fixture file you could create one inline like https://github.com/artefactual/archivematica/blob/2b3c4ba787a341354fa0e0320a39674e199ab5a5/tests/dashboard/fpr/test_command_import_pronom_ids.py#L41-L55

tests/common/test_command_populate_aip_checksums.py
```diff
@@ -0,0 +1,287 @@
...
+
+@pytest.mark.django_db
+@mock.patch("archivematica.storage_service.common.utils.generate_checksum")
+def test_command_for_uncompressed_aip(
+    generate_checksum: mock.Mock,
+    capsys: pytest.CaptureFixture[str],
+    uncompressed_package: models.Package,
+    aip_storage_fs_location: models.Location,
+) -> None:
+    uncompressed_package_checksum = "098f6bcd4621d373cade4e832627b4f9"
```
replaceafill: This is an `md5` checksum but the assertion below mentions `sha256`.

tests/common/test_command_populate_aip_checksums.py
```diff
@@ -0,0 +1,287 @@
...
+@mock.patch("archivematica.storage_service.common.utils.generate_checksum")
+def test_command_to_download_aips(
+    generate_checksum: mock.Mock,
+    get_compressed_package_checksum: mock.Mock,
+    capsys: pytest.CaptureFixture[str],
+    compressed_package: models.Package,
+    uncompressed_package: models.Package,
+    aip_storage_fs_location: models.Location,
+) -> None:
+    uncompressed_package_checksum = "098f6bcd4621d373cade4e832627b4f9"
```
replaceafill: This is an `md5` checksum but the assertion below uses the `sha256` algorithm.

tests/common/test_command_populate_aip_checksums.py
```diff
@@ -0,0 +1,287 @@
...
+    compressed_package_checksum = (
+        "c2924159fcbbeadf8d7f3962b43ec1bf301e1b4f12dd28a8b89ec819f3714747"
+    )
+
+    algo = "sha256"
+    get_compressed_package_checksum.return_value = (
+        compressed_package_checksum,
+        algo,
+    )
+    total_packages = len(models.Package.objects.all())
```
replaceafill: Would `models.Package.objects.count()` work better here and the test below? Also you could inline the call in the f-string and remove the `total_packages` variables.

tests/common/test_command_populate_aip_checksums.py
```diff
@@ -0,0 +1,287 @@
...
+    )
+
+    call_command(
+        "populate_aip_checksums",
+        "--location-uuid",
+        aip_storage_fs_location.uuid,
+    )
+
+    captured = capsys.readouterr()
+    assert captured.out.split("\n")[:-1] == [
```
replaceafill: Convert these `captured.out.split("\n")` into `captured.out.splitlines()` calls. And why is the indexing `[:-1]` necessary?

replaceafill (changes requested):
This is looking good @Dhwaniartefact. I wonder if using real pointer file fixtures would make all those `get_compressed_package_checksum` mocks irrelevant.

tests/common/test_command_populate_aip_checksums.py
```diff
@@ -0,0 +1,287 @@
...
+
+import pytest
+from django.core.management import call_command
+from django.core.management.base import CommandError
+
+from archivematica.storage_service.locations import models
+
+
+@pytest.fixture
+@pytest.mark.django_db
```
replaceafill: Forgot to mention. This combination raises a warning in your tests. See https://docs.pytest.org/en/stable/deprecations.html#applying-a-mark-to-a-fixture-function

I'd recommend you to remove the `django_db` marks from the fixtures. They are only needed in the test functions.

Dhwaniartefact: oh yes, it does. Let me remove it from all the fixtures.

tests/common/test_command_populate_aip_checksums.py
Dhwaniartefact: Thanks, Let me change it

tests/common/test_command_populate_aip_checksums.py
Dhwaniartefact: oops. Thanks, Let me change it

tests/common/test_command_populate_aip_checksums.py
Dhwaniartefact: sure, let me give a comment here.

tests/common/test_command_populate_aip_checksums.py
Dhwaniartefact: sure.

tests/common/test_command_populate_aip_checksums.py
Dhwaniartefact: sure let change this assertion.

tests/common/test_command_populate_aip_checksums.py
```diff
@@ -0,0 +1,262 @@
...
+    assert captured.out.splitlines() == [
+        f"AIP {compressed_package.uuid} updated with {algo} checksum {compressed_package_checksum}",
+        f"AIP {uncompressed_package.uuid} updated with {algo} checksum {uncompressed_package_checksum}",
+        f"Complete. Checksums added for all {models.Package.objects.count()} identified AIPs.",
+    ]
+
+
+@pytest.mark.django_db
+@mock.patch("archivematica.storage_service.common.utils.generate_checksum")
+def test_command_fails_when_checksum_is_missing_for_uncompressed_aips(
```
replaceafill:
```suggestion
def test_command_fails_when_checksum_is_missing_for_uncompressed_aip(
```

To maintain consistency with the test above:

```python
def test_command_fails_when_checksum_is_missing_for_compressed_aip(
```

tests/common/test_command_populate_aip_checksums.py
```diff
@@ -0,0 +1,262 @@
...
+    captured = capsys.readouterr()
+    assert captured.out.splitlines() == [
+        f"AIP {uncompressed_package.uuid} updated with sha256 checksum {uncompressed_package_checksum}",
+        f"Complete. Checksums added for all {models.Package.objects.count()} identified AIPs.",
+    ]
+
+
+@pytest.mark.django_db
+@mock.patch("archivematica.storage_service.common.utils.generate_checksum")
+def test_command_updates_checksum_for_download_aips(
```
replaceafill: It's easy to understand what the two previous tests are doing:

```python
def test_command_updates_checksum_for_compressed_aip(
...
def test_command_updates_checksum_for_uncompressed_aip(
...
```

But what is a "download AIP" in this context?

```python
def test_command_updates_checksum_for_download_aips(
```

replaceafill (approved):

replaceafill:
@Dhwaniartefact Please use `Squash and merge`.

PR #768, opened 2025-03-31
Do not show user API keys

replaceafill:
This is similar to https://github.com/artefactual/archivematica/pull/2041

It updates the detail and edit view of pipelines to not show the API key of the user that connects with the Archivematica pipeline. The current plain text field has been replaced with a password field to do so.

The user edit view has been updated to have a `Regenerate API key` checkbox similar to the one in the Dashboard which replaces the `if you change your password, your API key will also be changed to a new, randomly generated one` approach.

Captura desde 2025-03-31 09-54-00

Finally the Storage Service user views and their redirects have been updated to show the API key only once on re-generation.  This uses Django's [messages framework](https://docs.djangoproject.com/en/4.2/ref/contrib/messages/) to render an alert with a button to copy the API key to the browser's clipboard.

Captura desde 2025-03-31 09-54-19

Captura desde 2025-03-31 09-54-35

Connected to https://github.com/archivematica/Issues/issues/1736

### DIFF
--- a/src/archivematica/storage_service/administration/forms.py
+++ b/src/archivematica/storage_service/administration/forms.py
@@ -293,7 +293,7 @@ class UserChangeForm(auth.forms.UserChangeForm):
     )
     regenerate_api_key = forms.BooleanField(
         widget=forms.CheckboxInput,
-        label="Regenerate API key (shown below)?",
+        label="Regenerate API key?",
         required=False,
     )
 
--- a/src/archivematica/storage_service/administration/forms.py
+++ b/src/archivematica/storage_service/administration/forms.py
@@ -291,6 +291,11 @@ class UserChangeForm(auth.forms.UserChangeForm):
         choices=roles.USER_ROLES,
         help_text=ROLE_FIELD_HELP_TEXT,
     )
+    regenerate_api_key = forms.BooleanField(
+        widget=forms.CheckboxInput,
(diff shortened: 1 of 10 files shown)


src/archivematica/storage_service/administration/forms.py
```diff
@@ -291,6 +291,11 @@ class UserChangeForm(auth.forms.UserChangeForm):
         choices=roles.USER_ROLES,
         help_text=ROLE_FIELD_HELP_TEXT,
     )
+    regenerate_api_key = forms.BooleanField(
+        widget=forms.CheckboxInput,
+        label="Regenerate API key (shown below)?",
```
Dhwaniartefact: Would it be more appropriate to write this label as `label="Regenerate API key?",` ?

replaceafill: If you're reviewing commit by commit I'd also suggest you check the end result of all the commits applied. This is how it currently looks in the branch:

https://github.com/artefactual/archivematica-storage-service/blob/fb5d44d8a79168cd3bcdeea776d846259620ba6f/src/archivematica/storage_service/administration/forms.py#L296

Dhwaniartefact: Aah ok, got it.

Dhwaniartefact (approved):
Great changes.

PR #770, opened 2025-04-10
Add ability to set user roles from OIDC claims 

sbreker:
Added new OIDC setting that will allow admins to configure whether user
roles will be set from the OIDC claims, or assigned the system default
role. The setting name is 'OIDC_OP_SET_ROLES_FROM_CLAIMS' and defaults
to False.

Added new OIDC setting that will configure the token path for retrieving
the role info. The setting name is 'OIDC_OP_ROLE_CLAIM_PATH' and
defaults to "realm_access.roles" which is the standard Keycloak location
for user role info.

Added new OIDC setting that will configure what information is extracted
from the OIDC token. The setting name is 'OIDC_ACCESS_ATTRIBUTE_MAP'
and the setting contents should be JSON-decodable.

These settings should be configured individually for each provider
appending the provider name for secondary providers as necessary.

### DIFF
--- a/src/archivematica/storage_service/common/backends.py
+++ b/src/archivematica/storage_service/common/backends.py
@@ -1,9 +1,11 @@
 import json
 from typing import Any
+from typing import Optional
 
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.core.exceptions import ImproperlyConfigured
+from django.http import HttpRequest
 from django_cas_ng.backends import CASBackend
 from josepy.jws import JWS
 from mozilla_django_oidc.auth import OIDCAuthenticationBackend
@@ -25,7 +27,25 @@ class CustomCASBackend(CASBackend):
 class CustomOIDCBackend(OIDCAuthenticationBackend):
     """Provide OpenID Connect authentication."""
 
-    def get_settings(self, attr, *args):
+    def __init__(self, *args: Any, **kwargs: Any) -> None:
+        super().__init__(*args, **kwargs)
+        # Store additional settings as instance attributes.
+        self.OIDC_OP_SET_ROLES_FROM_CLAIMS = getattr(
+            settings, "OIDC_OP_SET_ROLES_FROM_CLAIMS", False
+        )
+
+        self.OIDC_OP_ROLE_CLAIM_PATH = getattr(
+            settings, "OIDC_OP_ROLE_CLAIM_PATH", "realm_access.roles"
+        )
+
+        self.OIDC_ACCESS_ATTRIBUTE_MAP = getattr(
+            settings, "OIDC_ACCESS_ATTRIBUTE_MAP", settings.DEFAULT_OIDC_CLAIMS
+        )
+
+        # Extract valid role keys from USER_ROLES.
+        self.VALID_ROLES = {role[0] for role in roles.USER_ROLES}
+
+    def get_settings(self, attr: str, *args: Any) -> Any:
         if attr in [
             "OIDC_RP_CLIENT_ID",
             "OIDC_RP_CLIENT_SECRET",
@@ -34,6 +54,9 @@ class CustomOIDCBackend(OIDCAuthenticationBackend):
             "OIDC_OP_USER_ENDPOINT",
             "OIDC_OP_JWKS_ENDPOINT",
             "OIDC_OP_LOGOUT_ENDPOINT",
+            "OIDC_OP_SET_ROLES_FROM_CLAIMS",
+            "OIDC_OP_ROLE_CLAIM_PATH",
+            "OIDC_ACCESS_ATTRIBUTE_MAP",
         ]:
             # Retrieve the request object stored in the instance.
             request = getattr(self, "request", None)
@@ -55,13 +78,18 @@ class CustomOIDCBackend(OIDCAuthenticationBackend):
         # not in the list, call the superclass's get_settings method.
         return OIDCAuthenticationBackend.get_settings(attr, *args)
 
-    def authenticate(self, request, **kwargs):
+    def authenticate(self, request: HttpRequest, **kwargs: Any) -> Any:
         self.request = request
         self.OIDC_RP_CLIENT_ID = self.get_settings("OIDC_RP_CLIENT_ID")
         self.OIDC_RP_CLIENT_SECRET = self.get_settings("OIDC_RP_CLIENT_SECRET")
         self.OIDC_OP_TOKEN_ENDPOINT = self.get_settings("OIDC_OP_TOKEN_ENDPOINT")
         self.OIDC_OP_USER_ENDPOINT = self.get_settings("OIDC_OP_USER_ENDPOINT")
         self.OIDC_OP_JWKS_ENDPOINT = self.get_settings("OIDC_OP_JWKS_ENDPOINT")
+        self.OIDC_OP_SET_ROLES_FROM_CLAIMS = self.get_settings(
+            "OIDC_OP_SET_ROLES_FROM_CLAIMS"
+        )
+        self.OIDC_OP_ROLE_CLAIM_PATH = self.get_settings("OIDC_OP_ROLE_CLAIM_PATH")
+        self.OIDC_ACCESS_ATTRIBUTE_MAP = self.get_settings("OIDC_ACCESS_ATTRIBUTE_MAP")
 
         return super().authenticate(request, **kwargs)
 
@@ -84,7 +112,7 @@ class CustomOIDCBackend(OIDCAuthenticationBackend):
 
         info: dict[str, Any] = {}
 
-        for oidc_attr, user_attr in settings.OIDC_ACCESS_ATTRIBUTE_MAP.items():
+        for oidc_attr, user_attr in self.OIDC_ACCESS_ATTRIBUTE_MAP.items():
             if oidc_attr in access_info:
                 info.setdefault(user_attr, access_info[oidc_attr])
 
@@ -94,18 +122,59 @@ class CustomOIDCBackend(OIDCAuthenticationBackend):
 
         return info
 
-    def create_user(self, user_info: dict[str, Any]) -> User:
+    def create_user(self, user_info: dict[str, Any]) -> Optional[User]:
+        """Create a new user when authentication was successful."""
+        role = self.get_user_role(user_info)
+        if role is None:
+            return None
+
         user = super().create_user(user_info)
         for attr, value in user_info.items():
             setattr(user, attr, value)
-        self.set_user_role(user)
+        roles.set_user_role(user, role)
         return user
 
-    def update_user(self, user: User, user_info: dict[str, Any]) -> User:
-        self.set_user_role(user)
+    def update_user(self, user: User, user_info: dict[str, Any]) -> Optional[User]:
+        """
+        Updates the user's role only if the setting allows roles to be set from OIDC claims.
+        If the setting is False roles are being managed by an admin so do not update the role.
+        """
+        if self.OIDC_OP_SET_ROLES_FROM_CLAIMS:
+            role = self.get_user_role(user_info)
+            if role is None:
+                return None
+            roles.set_user_role(user, role)
         return user
 
-    def set_user_role(self, user: User) -> None:
-        # TODO: use user claims accessible via user's authentication tokens.
-        role = roles.promoted_role(roles.USER_ROLE_READER)
-        user.set_role(role)
+    def get_user_role(self, user_info: dict[str, Any]) -> Optional[str]:
+        """
+        Returns the highest-permission valid role found in the OIDC token claims.
+        Returns the default user role if the setting is False.
+        Returns None if no valid roles are found.
+        """
+        if not self.OIDC_OP_SET_ROLES_FROM_CLAIMS:
+            return roles.promoted_role(settings.DEFAULT_USER_ROLE)
+
+        claim_path = self.OIDC_OP_ROLE_CLAIM_PATH.split(".")
+        role_claims = user_info
+
+        # Traverse the claim path to find the role claims.
+        for key in claim_path:
+            if isinstance(role_claims, dict):
+                role_claims = role_claims.get(key, {})
+            else:
+                return None  # Malformed structure.
+
+        # If the claim contains a single role, convert to list.
+        if isinstance(role_claims, str):
+            role_claims = [role_claims]
+
+        if not isinstance(role_claims, list):
+            return None  # Neither a string nor a list of roles.
+
+        # Iterate over ordered USER_ROLES and return the first match.
+        for role_key, _ in roles.USER_ROLES:
+            if role_key in role_claims:
+                return roles.promoted_role(role_key)
+
+        return None  # No match found.
--- a/src/archivematica/storage_service/common/helpers.py
+++ b/src/archivematica/storage_service/common/helpers.py
@@ -1,6 +1,8 @@
+import json
 from collections.abc import Iterable
 from os import environ
 from typing import Any
+from typing import Union
 
 from django.core.exceptions import ImproperlyConfigured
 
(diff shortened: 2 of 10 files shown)


src/archivematica/storage_service/common/backends.py
```diff
@@ -95,17 +115,53 @@ def decode_token(token: str) -> Any:
         return info
 
     def create_user(self, user_info: dict[str, Any]) -> User:
+        """Create a new user when authentication was successful."""
+        role = self.get_user_role(user_info)
+        if role is None:
+            return None
```
replaceafill, line 129: This makes the return value of this method `Optional[User]` now.

src/archivematica/storage_service/common/backends.py
```diff
@@ -95,17 +115,53 @@ def decode_token(token: str) -> Any:
...
+        user.set_role(role)
         return user
 
     def update_user(self, user: User, user_info: dict[str, Any]) -> User:
-        self.set_user_role(user)
+        """Updates the user's role only if the setting allows roles to be set from OIDC claims."""
+        if self.OIDC_OP_SET_ROLES_FROM_CLAIMS:
+            role = self.get_user_role(user_info)
+            if role is None:
+                return None
```
replaceafill, line 145: This makes the return value of this method `Optional[User]` now.

src/archivematica/storage_service/common/backends.py
```diff
@@ -25,7 +26,20 @@ def configure_user(self, user):
 class CustomOIDCBackend(OIDCAuthenticationBackend):
     """Provide OpenID Connect authentication."""
 
-    def get_settings(self, attr, *args):
+    def __init__(self, *args, **kwargs) -> None:
```
replaceafill:
```suggestion
    def __init__(self, *args: Any, **kwargs: Any) -> None:
```

src/archivematica/storage_service/common/backends.py
```diff
@@ -25,7 +26,20 @@ def configure_user(self, user):
...
+            settings, "OIDC_OP_SET_ROLES_FROM_CLAIMS", False
+        )
+        self.OIDC_OP_ROLE_CLAIM_PATH = getattr(
+            settings, "OIDC_OP_ROLE_CLAIM_PATH", "realm_access.roles"
+        )
+
+        # Extract valid role keys from USER_ROLES.
+        self.VALID_ROLES = {role[0] for role in roles.USER_ROLES}
+
+    def get_settings(self, attr, *args) -> Any:
```
replaceafill:
```suggestion
    def get_settings(self, attr: str, *args: Any) -> Any:
```

src/archivematica/storage_service/common/backends.py
```diff
@@ -55,13 +71,17 @@ def get_settings(self, attr, *args):
         # not in the list, call the superclass's get_settings method.
         return OIDCAuthenticationBackend.get_settings(attr, *args)
 
-    def authenticate(self, request, **kwargs):
+    def authenticate(self, request, **kwargs) -> Any:
```
replaceafill:
```suggestion
from django.http import HttpRequest
...
    def authenticate(self, request: HttpRequest, **kwargs: Any) -> Any:
```

src/archivematica/storage_service/common/backends.py
```diff
@@ -95,17 +115,53 @@ def decode_token(token: str) -> Any:
...
         return user
 
     def update_user(self, user: User, user_info: dict[str, Any]) -> User:
-        self.set_user_role(user)
+        """Updates the user's role only if the setting allows roles to be set from OIDC claims."""
+        if self.OIDC_OP_SET_ROLES_FROM_CLAIMS:
+            role = self.get_user_role(user_info)
+            if role is None:
+                return None
+            user.set_role(role)
```
replaceafill: Why is setting the role of the user now dependent on the `OIDC_OP_SET_ROLES_FROM_CLAIMS` setting?

> Added new OIDC setting that will allow admins to configure whether user
roles will be set from the OIDC claims, or assigned the system default
role. 

If I understand the new logic of the method correctly if the setting is disabled this will not set the system default role?

Also checking for the setting here and then calling `self.get_user_role` which also checks the setting seems unnecessary.

src/archivematica/storage_service/common/helpers.py
```diff
@@ -20,7 +20,7 @@ def is_true(env_str: str) -> bool:
 
 def get_oidc_secondary_providers(
     oidc_secondary_provider_names: Iterable[str],
-) -> dict[str, dict[str, str]]:
+) -> dict[str, dict[str, object]]:
```
replaceafill:
```suggestion
from typing import Union
...
) -> dict[str, dict[str, Union[str, bool]]]:
```

src/archivematica/storage_service/storage_service/settings/base.py
```diff
@@ -659,7 +666,13 @@ def _get_email(email):
...
 
     # map attributes from access token
-    OIDC_ACCESS_ATTRIBUTE_MAP = {"given_name": "first_name", "family_name": "last_name"}
+    default_claims = {"given_name": "first_name", "family_name": "last_name"}
+    try:
+        OIDC_ACCESS_ATTRIBUTE_MAP = json.loads(
+            environ.get("OIDC_ACCESS_ATTRIBUTE_MAP", default_claims)
+        )
+    except json.JSONDecodeError:
+        OIDC_ACCESS_ATTRIBUTE_MAP = default_claims
```
replaceafill: Should the README state that the value of the map needs to be JSON-encodable?

src/archivematica/storage_service/common/backends.py
```diff
@@ -95,17 +115,53 @@ def decode_token(token: str) -> Any:
...
+                    role = role.get(key)
+                else:
+                    role = None
+                    break
+
+            # If role is a list, find the first valid one
+            if isinstance(role, list):
+                for r in role:
+                    if r in self.VALID_ROLES:
+                        return roles.promoted_role(r)
```
replaceafill: Do providers like Keycloak guarantee order of the realm roles of the user? If I understand this correctly it'll return the first role it matches not necessarily the one with more permissions. Should that fact be stated in the README?

replaceafill (changes requested):
Thanks @sbreker, I really appreciate the new tests for this new feature!

src/archivematica/storage_service/common/backends.py
sbreker: Updated

src/archivematica/storage_service/common/backends.py
sbreker: Updated

src/archivematica/storage_service/common/backends.py
sbreker: Updated

src/archivematica/storage_service/common/helpers.py
sbreker: I cannot seem to get this to work. With the latest commit I get the error: `src/archivematica/storage_service/common/helpers.py:58: error: Incompatible return value type (got "dict[str, dict[str, object]]", expected "dict[str, dict[str, Union[str, bool]]]")  [return-value]`

src/archivematica/storage_service/storage_service/settings/base.py
sbreker: I have added this note

src/archivematica/storage_service/common/backends.py
sbreker: Fixed - see latest commit.

src/archivematica/storage_service/common/backends.py
sbreker: Fixed - see latest commit.

src/archivematica/storage_service/common/backends.py
sbreker: When `OIDC_OP_SET_ROLES_FROM_CLAIMS` is `False` and roles are not being set from token claims, when the user is first created it will be assigned the system default role. A Storage Service admin can then update this role as needed. In this scenario the role should not be updated (in `update_user()`) by the OIDC authentication process the next time the user logs in - the admin will want these roles to stay set as configured.

src/archivematica/storage_service/common/backends.py
sbreker: This is a good point and I was wondering how best to handle this. See latest commit - I have altered `get_user_role()` so that it returns the highest-permission valid role found in the OIDC token claims based on the order of the roles in `USER_ROLES` in `roles.py` - (they are ordered in descending level of permission in the definition of `USER_ROLES`).

src/archivematica/storage_service/common/helpers.py
replaceafill: Ah, I see. I think `mypy` cannot infer the provider config value. In those cases I usually create a separate type for the problematic value and annotate it explicitly. Something like this:

```diff
diff --git a/src/archivematica/storage_service/common/helpers.py b/src/archivematica/storage_service/common/helpers.py
index 1f391a4..f7e9f60 100644
--- a/src/archivematica/storage_service/common/helpers.py
+++ b/src/archivematica/storage_service/common/helpers.py
@@ -1,6 +1,7 @@
 from collections.abc import Iterable
 from os import environ
 from typing import Any
+from typing import Union
 
 from django.core.exceptions import ImproperlyConfigured
 
@@ -18,9 +19,12 @@ def is_true(env_str: str) -> bool:
     return env_str.lower() in ["true", "yes", "on", "1"]
 
 
+ProviderConfig = dict[str, Union[str, bool]]
+
+
 def get_oidc_secondary_providers(
     oidc_secondary_provider_names: Iterable[str],
-) -> dict[str, dict[str, object]]:
+) -> dict[str, ProviderConfig]:
     providers = {}
 
     for provider_name in oidc_secondary_provider_names:
@@ -42,7 +46,7 @@ def get_oidc_secondary_providers(
         )
 
         if client_id and client_secret:
-            providers[provider_name] = {
+            provider_config: ProviderConfig = {
                 "OIDC_RP_CLIENT_ID": client_id,
                 "OIDC_RP_CLIENT_SECRET": client_secret,
                 "OIDC_OP_AUTHORIZATION_ENDPOINT": authorization_endpoint,
@@ -53,5 +57,6 @@ def get_oidc_secondary_providers(
                 "OIDC_OP_SET_ROLES_FROM_CLAIMS": set_roles_from_claims,
                 "OIDC_OP_ROLE_CLAIM_PATH": role_claim_path,
             }
+            providers[provider_name] = provider_config
 
     return providers
```

src/archivematica/storage_service/common/backends.py
replaceafill: Understood and I appreciate the new test!

Should we add a comment above `roles.USER_ROLES` to state the list should maintain a descending level of permissions from now on?

src/archivematica/storage_service/common/backends.py
replaceafill: Ah, I see. I think it makes sense that the `update_user` method doesn't update the role in that case.

replaceafill (approved):
I left a couple of final suggestions but this looks good @sbreker :+1:

src/archivematica/storage_service/common/backends.py
sbreker: Good idea - I will investigate.

sbreker: Added comment in recent commit.

src/archivematica/storage_service/common/helpers.py
sbreker: Ah that did the trick...thanks @replaceafill! Fixed in latest commit...

replaceafill (approved):

src/archivematica/storage_service/common/backends.py
```diff
@@ -38,6 +38,11 @@ def __init__(self, *args: Any, **kwargs: Any) -> None:
             settings, "OIDC_OP_ROLE_CLAIM_PATH", "realm_access.roles"
         )
 
+        default_claims = {"given_name": "first_name", "family_name": "last_name"}
```
replaceafill: This default `default_claims` is set in three different places.

What do you think about using its value only from the `archivematica.storage_service.storage_service.settings.base` module?

Suppose it's renamed and defined earlier in the module:

```diff
--- a/src/archivematica/storage_service/storage_service/settings/base.py
+++ b/src/archivematica/storage_service/storage_service/settings/base.py
@@ -644,6 +644,9 @@ if OIDC_AUTHENTICATION:
         "OIDC_OP_ROLE_CLAIM_PATH", "realm_access.roles"
     )
 
+    # map attributes from access token
+    DEFAULT_OIDC_CLAIMS = {"given_name": "first_name", "family_name": "last_name"}
+
     OIDC_SECONDARY_PROVIDER_NAMES = environ.get(
         "OIDC_SECONDARY_PROVIDER_NAMES", ""
     ).split(",")
@@ -670,14 +673,12 @@ if OIDC_AUTHENTICATION:
 
     OIDC_USERNAME_ALGO = _get_email
 
-    # map attributes from access token
-    default_claims = {"given_name": "first_name", "family_name": "last_name"}
     try:
         OIDC_ACCESS_ATTRIBUTE_MAP = json.loads(
-            environ.get("OIDC_ACCESS_ATTRIBUTE_MAP", json.dumps(default_claims))
+            environ.get("OIDC_ACCESS_ATTRIBUTE_MAP", json.dumps(DEFAULT_OIDC_CLAIMS))
         )
     except json.JSONDecodeError:
-        OIDC_ACCESS_ATTRIBUTE_MAP = default_claims
+        OIDC_ACCESS_ATTRIBUTE_MAP = DEFAULT_OIDC_CLAIMS
 
     # map attributes from id token
     OIDC_ID_ATTRIBUTE_MAP = {"email": "email"}
```

Then the `archivematica.storage_service.common.backends` module could use it like this:

```python
        self.OIDC_ACCESS_ATTRIBUTE_MAP = getattr(
            settings, "OIDC_ACCESS_ATTRIBUTE_MAP", settings.DEFAULT_OIDC_CLAIMS
        )
```

And since the `archivematica.storage_service.common.helpers` module does not access the Django settings it could receive it as a parameter in `get_oidc_secondary_providers` like this:

```python
def get_oidc_secondary_providers(
    oidc_secondary_provider_names: Iterable[str],
    default_oidc_claims: dict[str, str],
) -> dict[str, ProviderConfig]:
```

And it'd be passed from the `archivematica.storage_service.storage_service.settings.base` module like this:

```python
    OIDC_PROVIDERS = get_oidc_secondary_providers(OIDC_SECONDARY_PROVIDER_NAMES, DEFAULT_OIDC_CLAIMS)
```

Let me know if it makes sense.

tests/integration/docker-compose.yml
```diff
@@ -56,6 +56,8 @@ services:
       OIDC_OP_LOGOUT_ENDPOINT_SECONDARY: "http://keycloak:8080/realms/secondary/protocol/openid-connect/logout"
       OIDC_OP_SET_ROLES_FROM_CLAIMS_SECONDARY: "true"
       OIDC_OP_ROLE_CLAIMS_PATH_SECONDARY: "realm_access.roles"
```
replaceafill: I just noticed. Should this be named `OIDC_OP_ROLE_CLAIM_PATH_SECONDARY`?

replaceafill (changes requested):

src/archivematica/storage_service/common/backends.py
sbreker: Great idea - thank you for the detailed feedback. I have implemented this change in a new commit.

tests/integration/docker-compose.yml
sbreker: Good catch! Fixed in latest commit.

src/archivematica/storage_service/storage_service/settings/base.py
replaceafill: Sorry for being misleading here @sbreker, but I think the note should read `JSON-decodable`?

replaceafill (approved):
Looks great @sbreker! Sorry about the JSON-encodable/decodable confusion :blush:

src/archivematica/storage_service/storage_service/settings/base.py
sbreker: Updated in latest commit :+1:

PR #772, opened 2025-04-24
Fixed the Problem(Step 1): Dataverse Transfer Fails at "Parse Dataverse METS …

Meghanxuxx:
Issue Related: https://github.com/archivematica/Issues/issues/1745

After reviewing the code, I found that we don't really need the [extract_and_remove_bundle](https://github.com/artefactual/archivematica-storage-service/blob/ea58d030ae54642f60ee206b20129d97bda60380/src/archivematica/storage_service/locations/models/dataverse.py#L286) step for the extraction. The [extract_contents.py](https://github.com/artefactual/archivematica/blob/ad01b681f5bcd568d235121b7daaf2601a8f824f/src/archivematica/MCPClient/clientScripts/extract_contents.py#L214) already recursively looks for files that need to be extracted,  so we can finish all the extraction in this step. Additionally, this file seems to only remove the "outermost" package, while keeping the default setting  `delete` to be False for files other than the "outermost". This helps prevent RData files from being deleted once they have been unpacked.

For this case, the code change in this PR is to remove the extract_and_remove_bundle function.

### DIFF
--- a/src/archivematica/storage_service/locations/models/dataverse.py
+++ b/src/archivematica/storage_service/locations/models/dataverse.py
@@ -9,9 +9,9 @@ import requests
 from django.db import models
 from django.utils.translation import gettext_lazy as _
 
-from archivematica.storage_service.locations.models import StorageException
-from archivematica.storage_service.locations.models.location import Location
-from archivematica.storage_service.locations.models.urlmixin import URLMixin
+from . import StorageException
+from .location import Location
+from .urlmixin import URLMixin
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


replaceafill:
@Meghanxuxx I think you need to rebase your pull request branch on top of the latest `qa/0.x` branch. For example, all of our imports in Python are absolute now.

Also notice [the linting job is failing](https://github.com/artefactual/archivematica-storage-service/actions/runs/14649318352/job/41119179687?pr=772) in the test workflow. We use [pre-commit](https://pre-commit.com/) for linting the Archivematica code bases. Once you install it you can run `pre-commit run --all-files` on the top of any Archivematica repository and it should automatically fix linting problems or point where the errors are coming from.

sarah-mason:
Hi @Meghanxuxx - as I mentioned in your Archivematica PRs, we'll be starting testing in a little over a week for 1.18 so we'll stop accepting contributions around then. If you want to address the failed tests in your PRs and have any questions, please let us know. Thank you for your contributions!

PR #785, opened 2025-07-11
Do not show private data in GPG encryption key views

replaceafill:
This update removes the `UIDs` and `Private Key ASCII Armor` fields from all `GPG` encryption key views and uses the `Keyid` field as the key identifier throughout the user interface.

Connected to https://github.com/archivematica/Issues/issues/1737


sevein (approved):
👍

PR #788, opened 2025-08-06
Add current password validation to user edit form

replaceafill:
This updates the user edit form to require the current password from
the user setting a new password.

Connected to https://github.com/archivematica/Issues/issues/1757


src/archivematica/storage_service/administration/views.py
```diff
@@ -119,7 +119,7 @@ def user_edit(request, id):
         password_form.save()
         messages.success(request, _("Password changed."))
         return redirect("administration:user_edit", id=edit_user.pk)
-    elif "password":
+    else:
```
sevein, line 123: 🤔 😆

sevein (approved):
LGTM

PR #791, opened 2025-08-20
Remove warnings about autocreated index fields

scollazo:
Removes models.W042 warning when running manage.py commands

Connected to https://github.com/archivematica/Issues/issues/1759

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -223,6 +223,12 @@ of these settings or provide values to mandatory fields.
   - **Type:** `integer`
   - **Default:** `900`
 
+- **`SS_SILENCED_SYSTEM_CHECKS`**:
+  - **Description:**  comma-separated list of ignored system checks.
+    e.g. mysql.W002,models.W042
+  - **Type:** `string`
+  - **Default:** `None`
+
 The configuration of the database is also declared via environment variables.
 Storage Service looks up the `SS_DB_URL` environment string. If defined, its
 value is expected to follow the form described in the [dj-database-url docs],
--- a/src/archivematica/storage_service/storage_service/settings/base.py
+++ b/src/archivematica/storage_service/storage_service/settings/base.py
@@ -754,3 +754,5 @@ if PROMETHEUS_ENABLED:
     )
     INSTALLED_APPS = INSTALLED_APPS + ["django_prometheus"]
     LOGIN_EXEMPT_URLS.append(r"^metrics$")
+
+SILENCED_SYSTEM_CHECKS = environ.get("SS_SILENCED_SYSTEM_CHECKS", "[]")
--- a/src/archivematica/storage_service/storage_service/settings/base.py
+++ b/src/archivematica/storage_service/storage_service/settings/base.py
@@ -754,6 +754,3 @@ if PROMETHEUS_ENABLED:
     )
     INSTALLED_APPS = INSTALLED_APPS + ["django_prometheus"]
     LOGIN_EXEMPT_URLS.append(r"^metrics$")
-
-# Remove models.W042 warning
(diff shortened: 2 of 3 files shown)


src/archivematica/storage_service/storage_service/settings/base.py
```diff
@@ -746,3 +746,6 @@ def _get_email(email):
     )
     INSTALLED_APPS = INSTALLED_APPS + ["django_prometheus"]
     LOGIN_EXEMPT_URLS.append(r"^metrics$")
+
+# Remove models.W042 warning
+DEFAULT_AUTO_FIELD = "django.db.models.AutoField"
```
replaceafill: See https://github.com/artefactual/archivematica/pull/2097#discussion_r2288326605

scollazo:
@replaceafill , I removed the defalt_auto_field and used systemd_silence_checks only for production. Does this look better?

replaceafill:
@scollazo I think instead of hardcoding the setting value we should expose the `SILENCED_SYSTEM_CHECKS` Django setting the same way expose other Django settings: with [application variables](https://github.com/artefactual/archivematica/blob/8df5aeeadcdd23b58946b36df45b11de75b60267/src/archivematica/dashboard/install/README.md#application-variables) in the Dashboard and [application-specific environment variables](https://github.com/artefactual/archivematica-storage-service/tree/122f1fcc823b119b089798599a4dec50e556ffca/install#application-specific-environment-variables) in the Storage Service. That way we could document their default values and explain where those codes (i.e. `models.W042`) come from and allow users to customize them.

scollazo:
Done!

install/README.md
```diff
@@ -223,6 +223,12 @@ of these settings or provide values to mandatory fields.
   - **Type:** `integer`
   - **Default:** `900`
 
+- **`SS_SILENCED_SYSTEM_CHECKS`**:
```
replaceafill, line 226: What do you think of dropping `SS_` prefix? We are not really consistent but most of the Django related variables don't have it. Also could we move this documentation bit after the `SECRET_KEY` variable?

install/README.md
```diff
@@ -223,6 +223,12 @@ of these settings or provide values to mandatory fields.
...
+  - **Description:**  comma-separated list of ignored system checks.
+    e.g. mysql.W002,models.W042
```
replaceafill, line 228: Let's add a link to https://docs.djangoproject.com/en/4.2/ref/settings/#silenced-system-checks here. See examples in `TIME_ZONE` or `SECRET_KEY` above.

src/archivematica/storage_service/storage_service/settings/base.py
```diff
@@ -754,3 +754,5 @@ def _get_email(email):
     )
     INSTALLED_APPS = INSTALLED_APPS + ["django_prometheus"]
     LOGIN_EXEMPT_URLS.append(r"^metrics$")
+
+SILENCED_SYSTEM_CHECKS = environ.get("SS_SILENCED_SYSTEM_CHECKS", "[]")
```
replaceafill, line 758: If you're expecting the user to set a comma separated list it needs to be converted to a list here:

```suggestion
SILENCED_SYSTEM_CHECKS = environ.get("SS_SILENCED_SYSTEM_CHECKS", "").split(",")
```

install/README.md
```diff
@@ -223,6 +223,12 @@ of these settings or provide values to mandatory fields.
...
+  - **Type:** `string`
+  - **Default:** `None`
```
replaceafill, line 230:
```suggestion
  - **Default:** `""`
```

replaceafill (changes requested):

PR #795, opened 2025-09-02
Add dependabot.yml config file

sevein:
Relates to https://github.com/artefactual/archivematica/pull/2102.

For now this is still being evaluated, so the focus is only on image versions listed in `/tests/integration/docker-compose.yml`.


replaceafill (approved):
Thank you @sevein! :tada:

PR #802, opened 2025-09-24
Update GPG key label in space forms and views

replaceafill:
This change updates the `GPG.key` field to use the Keyid label in space forms and views, aligning it with the label used in the encryption key views.

Connected to https://github.com/archivematica/Issues/issues/1737

### DIFF
--- a/src/archivematica/storage_service/locations/forms.py
+++ b/src/archivematica/storage_service/locations/forms.py
@@ -193,7 +193,9 @@ class GPGForm(forms.ModelForm):
         super().__init__(*args, **kwargs)
         system_key = gpgutils.get_default_gpg_key(gpgutils.get_gpg_key_list())
         self.fields["key"] = forms.ChoiceField(
-            choices=get_gpg_key_choices(), initial=system_key["fingerprint"]
+            choices=get_gpg_key_choices(),
+            initial=system_key["fingerprint"],
+            label=_("Keyid"),
         )
 
     class Meta:
--- a/src/archivematica/storage_service/locations/views.py
+++ b/src/archivematica/storage_service/locations/views.py
@@ -629,12 +629,7 @@ def space_list(request):
         child_dict_raw = model_to_dict(
             child, PROTOCOL[space.access_protocol]["fields"] or [""]
         )
-        child_dict = {
-            child._meta.get_field(field).verbose_name: get_child_space_value(
-                value, field, child
-            )
-            for field, value in child_dict_raw.items()
-        }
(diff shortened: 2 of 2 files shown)


src/archivematica/storage_service/locations/forms.py
```diff
@@ -193,7 +193,9 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         system_key = gpgutils.get_default_gpg_key(gpgutils.get_gpg_key_list())
         self.fields["key"] = forms.ChoiceField(
-            choices=get_gpg_key_choices(), initial=system_key["fingerprint"]
+            choices=get_gpg_key_choices(),
+            initial=system_key["fingerprint"],
+            label=_("Keyid"),
```
sevein, line 198: To me [Key ID](https://datatracker.ietf.org/doc/html/rfc4880#section-3.3) sounds more correct after looking at the spec.

src/archivematica/storage_service/locations/views.py
```diff
@@ -667,13 +667,9 @@ def get_child_space_label(child, field):
 
 
 def get_child_space_value(value, field, child):
-    """If the child space ``child`` is a GPG instance, and the field is key, we
-    return a human-readable representation of the GPG key instead of its
-    fingerprint: the name of the first user in its uids list.
-    """
```
sevein, line 670: Agree, it was too obvious, but maybe keep a tiny docstring on the why, e.g.:

    Show shorter Key ID instead of full fingerprint for GPG keys.

src/archivematica/storage_service/locations/views.py
```diff
@@ -667,13 +667,9 @@ def get_child_space_label(child, field):
...
     if field == "key" and isinstance(child, GPG):
         key = gpgutils.get_gpg_key(value)
-        return key["keyid"]
+        return key["keyid"] if key is not None else _("Keyid not found")
```
sevein: Reading the docstring earlir totally reminded me of @jrwdunham (👋) and his exhaustive code comments. He was also great at writing functional-y constructs. Maybe we should rewrite this as below to keep his spirit!

    return (key or {}).get("keyid", _("Keyid not found")).

sevein (approved):
LGTM

src/archivematica/storage_service/locations/forms.py
replaceafill: Yes, I noticed it during https://github.com/artefactual/archivematica-storage-service/pull/785 but decided to keep it like it was originally implemented https://github.com/artefactual/archivematica-storage-service/pull/198

PR #808, opened 2025-10-17
Paginate S3 browse listings by prefix

replaceafill:
This updates the S3 Space browse implementation to use a boto3
paginator with a delimiter so only the current prefix is inspected.

The new integration test records S3 keys and asserts that the browse
location API endpoint returns only the requested level's directories
and files.

Connected to https://github.com/archivematica/Issues/issues/1755


sevein (approved):
LGTM

PR #813, opened 2025-10-30
Add API endpoint for reviewing AIP deletion requests

replaceafill:
This change refactors the package request handling logic by extracting it from the views and API resources modules within the locations package into a dedicated module, enabling it to be shared across views, API endpoints, and management commands.

Additionally, it introduces a new API endpoint in the package resource that allows authorized users to approve or reject package deletion requests. Access to this endpoint is restricted by the same permission used in the “View delete requests” page of the user interface.

The API documentation has been updated in https://github.com/artefactual/archivematica-docs/pull/489

**Note**: while returning `message` on success and `error_message` on failure isn’t an ideal REST practice, this endpoint follows the convention already used by other Storage Service API endpoints to keep responses consistent.

Connected to https://github.com/archivematica/Issues/issues/1696


sevein (approved):
👍

PR #849, opened 2026-01-29
Add integration tests for admin GPG flows

replaceafill:
This adds integration tests that drive key creation, detail rendering,
import and deletion through the administration views using the real
GnuPG binary. This covers the end to end UI flow and passphrase
rejection that unit tests do not.

Connected to https://github.com/archivematica/Issues/issues/1781


sevein (approved):
LGTM

PR #854, opened 2026-02-04
Check rclone container existence more efficiently

liam-lloyd:
In the rclone model's _ensure_container_exists method, rclone ls is used to check whether a container exists on an rclone remote. If not, rclone mkdir is used to create it. rclone ls is recursive by default, however, so this call can take minutes if the container is large. This commit replaces the rclone ls call with rclone lsf --dirs-only, which lists only root level directories and therefore runs much faster.

### DIFF
--- a/src/archivematica/storage_service/locations/models/rclone.py
+++ b/src/archivematica/storage_service/locations/models/rclone.py
@@ -119,17 +119,25 @@ class RClone(models.Model):
         raise StorageException("rclone remote matching %s not found", self.remote_name)
 
     def _ensure_container_exists(self):
-        """Ensure that the S3 bucket or other container exists by asking it
-        something about itself. If we cannot retrieve metadata about it then
-        we attempt to create the bucket, else, we raise a StorageException.
+        """Ensure that the S3 bucket or other container exists by creating it if it doesn't.
+        If the creation attempt fails, we raise a StorageException.
         """
         LOGGER.debug("Test that container '%s' exists", self.container)
-        prefixed_container_name = f"{self.remote_prefix}{self.container}"
-        cmd = ["ls", prefixed_container_name]
+        remote_name = f"{self.remote_prefix}"
+        cmd = ["lsf", "--dirs-only", remote_name]
         try:
-            self._execute_rclone_subcommand(cmd)
+            containers_list = self._execute_rclone_subcommand(cmd).split("\n")
+            container_exists = False
+            for container in containers_list:
+                if container.rstrip("/") == self.container:
+                    container_exists = True
+            if not container_exists:
+                raise StorageException(
+                    "rclone container matching %s not found", self.container
+                )
         except StorageException:
             LOGGER.info("Creating container '%s'", self.container)
+            prefixed_container_name = f"{self.remote_prefix}{self.container}"
             create_container_cmd = ["mkdir", prefixed_container_name]
             try:
                 self._execute_rclone_subcommand(create_container_cmd)
--- a/tests/locations/test_rclone.py
+++ b/tests/locations/test_rclone.py
@@ -163,7 +163,9 @@ def test_rclone_ensure_container_exists(
     else:
         with pytest.raises(models.StorageException):
             rclone_space._ensure_container_exists()
-            subprocess.assert_called_with(["mkdir", "testremote:testcontainer"])
+
+    args, _ = subprocess.Popen.call_args
+    assert args[0] == ["rclone", "mkdir", "testremote:testcontainer"]
 
 
 @pytest.mark.parametrize(


src/archivematica/storage_service/locations/models/rclone.py
```diff
@@ -119,17 +119,25 @@ def remote_prefix(self):
...
-        something about itself. If we cannot retrieve metadata about it then
-        we attempt to create the bucket, else, we raise a StorageException.
+        """Ensure that the S3 bucket or other container exists by creating it if it doesn't.
+        If the creation attempt fails, we raise a StorageException.
         """
         LOGGER.debug("Test that container '%s' exists", self.container)
-        prefixed_container_name = f"{self.remote_prefix}{self.container}"
-        cmd = ["ls", prefixed_container_name]
+        remote_name = f"{self.remote_prefix}"
+        cmd = ["lsf", "--dirs-only", remote_name]
```
replaceafill: After some testing on AWS, this turns out to add a permission requirement: the call needs `s3:ListAllMyBuckets`, which the `ls <remote>:<container>` it replaces didn't (that only needed `s3:ListBucket` on the container).

Some context on why that matters to us: we don't document the permissions an rclone remote over S3 needs, so the only guidance an administrator has today is the IAM policy in the [S3 space documentation](https://www.archivematica.org/en/docs/storage-service-0.24/administrators/#s3-amazon): `s3:PutObject`, `s3:GetObject`, `s3:ListBucket`, `s3:DeleteObject` and `s3:GetBucketLocation` on the bucket ARN and its objects, with `s3:CreateBucket` described as optional. `s3:ListAllMyBuckets` isn't in that policy and can't be added to it as written, since it only takes effect with `"Resource": "*"`, which widens a single-bucket key to the whole account. Users may well have set up their rclone remotes differently, but a key following that policy is the least-privilege setup we point people to, and for such a key this call fails every time:

```
2026/08/30 06:57:33 ERROR : : error listing: AccessDenied: User: arn:aws:iam::000000000000:user/myuser is not authorized to perform: s3:ListAllMyBuckets because no identity-based policy allows the s3:ListAllMyBuckets action
	status code: 403, request id: XXXXXXXXXXXXXXXX, host id: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
2026/08/30 06:57:33 Failed to lsf with 2 errors: last error was: error in ListJSON: AccessDenied: User: arn:aws:iam::000000000000:user/myuser is not authorized to perform: s3:ListAllMyBuckets because no identity-based policy allows the s3:ListAllMyBuckets action
	status code: 403, request id: XXXXXXXXXXXXXXXX, host id: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```

The `except StorageException` then treats the denial like a missing container and runs `rclone mkdir` on every call, which needs `s3:CreateBucket`. IAM denies that before S3 checks whether the bucket exists, so an existing bucket the account owns still gets:

```
2026/08/30 06:57:34 Failed to mkdir: AccessDenied: User: arn:aws:iam::000000000000:user/myuser is not authorized to perform: s3:CreateBucket on resource: "arn:aws:s3:::mybucket" because no identity-based policy allows the s3:CreateBucket action
	status code: 403, request id: XXXXXXXXXXXXXXXX, host id: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
```

replaceafill (commented):
Thanks for the alternative approach @liam-lloyd. I reproduced the slowdown on my side: with ~30,000 keys in the bucket, the `rclone ls` in `_ensure_container_exists` dominates the latency of every rclone-space operation, not just replication, and this change does fix that.

The problem is that the new check needs permissions the operations themselves don't use: `s3:ListAllMyBuckets` for the listing and, once that is denied, `s3:CreateBucket` for the `mkdir` fallback. We don't document the permissions an rclone remote over S3 needs, so the reference point we have is the IAM policy documented for the S3 space, which grants five actions scoped to a single bucket and includes neither, so a key following it fails the check every time. I tested this against a real AWS bucket in `ca-central-1` with that exact policy: browsing a location in the space returns HTTP 200 on `qa/0.x` and HTTP 500 (`Unable to find or create container`) on this branch. Retrieve and delete go through the same check, so they change the same way.

None of this fails if the key has `s3:CreateBucket`, since AWS then answers the `mkdir` with a 409 that rclone ignores. So it affects keys without `s3:CreateBucket`, which is the setup the S3 space documentation implies by calling it optional. Users may have configured their rclone remotes with broader keys, but we can't assume they have.

This permission problem can't be replicated in CI with SeaweedFS: it returns a filtered bucket list for a restricted identity instead of denying the call, so `test_browsing_an_rclone_transfer_source_location_works_with_limited_permissions` finds the container and passes on this branch.

Would it be worth removing `_ensure_container_exists` altogether instead of looking for a cheaper check? I tried that locally against AWS and an S3-compatible backend: browse under the policy above returns 200 again, and storing is unchanged, since rclone's upload already tries to create the destination bucket with or without the check. `store_aip` then does no listing at all, which fixes the slowdown more completely than trading the listing of every object for a listing of every bucket. I only exercised browse and store that way, not retrieve and delete.

Either way, my concern is the permission requirement on the read paths rather than the approach itself. Let me know what you think.

PR #862, opened 2026-02-23
Add variable to toggle transfer threading in S3 space

replaceafill:
This makes boto3 threading configurable via an application-specific environment variable that allows operators to disable managed transfer threads when concurrency causes instability, throttling, or resource pressure in constrained deployments and S3-compatible backends.


sevein (approved):
👍

PR #871, opened 2026-03-04
Migrate non-server-side DataTables to Vue/TanStack

replaceafill:
This replaces non-server-side DataTables with a Vue/TanStack tables app, using backend-generated JSON payloads mounted into existing Django pages. It migrates administration tables, location list tables, and package request tables, including per-row approve/reject forms.


sevein (approved):
I know you're still figuring out the CI build, but other than that the code looks great to me and works wonderfully!

PR #873, opened 2026-03-06
Remove jQuery

replaceafill:
This removes jQuery from the frontend by migrating legacy DataTables
to the shared Vue tables app and replacing the remaining jQuery-driven
UI flows with frontend features.

It also cleans up obsolete includes/assets (including legacy IE
support), updates template/i18n usage, and extends tests to ensure the
new behavior remains stable.


sevein (approved):
LGTM

PR #896, opened 2026-03-27
Fix Browse button for spaces with empty roots

replaceafill:
This prevents the Browse button in location forms from failing when a space has no root path, as with remote storage backends such as S3.


sevein (approved):
👍

PR #920, opened 2026-04-29
Avoid cookie name collisions with Archivematica

replaceafill:
This change adds Storage Service-specific names for the CSRF and language cookies so they do not collide with Archivematica's Django cookie defaults when both apps share a host.


sevein (approved):
👍

PR #929, opened 2026-05-13
Fix replica pointers for remote AIPs

replaceafill:
During replication, the `Package` model calculates the replica
validation checksum through `self.get_local_path()`. That only works
when the main package is locally accessible; for remote storage, the
replica package content is already available in the destination
space's staging path instead.

This change uses the staged replica path for checksum validation, and
makes pointer lookup depend on the package's saved pointer location
and path. This avoids treating a remote AIP object key as a missing
local file when reading an existing internal pointer XML.

Connected to https://github.com/archivematica/Issues/issues/1796
Connected to https://github.com/archivematica/Issues/issues/1342


sevein (approved):
👍

PR #936, opened 2026-05-26
Capture OSErrors to avoid permission denied errors

NightTsarina:
Related to: archivematica/Issues#1019, artefactual/archivematica-storage-service#511

### DIFF
--- a/src/archivematica/storage_service/locations/models/space.py
+++ b/src/archivematica/storage_service/locations/models/space.py
@@ -820,9 +820,12 @@ class PosixMoveUnsupportedError(Exception):
 
 def _scandir_public(path):
     """Generate all directory entries, excluding hidden files."""
-    for entry in os.scandir(path):
-        if not entry.name.startswith("."):
-            yield entry
+    try:
+        for entry in os.scandir(path):
+            if not entry.name.startswith("."):
+                yield entry
+    except OSError:
+        return
 
 
 def _scandir_files(path):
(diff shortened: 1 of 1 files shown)


src/archivematica/storage_service/locations/models/space.py
```diff
@@ -849,14 +852,17 @@ def path2browse_dict(path):
 
     for entry in sorted(_scandir_public(path), key=lambda e: e.name.lower()):
         entries.append(entry.name)
```
replaceafill, line 854: Because `entries.append(entry.name)` happens before `entry.is_dir()` / `entry.stat()`, an `OSError` from those calls will leave the entry in `entries` while omitting it from `directories` and `properties`. Is that intended? If not, should we append only after the entry has been successfully evaluated, or remove it in the exception path?

src/archivematica/storage_service/locations/models/space.py
```diff
@@ -820,9 +820,12 @@ class PosixMoveUnsupportedError(Exception):
...
     """Generate all directory entries, excluding hidden files."""
-    for entry in os.scandir(path):
-        if not entry.name.startswith("."):
-            yield entry
+    try:
+        for entry in os.scandir(path):
+            if not entry.name.startswith("."):
+                yield entry
+    except OSError:
+        return
```
replaceafill, line 828: If we start ignoring `OSError` here and in `path2browse_dict` below, should we log the exception at debug or warning level so operators have some hint that browse results may be incomplete?

replaceafill (commented):
Thank you @NightTsarina. I left a couple of questions.

src/archivematica/storage_service/locations/models/space.py
NightTsarina: I think that will just add noise, and there is nothing to fix, as it is normal that some directories won't we world-readable (i.e. `/lost+found`).

Plus, this function returns the *contents* of a directory, and if the call to `os.scandir` fails, it just return an empty generator, which I think is the most reasonable result:

```
In [11]: list(_scandir_public('/usr'))
Out[11]: 
[<DirEntry 'share'>,
 <DirEntry 'local'>,
 <DirEntry 'games'>,
 <DirEntry 'bin'>,
 <DirEntry 'include'>,
 <DirEntry 'libexec'>,
 <DirEntry 'lib64'>,
 <DirEntry 'src'>,
 <DirEntry 'lib'>,
 <DirEntry 'sbin'>]

In [12]: list(_scandir_public('/lost+found/'))
Out[12]: []
```

src/archivematica/storage_service/locations/models/space.py
replaceafill: Answering my own question after digging into it: leaving `entries.append()` where it is looks right to me.

When `is_dir()` or `stat()` raises, the entry stays in `entries` but is left out of `directories` and `properties`. That is the same shape we already produce for directories that fail the `os.access()` check, and the `Space.browse` docstring already says `properties` may not cover every entry, so callers have to handle it anyway. Appending only after a successful evaluation would instead make such entries disappear from the listing, which seems worse for whoever is browsing the location.

replaceafill (approved):
Thank you @NightTsarina!

PR #937, opened 2026-05-26
Set UTC as default timezone

NightTsarina:
The current value of `America/Los_Angeles` is a historical artifact[1] from old Django versions, and new templates default to UTC.

Keeping the current default is causing the storage service to display times in the web UI and in log messages using a timezone that is different from the other AM components (they are already using UTC by default) and also different from the system's timezone, which is very confusing for users (and sysadmins!)

Since the `USE_TZ` setting is set to `True` by default, all the datetime objects are timezone-aware and so this change should not cause any issues with existing data.

[1]: https://docs.djangoproject.com/en/5.2/ref/settings/#std-setting-TIME_ZONE

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -90,7 +90,7 @@ of these settings or provide values to mandatory fields.
 - **`TIME_ZONE`**:
   - **Description:** application time zone. See [TIME_ZONE] for more details.
   - **Type:** `string`
-  - **Default:** `"America/Los_Angeles"`
+  - **Default:** `"UTC"`
 
 - **`SECRET_KEY`**:
   - **Description:** a secret key used for cryptographic signing. See [SECRET_KEY]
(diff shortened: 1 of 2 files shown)


replaceafill (changes requested):
Thanks @NightTsarina, this looks good. Could we also update the default value of the setting in https://github.com/artefactual/archivematica-storage-service/tree/qa/0.x/install#application-specific-environment-variables?

NightTsarina:
> Thanks @NightTsarina, this looks good. Could we also update the default value of the setting in https://github.com/artefactual/archivematica-storage-service/tree/qa/0.x/install#application-specific-environment-variables?

Ah, I forgot about that, thanks for the reminder. It is updated now, PTAL?

replaceafill (approved):

PR #950, opened 2026-06-17
Limit rsync execution time

### LINKED ISSUE #1800: Problem: Storage Service rsync moves can keep requests running indefinitely
**Expected behaviour**

Storage Service operations that move packages or files with rsync should eventually return a terminal result to the caller. If rsync cannot complete, the operation should fail with a useful error and clean up the rsync process.

**Current behaviour**

`Space.move_rsync` starts rsync with `subprocess.Popen` and then waits with `communicate()` without a Storage Service timeout.

As a result, an rsync process that stalls or never exits can keep a Storage Service request or task running indefinitely. The user-visible result is that package moves or transfer-source copies can appear stuck and may require manual intervention.

The current implementation also captures rsync output without a size limit, so a noisy failure can accumulate a large amount of output before the process exits or is interrupted externally.

**Steps to reproduce**

1. Use a Storage Service code path that calls `Space.move_rsync`, such as moving a package between filesystem-backed locations or copying from a transfer source that uses rsync.
2. Run it against an rsync process or remote endpoint that stays connected but never completes.
3. Observe that the request or task remains running until rsync, the worker, or the request is interrupted externally.

**Your environment (version of Archivematica, operating system, other relevant details)**

This appears to affect the latest Archivematica and Storage Service releases, as well as older releases. The relevant Storage Service code path has historically run `Space.move_rsync` synchronously through `subprocess.Popen(...).communicate()` with no timeout.

---

**For Artefactual use:**

Before you close this issue, you must check off the following:

- [ ] All pull requests related to this issue are properly linked
- [ ] All pull requests related to this issue have been merged
- [ ] A testing plan for this issue has been implemented and passed (testing plan information should be included in the issue body or comments)
- [ ] Documentation regarding this issue has been written and merged (if applicable)
- [ ] Details about this issue have been added to the release notes (if applicable)

sevein:
Move rsync subprocess handling into a top-level Storage Service module. This keeps Space.move_rsync focused on building the command and passing copy context to the runner.

Add an rsync I/O timeout of 300 seconds. This makes stalled transfers fail when rsync observes no input or output activity, while still allowing large transfers to continue as long as they keep making progress.

Keep a 24 hour wall-clock timeout as a final cleanup guard for rsync processes that continue producing output but never finish. Run rsync in its own process group so timeout handling can terminate rsync and child processes together.

Capture stdout and stderr through a bounded buffer and include the retained output in StorageException messages. This gives synchronous API callers a terminal failure and lets asynchronous callers record the task as errored.

Add focused unit tests for command construction, non-zero rsync exits, bounded output capture, and timeout cleanup.

Fixes https://github.com/archivematica/Issues/issues/1800.

### DIFF
--- a/src/archivematica/storage_service/common/rsync.py
+++ b/src/archivematica/storage_service/common/rsync.py
@@ -1,0 +1,174 @@
+"""Run rsync commands with Storage Service failure boundaries.
+
+The storage models build the rsync command line because they know the source,
+destination, transport style, and permission flags needed for each move. This
+module owns the process-running policy shared by those callers: bound captured
+output, enforce a maximum wall-clock runtime, and clean up the whole process
+group when rsync does not exit on its own.
+
+Rsync also receives ``--timeout`` from callers using
+``settings.RSYNC_IO_TIMEOUT_SECONDS``. That is rsync's idle I/O timeout, not a
+total runtime limit. The runner's wall-clock timeout is a separate safety net
+so a transfer that keeps producing some output cannot keep an API request or
+worker alive forever.
+"""
+
+import logging
+import os
+import selectors
+import signal
+import subprocess
+import time
+from collections.abc import Mapping
+from collections.abc import Sequence
+from typing import IO
+from typing import cast
+
+from django.conf import settings
+
+LOGGER = logging.getLogger(__name__)
+
+__all__ = ("RsyncError", "run_rsync")
+
+# Keep enough stderr/stdout for diagnosis without holding unbounded output.
+RSYNC_OUTPUT_LIMIT_BYTES = 65536
+
+# Give rsync a short window to handle SIGTERM before forcing SIGKILL.
+RSYNC_TERMINATE_GRACE_SECONDS = 10
+
+
+class RsyncError(Exception):
+    """Raised when rsync exits unsuccessfully or exceeds runtime limits."""
+
+
+class _BoundedOutput:
+    """Collect only the tail of process output for failure diagnostics."""
+
+    def __init__(self, limit: int) -> None:
+        self.limit = limit
+        self.output = bytearray()
+        self.truncated = False
+
+    def append(self, chunk: bytes) -> None:
+        """Append output while keeping the buffer within its byte limit."""
+        self.output.extend(chunk)
+        if len(self.output) > self.limit:
+            del self.output[: len(self.output) - self.limit]
+            self.truncated = True
+
+    def text(self) -> str:
+        """Return decoded buffered output, noting when earlier output was cut."""
+        result = self.output.decode(errors="replace")
+        if self.truncated:
+            return f"[output truncated to last {self.limit} bytes]\n{result}"
+        return result
+
+
+def _kill_process_group(process: subprocess.Popen[bytes]) -> None:
+    """Terminate an rsync process group, escalating to SIGKILL if needed."""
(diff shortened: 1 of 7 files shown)


src/archivematica/storage_service/common/rsync.py
```diff
@@ -0,0 +1,182 @@
+"""Run rsync commands with Storage Service failure boundaries.
```
replaceafill, line 1: Could we move this module to the `common` subpackage? The behavior seems reusable outside locations; for example, the `import_aip` management command has a local `rsync` copy path that looks like it could use this runner with little adaptation.

src/archivematica/storage_service/rsync.py
```diff
@@ -0,0 +1,182 @@
...
+LOGGER = logging.getLogger(__name__)
+
+__all__ = ("RSYNC_IO_TIMEOUT_SECONDS", "run_rsync")
+
+# Fail rsync when it observes no I/O for this many seconds. This does not limit
+# total transfer duration; large copies can run longer while making progress.
+RSYNC_IO_TIMEOUT_SECONDS = 300
+
+# Safety net for rsync processes that keep doing some I/O but never finish.
+RSYNC_PROCESS_TIMEOUT_SECONDS = 86400
```
replaceafill: These defaults make sense to me, but should we allow users to tune them through environment variables/settings? Some deployments may have very large or slow transfers where a fixed 5-minute I/O timeout or 24-hour process timeout is too constrained.

src/archivematica/storage_service/rsync.py
```diff
@@ -0,0 +1,182 @@
...
+        """Return decoded buffered output, noting when earlier output was cut."""
+        result = self.output.decode(errors="replace")
+        if self.truncated:
+            return f"[output truncated to last {self.limit} bytes]\n{result}"
+        return result
+
+
+def _storage_exception(message: str) -> "StorageException":
+    """Build StorageException lazily to avoid importing location models here."""
+    from archivematica.storage_service.locations.models import StorageException
```
replaceafill: Related to the request above about moving the module to `common`, should the runner raise its own rsync-specific exception instead of lazily importing `StorageException`? Then `Space.move_rsync` could convert that exception into a `StorageException`, while other callers like `import_aip` could map it to their own exception type.

src/archivematica/storage_service/common/rsync.py
```diff
@@ -0,0 +1,182 @@
...
+    output = _BoundedOutput(RSYNC_OUTPUT_LIMIT_BYTES)
+    started = time.monotonic()
+
+    if process.stdout is None:
+        process.wait(timeout=timeout)
+        return "", time.monotonic() - started
+
+    os.set_blocking(process.stdout.fileno(), False)
+    selector = selectors.DefaultSelector()
+    selector.register(process.stdout, selectors.EVENT_READ)
```
replaceafill, line 108: I had never seen the `selectors` module! 🤯

replaceafill (changes requested):
This looks good @sevein. I think we should move the new module to the `common` subpackage, and I left a couple of comments about the `import_aip` management command but we don't have to modify that in this pull request. I also reviewed the other parts where we call `rsync` and I don't think they are good candidates for reusing this new functionality.

replaceafill (approved):

PR #965, opened 2026-07-07
Skip disabled replicators when creating replicas

replaceafill:
This updates package replica creation to ignore disabled Replicator locations when selecting related replicators from the source location, preventing new replicas from being created in disabled locations that remain linked to an AIP Storage location.

Connected to https://github.com/archivematica/Issues/issues/1809


sevein (approved):
👍

PR #969, opened 2026-07-10
Replace MinIO with SeaweedFS in integration tests

replaceafill:
Use SeaweedFS as the S3-compatible object store in the integration Compose environment, provisioning IAM-style users and AWS-like S3 policies through the setup service while keeping the existing rclone remote name.


sevein (approved):
Nice! 👍

PR #974, opened 2026-07-22
Adopt uv project workflows

sevein:
Replace pip-tools and tox workflows with a universal uv lock and a pinned Python 3.14.6 default. Use uv for local, CI, and container commands. Keep the supported Python range at 3.10 and exercise it through the existing integration matrix.

Connected to https://github.com/archivematica/Issues/issues/1788


replaceafill (approved):
Thanks @sevein!

PR #980, opened 2026-08-03
Preserve interrupted async operation results

sevein:
This change prevents interrupted asynchronous operations from disappearing.
When an operation stops heartbeating, Storage Service now keeps its status
record and reports a terminal error with an unknown final outcome.

## How asynchronous operations work

Some Storage Service requests start work in a background thread. Storage
Service records the operation and gives the caller an identifier. The caller
uses that identifier to poll the operation until it finishes:

| Poll result | Meaning |
| --- | --- |
| `completed=false` | Work is still running; poll again |
| `completed=true` | Work ended; read its result or error |
| `404` | No status record exists for that identifier |

A watchdog refreshes the heartbeat of each running operation. If the process
that owns the operation disappears, its heartbeat stops. Storage Service can
then tell that it lost the operation, but not how far the work progressed.

The operation may have produced an external side effect before it stopped.
Retrying it without knowing the outcome could therefore duplicate work.

## The problem

The failure that motivated this change exposed two related gaps:

| Failure mode | Effect | Fixed here? |
| --- | --- | --- |
| A live operation never completes | Caller polls forever | No |
| An operation stops heartbeating | Record deleted; poll returns `404` | Yes |

The first gap tied up the workers handling the polling calls. It requires an
overall observation deadline in the caller.

This PR fixes the second gap. The watchdog previously deleted the status record
after its heartbeat expired. The caller could not tell whether the identifier
was invalid, the result had aged out, or the operation had been interrupted.
The watchdog now retains the record as a terminal interruption.

## Observable behaviour

| State | Before | After |
| --- | --- | --- |
| Running or normal result | Existing `200` | Unchanged |
| Task exception | `200`; terminal error | Unchanged |
| Heartbeat expires | Row deleted; `404` | Retained; `200` terminal error |
| Result retention expires | Row deleted; `404` | Unchanged |

With the repository defaults, heartbeat expiry happens after 120 seconds. The
interrupted result is then retained for approximately 24 hours through the
existing completed-result cleanup path.

The error tells clients that the operation was interrupted, its final outcome
is unknown, and it must not be retried automatically.

## What changed

- Expired incomplete rows become completed errors using the existing async
  error format. No schema or migration change is required.
- The watchdog records local completion and live heartbeats before checking for
  expiry. Row locking and terminal-state checks ensure that a late result cannot
  replace an already recorded terminal state.
- Interrupted operations produce a warning containing their IDs and increment
  the Prometheus counter `async_manager_interrupted_tasks_total`.

## Compatibility and downstream work

The response schema is unchanged, but the behaviour is not fully backward
compatible: an interrupted operation now returns `200` instead of `404`, and
`was_error=true` can now describe a synthetic interruption as well as an
exception raised by the task. Interrupted rows also live until the normal
result-retention deadline, and a late result no longer overwrites them.

### DIFF
--- a/src/archivematica/storage_service/locations/models/async_manager.py
+++ b/src/archivematica/storage_service/locations/models/async_manager.py
@@ -15,6 +15,7 @@ import threading
 import time
 import traceback
 
+from django.db import transaction
 from django.utils import timezone
 
 from archivematica.storage_service.locations import metrics
@@ -86,35 +87,6 @@ class AsyncManager:
         with AsyncManager.lock:
             now = timezone.now()
 
-            # Mark any tasks that have expired before finishing as interrupted
-            # (i.e. interrupted due to a server restart)
-            Async.objects.filter(
-                completed=False,
-                updated_time__lte=(now - TASK_TIMEOUT_SECONDS),
-            ).update(
-                completed=True,
-                was_error=True,
-                completed_time=now,
-                updated_time=now,
-                _error=INTERRUPTED_TASK_ERROR,
-            )
-
-            # Delete any tasks whose results have expired
-            Async.objects.filter(
-                completed=True,
-                completed_time__lte=(now - MAX_TASK_AGE_SECONDS),
-            ).delete()
-
-            # Touch the update time of any running task.  If we crash/restart then these will expire.
-            running_task_ids = [
-                task.async_id
-                for task in AsyncManager.running_tasks
-                if task.thread.is_alive()
-            ]
-            Async.objects.filter(id__in=running_task_ids).update(
-                updated_time=timezone.now()
-            )
-
             # Find any tasks that have completed since we last looked
             completed_tasks = [
                 task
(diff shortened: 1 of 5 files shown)


src/archivematica/storage_service/locations/models/async_manager.py
```diff
@@ -121,6 +115,56 @@ def _watchdog_loop():
                         % (task.async_id)
                     )
 
+            # Touch the update time of any running task. If we crash or restart,
+            # these tasks will expire.
+            running_task_ids = [
+                task.async_id
+                for task in AsyncManager.running_tasks
+                if task.thread.is_alive()
```
replaceafill: Could `running_task_ids` be derived from the tasks remaining after `completed_tasks` are removed, or could both groups be built from one liveness snapshot?

I’m wondering about a thread that is alive during the first check but finishes before this one. If its heartbeat is already stale, it would be in neither group and the expiry update below would mark it interrupted, causing its known result to be discarded.

Am I reading that race correctly?

sevein: Thanks, I think you're right. The watchdog now checks the thread status once, saves completed tasks, refreshes the remaining tasks, and then expires stale records.

I also added an `async_operation_interrupted` error code, so clients can identify an interrupted operation without matching the full error message.

replaceafill (approved):

PR #1000, opened 2026-08-28
Wait for Keycloak before running the integration tests

replaceafill:
The test container was gated on mysql and seaweedfs being healthy, but keycloak was only listed under `links`, which merely means the container has started. Keycloak (start-dev --import-realm) opens its HTTP listener only after the realm import has finished, so any test that reaches the OIDC login page early enough in the run finds nothing listening and times out. Because pytest-randomly shuffles the test order, this made the OIDC integration tests fail intermittently.

Add a healthcheck to the keycloak service that requests the OIDC discovery document of both realms the tests log in to, and make the test container depend on it being healthy. The Keycloak image ships no curl or wget, so the check uses bash's /dev/tcp. Drop the redundant `links` block now that all dependencies are declared with `depends_on`.

### DIFF
--- a/tests/integration/docker-compose.yml
+++ b/tests/integration/docker-compose.yml
@@ -81,10 +81,8 @@ services:
         condition: service_healthy
       seaweedfs-setup:
         condition: service_completed_successfully
-    links:
-      - "seaweedfs"
-      - "mysql"
-      - "keycloak"
+      keycloak:
+        condition: service_healthy
 
   seaweedfs:
     image: "chrislusf/seaweedfs:4.45"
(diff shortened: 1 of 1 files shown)


sevein (approved):
LGTM! A potential future cleanup for this document could be using the top-level `configs` section to define the different scripts we're using, e.g.:

```
configs:
  keycloak-healthcheck:
    content: |
      for realm in demo secondary; do
        exec 3<>/dev/tcp/127.0.0.1/8080 &&
        printf 'GET /realms/%s/.well-known/openid-configuration HTTP/1.1\r\nHost: localhost\r\nConnection: close\r\n\r\n' "$$realm" >&3 &&
        head -n 1 <&3 | grep -q ' 200 ' || exit 1
      done
```

You can refer to them from a service like in the following example:

```
    configs:
      - source: keycloak-healthcheck
        target: /healthcheck.sh
```

Looks like `content` was introduced in Docker Compose v2.23 (late 2023).

replaceafill:
Thanks @sevein! I didn't know about the [configs](https://docs.docker.com/reference/compose-file/configs/) element and the result https://github.com/artefactual/archivematica-storage-service/pull/1000/commits/98c46bb70d5952d111df0e9d30deb64314ee602e looks great.

PR #1012, opened 2026-09-17
Keep processing configuration out of reingested AIPs

replaceafill:
The reingest request now works on a copy whenever it has to add a processing configuration and discards the copy once the request is answered, so the stored AIP is never written to. Compressed and remote AIPs already worked on copies and are unchanged, and so are requests with the default configuration.

The fix is introduced by first extending test coverage of the reingest request and finalization, then putting the compression tools behind a small archiver interface with a fake for tests, and moving the request into its own typed module.

Beyond the issue, a failing compression or extraction is now an error; before, its exit status was only logged. The command lines and their statuses are still logged as before.

Connected to https://github.com/archivematica/Issues/issues/1813


sevein (approved):
LGTM!

PR #1018, opened 2026-09-18
Extract reingested AIPs by their recorded compression

replaceafill:
At the end of a reingest, the AIP the pipeline sends back was extracted with unar, detecting its format, although its compression is known: the pointer file the pipeline wrote, or else the compression event it sent, names it, and the stored AIP is already extracted by the tool of the compression its pointer file records. The issue describes which entries unar cannot decode and what that did to stored AIPs.

The reingested AIP is now extracted by the tool of that compression, determined before the extraction, as the stored AIP already is. unar is no longer involved in reingest; it still extracts compressed packages that have no pointer file, and the DSpace and Archipelago spaces still call it directly, which this change leaves as they are.

Connected to https://github.com/archivematica/Issues/issues/1818


sevein (approved):
LGTM

